PR #9, opened 2019-06-12
initial maccy obj once

chengjk:
initial maccy obj at init(). other wise 'maccy.stgart()' and  'maccy.popUp()' is not same obj.
fix it , and then 'refresh() at Maccy#popUp()' is unnecessary.


p0deje:
Thank you!

PR #10, opened 2019-06-13
add hotKey for menu Item

chengjk:
add hotKey for menu Item. it useful when choose item, especially when arrow key event broken


p0deje:
I believe that arrow keys are now fixed (https://github.com/p0deje/Maccy/releases/tag/0.3.4) so would you still like to merge this? If yes, can you please demonstrate how it looks and behaves and please rebase the PR with master.

chengjk:
add num as hot key at the end of menu item .  
image . 
 hotkey will refresh after search.  
image
and  both of them  does not affect  system item.

p0deje:
Alright, I understand the reasoning for this feature, but I believe there are few things to take care before landing this:

1. It doesn't work with more than 9 items. I'd like to keep the feature of using any amount of history items the user wants, so I think this feature should add hotkey only for the first 9 items while keeping others untouched. Currently, it looks buggy:

Screenshot 2019-06-14 at 11 44 24

2. The hotkey should only work with a modifier key (e.g. `Cmd+2` instead of `2`. User should be able to type numbers during search-as-you-type.
3. Please run `swiftlint` and fix the violations.

Maccy/History.swift
```diff
@@ -5,7 +5,7 @@ class History {
   private let sizeKey = "historySize"
 
   init() {
-    UserDefaults.standard.register(defaults: [sizeKey: 999])
+    UserDefaults.standard.register(defaults: [sizeKey: 9])
```
p0deje: This should be reverted - we still allow to override this value using `defaults`.

Maccy/Menu.swift
```diff
@@ -10,7 +10,9 @@ class Menu: NSMenu {
 
   override init(title: String) {
```
p0deje: Can this be removed now?

chengjk: yes . u are right

chengjk:
finish . revert history default to  999.  max hotkey  is 9.  hotkey work with 'Cmd'.

p0deje:
I have fixed swiftlint issues, squashed and committed in 29b0c3eb67d5bd17a67ae4c63fa8e33fe38cc8da. Thank you for the contribution!

PR #11, opened 2019-06-30
Add Homebrew cask and update readme with installation instructions

jlowgren:
I submitted a [maccy](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/maccy.rb) cask to Homebrew, so that it can be installed via:

```bash
brew cask install maccy
```


p0deje:
Thank you for the PR and, more importantly, for Homebrew/homebrew-cask#65459.

Issue #16, opened 2019-07-19
Focus on "Quit" on search


saeedseyfi:
Hi,
Thanks for the nice app <3
There is an issue with search functionality:

**What happens:**
- When I start typing(searching), it sets the focus on Quit. Even if there is a result for search!

**Whats expected:**
1. Quit will close the app completely, so it never should get a focus automatically.
2. If there is a result for the searched text, it should focus on the first item.
3. (Optional) You can add Hide option that hides the menu. On search, if there is no result, it can focus on Hide.

Screen record: [Record.mov.zip](https://github.com/p0deje/Maccy/files/3411130/Record.mov.zip)

p0deje:
Thank you for the report!

I have fixed this issue in b5fcf193f15a0b8e87242d394d4d0b59a1bafb7e, but it's not released yet.

p0deje:
Please try v0.3.7 - it should be fixed.

saeedseyfi:
@p0deje How can I get the update by brew?

p0deje:
@saeedseyfi Somebody needs to update https://github.com/Homebrew/homebrew-cask/blob/master/Casks/maccy.rb as it points to 0.3.6

saeedseyfi:
What is the new `sha256` for 0.3.7? I'll create a pr to update the cask.

p0deje:
08e079ae2071bea7e1f011f47ba601b79d9440daa8e80aa5d3946d558a80ab43

saeedseyfi:
@p0deje What do you think if cask gets the latest version?
https://github.com/Homebrew/homebrew-cask/pull/66608/files

saeedseyfi:
@p0deje Please take a look at the this, the build is failing because of the specified `depends_on` in the cask file: https://travis-ci.org/Homebrew/homebrew-cask/builds/562531636
What do you think?

p0deje:
This failure was introduced in #65568, so your changes are unrelated. I don't know how brew/cask maintainers handle such situations.

p0deje:
I should also be possible to just point to the latest version to avoid the need to bump it every time:

```diff
diff --git a/Casks/maccy.rb b/Casks/maccy.rb
index 27abe054d..68a20a450 100644
--- a/Casks/maccy.rb
+++ b/Casks/maccy.rb
@@ -1,8 +1,8 @@
 cask 'maccy' do
-  version '0.3.6'
[...]

PR #18, opened 2019-07-24
Prevent auto focus on system items

saeedseyfi:
The new release partially fixed issue #16. The proposed changes will prevent the focus on "Clear" item on search if has no result.

### DIFF
--- a/Maccy/Menu.swift
+++ b/Maccy/Menu.swift
@@ -47,7 +47,9 @@ class Menu: NSMenu, NSMenuDelegate {
       }
     }
 
-    highlight(highlightableItems(items).first)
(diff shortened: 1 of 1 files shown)


p0deje:
Is it a problem that "Clear" is focused? 

With this change, it is now impossible to navigate system items with arrow keys as far as I can tell.

saeedseyfi:
Yes, exactly that is the problem. "Clear" will delete the whole search history immediately, why should it be auto focused?!

Please check the new commit, navigation by arrow keys is fixed now.

p0deje:
Looks good. Thanks for the PR!

PR #21, opened 2019-08-09
Native status icon hiding & the beginning of normal settings

DaFuqtor:
### To hide the status bar icon
1. Drag it away from status bar holding <kbd>Command</kbd> until you see a cross.
2. Let it go.
> To recover the icon, open Maccy when it's already running.

### Settings: The Beginning

Hey, it would be nice to add all the settings here. I've added just one — "Paste by Default" — and that's already cool.

new Settings

Hiding status bar item

Recovering status bar item

### DIFF
--- a/Maccy/AppDelegate.swift
+++ b/Maccy/AppDelegate.swift
@@ -16,4 +16,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     maccy.start()
     hotKey.handler = { self.maccy.popUp() }
   }
+  func applicationShouldHandleReopen(_ sender: NSApplication,
+                                   hasVisibleWindows flag: Bool) -> Bool {
+    maccy.statusItem.isVisible = true
+    return true
+  }
 }
--- a/Maccy/Maccy.swift
+++ b/Maccy/Maccy.swift
@@ -2,14 +2,26 @@ import Cocoa
 
 class Maccy {
   private let about = About()
-  private let statusItem = NSStatusBar.system.statusItem(withLength: NSStatusItem.squareLength)
+  let statusItem = NSStatusBar.system.statusItem(withLength: NSStatusItem.squareLength)
   private let menu = Menu(title: "Maccy")
 
-  private let showInStatusBar = "showInStatusBar"
   private let pasteByDefault = "pasteByDefault"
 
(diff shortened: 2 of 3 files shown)


p0deje:
Thanks for your interest in the app and the contributions! I'd like to go over some of the changes to clarify how I see them:

1. For the removal of the status menu icon - it is a great idea! I'd like to keep `showInStatusBar` configuration so that existing user don't need to do anything to hide it again.
2. For the settings, I'd prefer not to implement it this way. Not all the settings are true/false, so it won't be easy to implement them inside the menu itself. I'd rather have "Preferences" menu item which will open a typical preferences window with all the configuration there.

DaFuqtor:
> keep `showInStatusBar` configuration so that existing user don't need to do anything to hide it again

Well, let's just get the configuration (in the next version) and  make `statusItem.isVisible = false` if the user has already changed `showInStatusBar` to 0. And later, in future versions, completely remove `showInStatusBar` setting due to its "duplication" of built-in macOS abilities.

---

> Not all the settings are true/false...
typical preferences window

If you plan to add more settings, this sound good. For present number of configurations it would be good to add submenu `Preferences` button (with an arrow on the right side, I mean), which would store all 4 settings in it. For boolean preferences, all thing is obvious. For `historySize` — a slider, like in the native Volume status item. For `hotKey` — just input that receives any combination.

p0deje:
> Well, let's just get the configuration (in the next version) and make statusItem.isVisible = false if the user has already changed showInStatusBar to 0. And later, in future versions, completely remove showInStatusBar setting due to its "duplication" of built-in macOS abilities.

Good idea. I like that. What do you think about extracting this code into a separate PR?

> If you plan to add more settings, this sound good. For present number of configurations it would be good to add submenu Preferences button (with an arrow on the right side, I mean), which would store all 4 settings in it. For boolean preferences, all thing is obvious. For historySize — a slider, like in the native Volume status item. For hotKey — just input that receives any combination.

I think the submenu with sliders and inputs might look a bit overwhelming, but I agree this is a solution. I would still prefer to have all the settings implemented at once rather than merging a PR with one option and then adding everything else. Frankly, I don't have time to work it now.

PR #49, opened 2019-09-28
ignore events based on `ignoreEvents` pref

thewisenerd:
set this to true before invoking pass, unset at least 2*timerFrequency post copying password, assuming changeCount is updated by then.

### DIFF
--- a/Maccy/History.swift
+++ b/Maccy/History.swift
@@ -35,7 +35,7 @@ class History {
 
   func add(_ string: String) {
     if UserDefaults.standard.bool(forKey: ignoreEventsKey) {
-        return
+      return
     }
 
     remove(string)
--- a/Maccy/History.swift
+++ b/Maccy/History.swift
@@ -11,6 +11,8 @@ class History {
     }
   }
 
+  private let ignoreEventsKey = "ignoreEvents"
+
   init() {
     UserDefaults.standard.register(defaults: [sizeKey: 200])
     if ProcessInfo.processInfo.arguments.contains("ui-testing") {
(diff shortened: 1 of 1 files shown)


p0deje:
What's the use case of this? I don't quite follow.

thewisenerd:
so, right now, the password copy handling is done as such

```
if let var = pasteboard.items.last {
  copyHooks()
} else {
  removeHooks()
}
```

the else'y bits would be triggered when the password manager would do a `pasteboard.clearContents()`

however, this need not be the case, if the password manager can preemptively tell the clipboard manager that they should start ignoring events, and the password manager co-operates.

since the clipboard data picking happens on `changeCount` *and* `?.last`, `changeCount` will take care of the ignore bits, and the clipboard manager would never read the events which were triggered when the flag was set.

This fits into my workflow of using [passwordstore.org](https://www.passwordstore.org), where there is a platform defined `clip()` function;

it overall looks something like this (for the sake of brevity):
```
clip() {
  before = $(pbpaste)
  defaults write "org.p0deje.Maccy" ignoreEvents true
  ( sleep 5; defaults write "org.p0deje.Maccy" ignoreEvents false ) &
  ( sleep $CLIP_TIME; echo -n "$before" | pbcopy ) &
  echo -n "$1" | pbcopy
}
```

thewisenerd:
the `2*timerFrequency` to await until `ignoreEvents` is unset is so that changeCount is actually reflected

p0deje:
Thank you, merged in 4fe79d8

PR #50, opened 2019-09-28
ignore blank strings from history.add

thewisenerd:


Maccy/History.swift
```diff
@@ -31,7 +31,25 @@ class History {
     return savedHistory
   }
 
+  func isBlank(_ string: String) -> Bool {
+    if string.count == 0 {
+      return true
+    }
+
+    if let first = string.first {
+      if first == "\n" {
```
p0deje: This will return true for `\n something`. I guess that's not the intention.

Maccy/History.swift
```diff
@@ -31,7 +31,25 @@ class History {
     return savedHistory
   }
 
+  func isBlank(_ string: String) -> Bool {
```
p0deje: There is actually no need to have a custom function - you can use Swift for that:

```swift
string.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
```

Here are examples of how it works:

```
(lldb) po "foo".trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
false

(lldb) po "".trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
true

(lldb) po " ".trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
true

(lldb) po "\n".trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
true

(lldb) po "\n ".trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
true

(lldb) po "\n foo".trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
false

(lldb) po " foo".trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
false
```

Maccy/History.swift
```diff
@@ -31,7 +31,25 @@ class History {
...
+        return true
+      }
+    }
+
+    return false
+  }
+
   func add(_ string: String) {
+    if isBlank(string) {
+        return
```
p0deje: Please use 2 spaces indentation.

thewisenerd:
resolved

PR #55, opened 2019-10-19
Enable/disable the search field

inbalboa:
did for myself, but maybe you'll accept :-)

### DIFF
--- a/Maccy/Maccy.swift
+++ b/Maccy/Maccy.swift
@@ -27,6 +27,7 @@ class Maccy: NSObject {
 
   override init() {
     UserDefaults.standard.register(defaults: [UserDefaults.Keys.showInStatusBar: UserDefaults.Values.showInStatusBar])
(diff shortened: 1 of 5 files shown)


p0deje:
Thanks for the pull request!

What's the reason for the disabling search field? Is it because you don't like it to be visible or because you want to avoid search functionality at all?

inbalboa:
I think the menu looks prettier without this item. But it's just for my taste :)
It would be cool if the search functionality still worked after hiding the item, but I understand that it is difficult to implement.

p0deje:
I guess all you need to do is set `isVisible` to `false`. That will make it hidden while keeping search functionality.

p0deje:
I've just pushed a different solution that allows to still use search-as-you-type. It's configurable via `hideSearch` setting.

Thank you for the pull request and idea! Even though I haven't merged your changes, I hope you'll like the result.

inbalboa:
Cool, thank you!

Issue #5, opened 2019-04-11
Add Launch at Login feature


patricks:
A launch at login feature would be nice.

p0deje:
Why don't you just add Maccy to your "Login Items"?

patricks:
I know that this would be possible, but a feature like other apps provide ist, would be nice 😉

p0deje:
In order to make it work, I have to implement:

1. a launcher application that allows registering Maccy as a login application (like [this](https://theswiftdev.com/2017/10/27/how-to-launch-a-macos-app-at-login/)) or pull in [dependency](https://github.com/sindresorhus/LaunchAtLogin)
2. create a new settings window where this should be configured

For a person like me who is still not good at Mac app development, it's simply too much work compared to the user doing just a few clicks to add Maccy to login item. I'll be happy to accept PRs implementing ☝️ but I don't see myself working on this in the nearest future.

jaredreich:
@p0deje What's wrong with pulling in `sindresorhus`'s https://github.com/sindresorhus/LaunchAtLogin dependency? I would argue that this feature is absolutely mandatory. Nobody wants to manually add/remove their login items from within system preferences. This alone made me instantly delete the app.

Cool project though, and good work so far!

p0deje:
Reopening this for the future. It's low priority for me, but I'll happily merge PR adding the support.

georgesnow:
https://medium.com/@hoishing/adding-login-items-for-macos-7d76458f6495

this is similar to the GitHub repo posted above. this one I think looks a bit simpler.

DaFuqtor:
I personally prefer using [Clipy/LoginServiceKit](//github.com/Clipy/LoginServiceKit) instead of [sindresorhus/LaunchAtLogin](//github.com/sindresorhus/LaunchAtLogin), because it allows the user to control the setting (which was set in the app) from the default System Preferences later.
I just like to use the built-in features.

PR #57, opened 2019-10-25
Adds the launch at login option in the menu

kmpnz:
See title and https://github.com/p0deje/Maccy/issues/5

Let me know what you think.


kmpnz:
@p0deje The build is failing because of missing carthage build steps. The [LaunchAtLogin](https://github.com/sindresorhus/LaunchAtLogin]) framework requires it: https://github.com/sindresorhus/LaunchAtLogin#can-you-support-cocoapods. I don't have access to the bitrise, so I can't modify it, but those steps need to be added: https://www.bitrise.io/integrations/steps/carthage to make it build. 

It could also a good idea to add the `bitrise.yml` to the repository and contributors can modify it to the pull request needs: https://devcenter.bitrise.io/tips-and-tricks/use-bitrise-yml-from-repository/

p0deje:
@kmpnz Hey! Thank you so much for the pull request. I'm yet to pull this locally and try myself, but meanwhile, let me ask something and leave some comments.

> It could also a good idea to add the bitrise.yml to the repository and contributors can modify it to the pull request needs: https://devcenter.bitrise.io/tips-and-tricks/use-bitrise-yml-from-repository/

I've just pushed it in 56cc452b0ab297bc743ca6b0d642320c14ee8309, so you probably can update the PR to include the Carthage step.

In one of the comments (https://github.com/p0deje/Maccy/issues/5#issuecomment-546512846) @DaFuqtor mentioned that using Clipy/LoginServiceKit would allow controlling the login item from the system preferences pane. Do you think it would be possible to somehow respect the system-wide setting and still use sindresorhus/LaunchAtLogin?

DaFuqtor:
The only question is, why use [sindresorhus/LaunchAtLogin](//github.com/sindresorhus/LaunchAtLogin) at all? It does the same thing as the [Clipy/LoginServiceKit](//github.com/Clipy/LoginServiceKit), except that it doesn't add the app to user's login items.

p0deje:
> The only question is, why use sindresorhus/LaunchAtLogin at all? It does the same thing as the Clipy/LoginServiceKit, except that it doesn't add the app to user's login items.

Fair enough. I guess the only advantage of LaunchAtLogin is to be able to sandbox Maccy, but given I'm not planning to distribute it in App Store in the nearest future (if at all), we can probably just go with LoginServiceKit.

DaFuqtor:
Well, there is an app named [Magnet](//www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=2ahUKEwiP_N2Y4L7lAhUCwqYKHalwDkgQFjAAegQIARAB&url=https%3A%2F%2Fapps.apple.com%2Fus%2Fapp%2Fmagnet%2Fid441258766%3Fmt%3D12&usg=AOvVaw0a1L_8IMXKcmv-Cy497dqi) in App Store, which has "Launch at Login" preference in the app and in the user's Login Items. So, it's possible to distribute apps with this feature.

georgesnow:
There is really 2 reasons adding the item to the system pane(user convenience and transparency). It’s respecting And explicitly in one place for the user to see the app has been added with any other apps. And https://github.com/sindresorhus/LaunchAtLogin explicitly mentions can used sandbox or non-sandbox. I didn’t see in Clipy’s docs saying it was or wasn’t. The sandbox well developer preference when it’s non-MAS app. The first is a bit more important.

kmpnz:
@p0deje Thanks for adding the `bitrise.yml.` Will integrate it

Regarding `LoginServiceKit` vs `LaunchAtLogin`: I don't have a strong preference here. Let me play around with `LoginServiceKit`

p0deje:
I've just pushed https://github.com/p0deje/Maccy/commit/a0be3a15383b3c9b338bde7fe7008b02fb248b2e which should solve the problem with Git clone on CI.

kmpnz:
Thanks @p0deje ! Building, looks better already 🤞

p0deje:
Ok, it builds fine now, there is one test failure but that's just flaky. So we can ignore.

p0deje:
@kmpnz There is quite a lot of noise in the diff. I guess most of it is because you are using a newer version of CocoaPods than I do (I use CocoaPods.app). If you don't mind, I'll just revert those changes by myself and squash to your commits. Do you have any objections?

kmpnz:
@p0deje Nope, not at all. Go ahead and lets merge this baby 🎉

p0deje:
@kmpnz Thank you very much! This is now merged in acee9aa202380970e0e4a12ac1472f4bd347f019 and released in 0.7.0.

Gsazraetr:
I must say I am not a fan of this new implementation. Maccy's minimalism is arguably its greatest strength; this is an additional menu item that is wholly redundant after it is set.
There is nothing disgraceful about adding an app to Login Items. Is there a way to hide this?

p0deje:
At this point, there is no way to hide, but at some point in the future, I'd like to replace it with more generic preferences submenu. Would it work for you @Gsazraetr?

Gsazraetr:
Absolutely. `About` and maybe even `Quit` can then be moved there too. Take your time, this is a just a minor gripe. You've done great work on the app so far.

Issue #75, opened 2020-02-03
Copying text in Word and losing formatting


simpalsimon:
Love the app. If I'm working with two word documents open, and want to skim through one document and copy all the bits which I want to have on the clipboard for easy re-ordering, I lose the formatting of the text by switching between what to paste.

So if I copy two bits of text from the same word document, and both have words in bold or italicised, this will be lost if I switch between the two bits in the maccy menu. Is this intended?

p0deje:
It's not quite intended, but it's the way it is. There was a request to allow pasting without formatting so anything pasted from Maccy will lose formatting. I don't know how to support both with and without formatting at the same time as they are contradicting.

simpalsimon:
Could this feature be toggle-able? Just a friendly suggestion.

FernandoMiguel:
The way I did in the past is cmd shift V pastes without format

simpalsimon:
> The way I did in the past is cmd shift V pastes without format

Can this be done in the current version?

FernandoMiguel:
> > The way I did in the past is cmd shift V pastes without format
> 
> Can this be done in the current version?

don't think this app ever supported.
but is something, MS Office, chrome, and even some web editors supported for many years

simpalsimon:
> > > The way I did in the past is cmd shift V pastes without format
> > 
> > 
> > Can this be done in the current version?
> 
> don't think this app ever supported.
> but is something, MS Office, chrome, and even some web editors supported for many years

Would be good to have this re-integrated into maccy then. Would like to paste with formatting preserved.

p0deje:
> The way I did in the past is cmd shift V pastes without format
We cannot do this in Maccy, can we?

alapolloni:
Hi.  I think that Maccy should keep/save the formatting.

Then, to paste without....you do the normal Maccy select action Shift+Cmd+C -> select what you want -> and then use the standard MacOS shortcut to past without formatting.  
> Option+Shift+Command+V to paste text without any formatting

While there are other alternatives this seems the most straightforward/easiest to implement.

p0deje:
Hm, then you cannot use keyboard shortcuts to select for example item number 5 on the list. Don't you think that makes the whole UX inconsistent?

alapolloni:
I think I'm missing something from your question.  
[...]

PR #84, opened 2020-03-17
Add support for PasteboardType.png in Clipboard

### LINKED ISSUE #83: Screenshot to clipboard does not insert the screenshot to history
Making a screenshot that saves itself to the clipboard (Control+Shift+Command+4) does not create a history item in Maccy. It works in Clipy, though.

Maccy 0.9.0, installed via brew.

Defman21:
Fixed #83 

Note: declaring the data as `.tiff` later (when selecting an item from the history menu and pasting it in the clipboard) works fine. I guess it auto-converts image data under the hood. Not sure if we need to distinguish tiff/png at the moment of creating a `HistoryItem` instance and then when we copy it to the clipboard, set the corresponding type for it (in `Clipboard.swift:copy`).

@p0deje thoughts?

### DIFF
--- a/Maccy/AppDelegate.swift
+++ b/Maccy/AppDelegate.swift
@@ -56,5 +56,18 @@ class AppDelegate: NSObject, NSApplicationDelegate {
       })
       UserDefaults.standard.migrations["2020-02-22-history-item-add-number-of-copies"] = true
     }
+
+    if UserDefaults.standard.migrations["2020-03-18-store-pasteboard-type"] != true {
+      UserDefaults.standard.storage = UserDefaults.standard.storage.map({ item in
+        let migratedItem = item
+        if let _ = NSImage(data: item.value) {
+          migratedItem.types = [.tiff]
+        } else {
+          migratedItem.types = [.string]
+        }
+        return migratedItem
+      })
+      UserDefaults.standard.migrations["2020-03-18-store-pasteboard-type"] = true
+    }
   }
 }
--- a/Maccy/AppDelegate.swift
+++ b/Maccy/AppDelegate.swift
@@ -61,9 +61,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
       UserDefaults.standard.storage = UserDefaults.standard.storage.map({ item in
         let migratedItem = item
         if NSImage(data: item.value) != nil {
-          migratedItem.types = [.tiff]
+          migratedItem.type = .tiff
         } else {
-          migratedItem.types = [.string]
+          migratedItem.type = .string
         }
         return migratedItem
       })
--- a/Maccy/AppDelegate.swift
+++ b/Maccy/AppDelegate.swift
@@ -60,7 +60,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     if UserDefaults.standard.migrations["2020-03-18-store-pasteboard-type"] != true {
       UserDefaults.standard.storage = UserDefaults.standard.storage.map({ item in
         let migratedItem = item
-        if let _ = NSImage(data: item.value) {
+        if NSImage(data: item.value) != nil {
           migratedItem.types = [.tiff]
         } else {
           migratedItem.types = [.string]
--- a/Maccy/AppDelegate.swift
+++ b/Maccy/AppDelegate.swift
@@ -29,7 +29,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
       if let oldStorage = UserDefaults.standard.array(forKey: UserDefaults.Keys.storage) as? [String] {
         UserDefaults.standard.storage = oldStorage.compactMap({ item in
           if let data = item.data(using: .utf8) {
-            return HistoryItem(value: data)
+            return HistoryItem(typesWithData: [.string: data])
           } else {
             return nil
           }
@@ -60,10 +60,10 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     if UserDefaults.standard.migrations["2020-03-18-store-pasteboard-type"] != true {
       UserDefaults.standard.storage = UserDefaults.standard.storage.map({ item in
         let migratedItem = item
-        if NSImage(data: item.value) != nil {
-          migratedItem.type = .tiff
+        if item.value != nil && NSImage(data: item.value!) != nil {
+          migratedItem.typesWithData = [.tiff: item.value!]
         } else {
-          migratedItem.type = .string
+          migratedItem.typesWithData = [.string: item.value!]
         }
         return migratedItem
       })
--- a/Maccy/Clipboard.swift
+++ b/Maccy/Clipboard.swift
@@ -6,7 +6,7 @@ class Clipboard {
 
   private let pasteboard = NSPasteboard.general
   private let timerInterval = 1.0
-  private let supportedTypes: [NSPasteboard.PasteboardType] = [.tiff, .png, .string]
+  private let supportedTypes: [NSPasteboard.PasteboardType] = [.tiff, .png, .string, .fileURL]
 
   // See http://nspasteboard.org for more details.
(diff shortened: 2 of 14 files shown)


Defman21:
I implemented saving image type to HistoryItem and moved the `if else` statement that set the Pasteboard type from `Menu.copy` to `HistoryItem.getPasteboardType`. I'm not familiar with Swift so you might have a better idea at implementing that.

p0deje:
What do you think if we instead store all the pasteboard item types in the history item itself? This will allow to avoid all this if/else branching and will make it much easier to support different types in the future. We'll also preserve the types just like they were copied initially.

Something along these lines (haven't tested).

```diff
diff --git a/Maccy/Clipboard.swift b/Maccy/Clipboard.swift
index aff0100..5930abe 100644
--- a/Maccy/Clipboard.swift
+++ b/Maccy/Clipboard.swift
@@ -6,6 +6,7 @@ class Clipboard {
 
   private let pasteboard = NSPasteboard.general
   private let timerInterval = 1.0
+  private let supportedTypes: [NSPasteboard.PasteboardType] = [.tiff, .png, .string]
 
   // See http://nspasteboard.org for more details.
   private let ignoredTypes: Set = [
@@ -78,18 +79,15 @@ class Clipboard {
     // See https://github.com/p0deje/Maccy/issues/78.
     pasteboard.pasteboardItems?.forEach({ item in
       if !shouldIgnore(item.types) {
-        if item.types.contains(.tiff) {
-          if let data = item.data(forType: .tiff) {
-            let historyItem = HistoryItem(value: data)
-            historyItem.type = .image
-            onNewCopyHooks.forEach({ $0(historyItem) })
-          }
-        } else {
-          if let data = item.data(forType: .string) {
-            let historyItem = HistoryItem(value: data)
-            historyItem.type = .string
-            onNewCopyHooks.forEach({ $0(historyItem) })
+        for type in supportedTypes {
+          if item.types.contains(type) {
+            if let data = item.data(forType: type) {
+              let historyItem = HistoryItem(value: data)
+              historyItem.types = item.types
+              onNewCopyHooks.forEach({ $0(historyItem) })
+            }
           }
+          break
         }
       }
     })
diff --git a/Maccy/HistoryItem.swift b/Maccy/HistoryItem.swift
index f13d88b..202ee02 100644
--- a/Maccy/HistoryItem.swift
+++ b/Maccy/HistoryItem.swift
@@ -1,17 +1,12 @@
 import AppKit
 
 public class HistoryItem: Equatable, Codable {
-  public enum Types: String, Codable {
-    case string
-    case image
-  }
-
   public let value: Data!
   public var firstCopiedAt: Date!
   public var lastCopiedAt: Date!
   public var numberOfCopies: Int!
   public var pin: String?
-  public var type: Types!
+  public var types: [NSPasteboard.PasteboardType] = []
 
   public static func == (lhs: HistoryItem, rhs: HistoryItem) -> Bool {
     return lhs.value == rhs.value
diff --git a/Maccy/Menu.swift b/Maccy/Menu.swift
index 94140e9..ae95af8 100644
--- a/Maccy/Menu.swift
+++ b/Maccy/Menu.swift
@@ -222,11 +222,7 @@ class Menu: NSMenu, NSMenuDelegate {
   }
 
   private func copy(_ item: HistoryMenuItem) {
-    if item.item.type == .image {
-      clipboard.copy(item.item.value, .tiff)
-    } else {
-      clipboard.copy(item.item.value, .string)
-    }
+    clipboard.copy(item.item.value, item.item.types)
   }
 
   private func copyAndPaste(_ item: HistoryMenuItem) {
```

Defman21:
This would break `Codable` for some reason. Though it's a good idea, maybe worth trying to make it work with `Codable`.

Defman21:
I made it working by storing raw values of PasteboardType's (so it's `[String]`) and decode them to PasteboardType's on initialization.

p0deje:
> I made it working by storing raw values of PasteboardType's (so it's [String]) and decode them to PasteboardType's on initialization.

Have you tried to just add the protocol?

```swift
extension NSPasteboard.PasteboardType: Codable {
}
```

> Note: even if we're saving all types, we still have to have the if-else branch to determinate a single type for clipboard.setData(..., forType: ...). Maybe we could simplify it to storing only the type we're going to pass to setData?

Right. I totally forgot about this. Maybe we need to store every data for every type separately and then when copying add accordingly?

```swift
public var data: [NSPasteboard.PasteboardType: Data]

// ...

func copy(_ typesWithData: [NSPasteboard.PasteboardType: Data]) {
  pasteboard.declareTypes(typesWithData.keys, owner: nil)
  for (type, data) in typesWithData {
    pasteboard.setData(data, forType: type)
  }
}
```

Defman21:
Extending PasteboardType with Codable does not work unfortunately. `typesWithData` looks reasonably, gonna try that.

Defman21:
Why would we even want to set different data for different types though in context of one HistoryItem? I implemented that and I don't see any cases but it makes the code a bit complicated.

p0deje:
> Upd: nevermind. Seems like copying text produces a bunch of different PasteboardType's. I guess it might be good to support all of them and let macOS decide what to paste from the pasteboard. This might even fix #75.

Yeah, that's what I had in mind.

Defman21:
I'm a bit busy right now so I'm gonna push what I currently have. There's a bunch of TODO's left and some clean-ups to be made, gonna work on that later.

Defman21:
@p0deje got a bit of time. Currently this whole feature works like that:

1. If it's an image (.png, .tiff, .jpg, clipboard screenshot), it displays an actual image.
2. If it's a file, it:

2.1. Tries to display an Apple icon for the item (like in Finder).
2.2. Display a pin icon if there's no Apple icon. This one is something I don't like really, but I'm not sure what icon should we display in this case.

Your thoughts?

Defman21:
I also implemented path shortening if the path is too long, so `/Users/defman/Pictures/a_very_long_name_for_a_meme.jpg` becomes `/U/d/P/a_very_long_name_for_a_meme.jpg` (and it displays `File located in /Users/defman/Pictures` in the tooltip)

p0deje:
Sorry for the late reply. I've started seeing some performance issues for loading all this data from disk (which is done by UserDefaults). I couldn't find a way to make the loading faster so I actually took a stab at reworking the storage from UserDefaults to CoreData. The changes are in progress in https://github.com/p0deje/Maccy/tree/coredata branch. Loading/saving definitely became faster as now done in memory, but I still have a lot to finish before this can be merged.

However, the changes in CoreData mostly make this pull request obsolete as it's about the same. I'll probably take some code from here in regards to file copy handling and rendering.

My apologies for not notifying earlier as I see you've done a tremendous amount of work. Thanks for pushing the project in the right direction!

Defman21:
Okay, good to know! Thanks for notifying me. Should I close the PR?

p0deje:
Let's keep it open - I'll review and cherry-pick some changes maybe. I'll close.

p0deje:
@Defman21 Thanks again for all the hard work you did in the PR. I went through the code, taken some bits from there related to file copying and pushed 183f8be3d0965e2482e6604caa53111c1f9a23f0. I've added you as a co-author - hope you don't mind. I've decided not to implement a custom path shortener for now as search + tooltip works.

I believe the PR can be closed now as all the missing bits are in master.

Defman21:
I don't mind being a co-author :) Looking at the code though... Does it copy a string `file://...` and pastes it as a string? Or NSPasteboard handles it correctly and under the hood it converts the FileURL to an actual file?

p0deje:
Since now we store the whole contents of NSPasteboard item, it is exactly the same as copying the file. I believe it internally retrieves file from fileURL and pastes it.

I tested this by copying the file, then copying something, then selecting the copied file and pasting in Finder. The file was pasted to the current folder in Finder.

PR #101, opened 2020-05-14
Limit displayed menu size

huseyinyagli:
Hi,

I don't know Swift. I just used VSCode to propose a very little but useful feature.
Could you consider adding this feature?

### DIFF
--- a/Maccy/History.swift
+++ b/Maccy/History.swift
@@ -11,8 +11,10 @@ class History {
   }
 
   init() {
-    UserDefaults.standard.register(defaults: [UserDefaults.Keys.size: UserDefaults.Values.size])
-    UserDefaults.standard.register(defaults: [UserDefaults.Keys.menuSize: UserDefaults.Values.menuSize])
+    UserDefaults.standard.register(defaults: [
+      UserDefaults.Keys.size: UserDefaults.Values.size,
+      UserDefaults.Keys.menuSize: UserDefaults.Values.menuSize
+    ])
     if ProcessInfo.processInfo.arguments.contains("ui-testing") {
       clear()
     }
--- a/Maccy/History.swift
+++ b/Maccy/History.swift
@@ -13,7 +13,7 @@ class History {
   init() {
     UserDefaults.standard.register(defaults: [
       UserDefaults.Keys.size: UserDefaults.Values.size,
-      UserDefaults.Keys.menuSize: UserDefaults.Values.menuSize
+      UserDefaults.Keys.maxMenuItems: UserDefaults.Values.maxMenuItems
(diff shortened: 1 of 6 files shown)


p0deje:
Hi, thanks for the contribution!

I'm not sure what exactly this feature tries to achieve. Could you elaborate?

huseyinyagliudemy:
Hi p0deje. Sorry for the lack of explanations.
When using maccy, I want to have a very large clipboard history since I'm copying lots of things. But I don't want to have a menu which covers the entirety of the screen when it pops up.
So I tried to add a MAX_MENU_SIZE to settings and limit the size of the popup menu to this size without reducing history size.

Maccy/History.swift
```diff
@@ -11,6 +11,7 @@ class History {
 
   init() {
     UserDefaults.standard.register(defaults: [UserDefaults.Keys.size: UserDefaults.Values.size])
+    UserDefaults.standard.register(defaults: [UserDefaults.Keys.menuSize: UserDefaults.Values.menuSize])
```
p0deje, line 14: You should join this into a single statement:
```suggestion
    UserDefaults.standard.register(defaults: [
      UserDefaults.Keys.menuSize: UserDefaults.Values.menuSize
      UserDefaults.Keys.size: UserDefaults.Values.size,
      ])
```

huseyinyagli:
> Thanks for the proposal! I don't think it brings much value since functionally it works the same as `size` option. However, I believe you can improve this to bring more value:
>     * All the items that don't satisfy menu size are hidden, not removed
>     * All hidden items can be searched

That's what I was trying to do. I just figured out how you were managing your historyItem <-> menuItem logic.

p0deje:
Does it now work with search? Can it find invisble items?

huseyinyagli:
> Does it now work with search? Can it find invisble items?

Hi Alex. Yes, it works with search. It can find invisible items.

rayshan:
Thanks @huseyinyagli, +1 on this feature.

huseyinyagli:
@p0deje  Hey Alex. Is there anything you want me to do on this pull request?

p0deje:
No, you did a great job. Thank you! I'm going to merge it as-is and then apply some changes:

1. Set default to 0.
2. Make 0 turn off this setting so that it means that all stored items will be visible.
3. Smaller style improvements.

p0deje:
I've squashed all in 201e1d6d6db0d9c8e0996d686509a807bfe3dadb and merged. Thank you once again!

huseyinyagli:
Thank you for improving merging the feature. Happy to help in such a useful project.

PR #141, opened 2020-08-02
Add display of prefix and suffix for clipboard entry

### LINKED ISSUE #139: Show suffix of clipboard entry, too
When the clipboard entry is too long, Maccy shows me the prefix.  But, especially in the case of file names, the prefix alone is not that helpful.  What do you think about showing a prefix, then an ellipsis, then a suffix, of each entry?

So that `/Users/jrl/Documents/Projects/2020/05/Corona-Strategy.docx` would not be abbreviated `/Users/jrl/Documents/Projects/...`, but rather `/Users/jrl/.../Corona-Strategy.docx`.  (I didn't could characters for this example, but I hope you get the idea.)

thekaleidoscope:
This commit will show both prefix and suffix of an entry in clipboard
by dividing the string's input to show prefix to half max length and similarly
show the suffix for another half max length. The text in between is
truncated and replaced with ellipsis.

Fixes #139

### DIFF
--- a/Maccy/HistoryMenuItem.swift
+++ b/Maccy/HistoryMenuItem.swift
@@ -123,8 +123,10 @@ class HistoryMenuItem: NSMenuItem {
       return string
     }
 
-    let index = string.index(string.startIndex, offsetBy: maxLength - 1)
-    return "\(string[...index])..."
+    let halfMaxLength: Int = maxLength / 2
+    let indexStart = string.index(string.startIndex, offsetBy: halfMaxLength - 1)
+    let indexEnd = string.index(string.endIndex, offsetBy: -halfMaxLength)
+    return "\(string[...indexStart])...\(string[indexEnd...])"
   }
 
   private func defaultTooltip(_ item: HistoryItem) -> String {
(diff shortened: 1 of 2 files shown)


p0deje:
Thank you for your contribution!

Can you show any screenshots? What do you think if _..._ is moved more towards the end - for example, we show 40 chars at the beginning and 10 chars at the end?

thekaleidoscope:
@p0deje sure, but I found 50% for prefix and suffix was serving better as we can't say with certainty giving 80% of total length to prefix will be more beneficial in all cases, and there are plenty of examples like URL or key pass where getting both prefix and suffix equal weightage serves it better. However, if you are certain 4:1 will be the right choice, then sure, I'll update the code with it.

Screenshot:
Screenshot 2020-08-03 at 5 04 13 PM

thekaleidoscope:
Hi @p0deje, let me know your thoughts, I can move it to the 4:1 ratio like you said

p0deje:
@thekaleidoscope It's great, thanks for sending PR. I'll merge, play around with ratios and see if they need any adjustment.

p0deje:
I like 2:1 ratio more as I typically search by the beginning of the text - not the end. Does it look good for you?
Screen Shot 2020-08-06 at 11 52 32

thekaleidoscope:
Yes, it looks much better tbh

p0deje:
Then it's settled https://github.com/p0deje/Maccy/commit/4154e1a1e8582ada4533f55fb3119fc3bf338506

PR #147, opened 2020-08-10
Added KeeWeb in default app exclusion

seupedro:
KeeWeb is very popular. It deserves to be also in the exclusion list 

https://github.com/keeweb/keeweb

image

### DIFF
--- a/Maccy/AppDelegate.swift
+++ b/Maccy/AppDelegate.swift
@@ -88,7 +88,8 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         "de.petermaurer.TransientPasteboardType",
         "com.typeit4me.clipping",
         "Pasteboard generator type",
-        "com.agilebits.onepassword"
+        "com.agilebits.onepassword",
+        "net.antelle.keeweb"
       ]
       UserDefaults.standard.migrations["2020-04-25-allow-custom-ignored-types"] = true
     }
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ or overwritten:
 * `com.typeit4me.clipping`
 * `de.petermaurer.TransientPasteboardType`
 * `Pasteboard generator type`
+* `net.antelle.keeweb`
 
 You can add additional custom types using preferences or `defaults`:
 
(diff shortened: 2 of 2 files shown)


Maccy/AppDelegate.swift
```diff
@@ -88,7 +88,8 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         "de.petermaurer.TransientPasteboardType",
         "com.typeit4me.clipping",
         "Pasteboard generator type",
-        "com.agilebits.onepassword"
+        "com.agilebits.onepassword",
+        "net.antelle.keeweb"
```
p0deje, line 92: To add KeeWeb, you need to create a brand new migration with a separate key (`2020-08-11-add-keeweb-to-ignored-types`.\)

seupedro: not sure what you mean

p0deje: What I mean is that there is this guard to make sure we don't run migration more than once:

```swift
if UserDefaults.standard.migrations["2020-04-25-allow-custom-ignored-types"] != true {
  # migration code
  UserDefaults.standard.migrations["2020-04-25-allow-custom-ignored-types"] = true
}
```

Anyone who has upgraded Maccy since April already has this migration run, so it won't be applied. It's easy to check what migrations are run with this following command:

```
$ defaults read org.p0deje.Maccy migrations
{
    "2020-02-22-history-item-add-copied-at" = 1;
    "2020-02-22-history-item-add-number-of-copies" = 1;
    "2020-04-18-switch-storage-to-core-data" = 1;
    "2020-04-25-allow-custom-ignored-types" = 1;
    "2020-06-19-use-keyboardshortcuts" = 1;
}
```

So if we want to add KeeWeb to all users when they upgrade Maccy in the next version, we need a new migration. Something like this should do the trick:

```swift
if UserDefaults.standard.migrations["2020-08-11-add-keeweb-to-ignored-types"] != true {
  UserDefaults.standard.ignoredPasteboardTypes =
    UserDefaults.standard.ignoredPasteboardTypes.union(["2020-04-25-allow-custom-ignored-types"])
  UserDefaults.standard.migrations["2020-08-11-add-keeweb-to-ignored-types"] = true
}
```

p0deje:
Merged in cd60574, thank you!

PR #149, opened 2020-08-17
Enable Maccy to prompt for a11y permissions for pasting automatically.

shadanan:
This PR updates the Maccy entitlements so that Maccy prompts the user
for Accessbility permissions when attempting to paste immediately.

Sandboxed apps apparently cannot prompt for accessbility permissions:
https://developer.apple.com/forums/thread/24288

maccy-a11y-prompt

### DIFF
--- a/Maccy/Maccy.entitlements
+++ b/Maccy/Maccy.entitlements
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>com.apple.security.app-sandbox</key>
-	<true/>
+	<false/>
 </dict>
 </plist>


p0deje:
First of all, apologies for such a slow response - I was on vacation and haven't been checking my email.
Second, it's ridiculous that Apple prevents accessibility when the sandbox is enabled. I had no idea!
Third, I still need to keep the sandbox for the future distribution in the App Store, so your solution can't be merged. I'll search for other possible workarounds.

shadanan:
Sounds good. Feel free to close.

p0deje:
Fixed via ad5107525c11616ce9c4000cb88bc7328a5f3d5a

PR #154, opened 2020-08-24
Simplified Chinese Support

simonla:


p0deje:
Thank you for the PR! I'm definitely going to merge this, but it would be great if you could add 4 more translations that were added since you opened PR. I'm talking about recent changes in https://github.com/p0deje/Maccy/blob/master/Maccy/en.lproj/Localizable.strings#L21-L24 which show accessibility permissions window when pasting. Could you add these 4 translations to simplified Chinese file?

simonla:
done

p0deje:
Thank you!

Issue #144, opened 2020-08-07
Maccy window steals focus and pasting fails


C-Saunders:
The issue seemed to start when I updated to 0.14.0 earlier today. I'm running MacOS 10.15.6.

I installed 0.13.0 (by downloading the zip file, I usually install via homebrew, but you can't install specific versions of casks) and I am now able to paste from Maccy as expected again.

Thanks for an excellent tool and please let me know if you need any more information I can provide!

p0deje:
Maccy should now steal focus, that's expected. However, it has to return focus when the window is closed so pasting should work just fine.

Can you explain step-by-step what exactly is happening?

C-Saunders:
> Does it fail to paste for all applications or a specific one?

It seems to be happening for all applications. I tried Chrome, VS Code, TextEdit, Slack, and iTerm.

> Can you explain step-by-step what exactly is happening?

- Put my cursor somewhere I can enter text
- Hit hotkeys (cmd + shift + v)
- Maccy window pops up (I have it set to screen center)
- I select an item (mouse, enter key, cmd+#, doesn't seem to change behavior) 
- Focus doesn't return to where it previously was so pasting fails
- The selected item is now the paste-able item and it works if I put focus back where it was and paste normally

Thanks!

p0deje:
Also, any chance you can record a screencast showing the issue? Maybe I can spot something...

C-Saunders:
> Can you try to remove Maccy from "System Preferences" -> "Security & Privacy" -> "Accessibility" and try again? It should ask for permission and work after you give it.

I did this. It asked for permission, I granted it, no change :(

> Can you post the output of `defaults read org.p0deje.Maccy`

```js
{
    "KeyboardShortcuts_popup" = "{\\"carbonModifiers\\":768,\\"carbonKeyCode\\":9}";
    "NSWindow Frame SUUpdateAlert" = "970 768 620 392 0 0 2560 1417 ";
    "NSWindow Frame com.sindresorhus.Preferences.FrameAutosaveName" = "968 502 550 540 0 0 2560 1417 ";
    SUEnableAutomaticChecks = 1;
    SUHasLaunchedBefore = 1;
    SULastCheckTime = "2020-08-07 14:59:04 +0000";
    SUSendProfileInfo = 0;
    SUUpdateRelaunchingMarker = 0;
    fuzzySearch = 1;
[...]

Issue #129, opened 2020-06-27
Request:support chinese type in search field


Ti800:
im so lucky that find the incrediable app.❤

Ti800:
it seems like a bug that cant activate typing mode correctly.
while opening the setting,its working again.
<img width="416" alt="截屏2020-06-27 下午4 54 09" src="https://user-images.githubusercontent.com/53323379/85918689-91911b80-b897-11ea-8f0c-5aaf92bf5cc3.png">

p0deje:
I'm sorry, but what is the exact issue? Can you give me any steps to reproduce?

Ti800:
![屏幕录制2020-06-28 上午8 45 47](https://user-images.githubusercontent.com/53323379/85934892-19b90480-b91c-11ea-8df7-c0c704134b10.gif)
1.Open app, if you are typing in alphabet,it would be ok even without typing mode.
![屏幕录制2020-06-28 上午8 58 54](https://user-images.githubusercontent.com/53323379/85935016-13c42300-b91e-11ea-8db8-0bb07d7feb94.gif)
2.i was forced to quit  typing mode when typing space.

p0deje:
I've started an attempt to make it work but got into a rabbit hole. Here are some observations on what I did and what are the problems:

1. Ideally, instead of manually appending the `queryField.stringValue` from Carbon event we receive, we'd just make `queryField` first responder and let the event be handled by AppKit. This would automatically support custom input popovers like Chinese or when one press `e` for few seconds.
2. It's not possible to make `queryField` first responder because menu bar apps are _not_ activated when menu is opened. From AppKit perspective, such apps ignore all the events.
3. To make `queryField` first responder, we can force-activate the app when menu opens. When activation happens, the menu is immediately closed.
4. To workaround the issue, we can activate and popup menu using `DispatchQueue.main.async`. For some reason, it works.
5. Even with all these hacks, we have to activate a previous application after menu is closed. So far I haven't found a way to make it work with pasting functionality.

Some progress can be seen in 6d1f144e0afc1502a71cebf8af58d7bd0e9e2323, but I am no longer sure I can make it work reliably. I've already spent hours piling one hack over another 😞

nemoii:
I need this feature too, thanks so much @p0deje

p0deje:
Released in 0.14.0

PR #180, opened 2020-09-30
Add preference to avoid taking app focus

pwxn:
Changes to focus in #129 to enable better text editing are great, and helpful for many users.

However, there are some use cases where I'd rather sacrifice system text features to not take focus away from the current app. Here's an example:
1. Maccy is configured to paste automatically 
2. A text editor (such as vsCode) is configured to format when it loses focus. 
3. Activate Maccy, editor loses focus and removes trailing spaces from the current line (or other code linting/formating)
4. Paste your text from Maccy, but not where you intended due to the formatting.  

Text from the advanced preferences page sums it up:
> By default, Maccy will take the application focus while the popup is displayed. Maccy needs app focus to allow advanced text handling in the search bar. If you enable this, you will only be able to search for basic ASCII. However, certain situations (such as a text editor that auto-formats when it loses focus) may make this desirable.

related issue: #144 with more use cases in the comments(Hotkey Window, Context menus)

### DIFF
--- a/Maccy/FilterMenuItemView.swift
+++ b/Maccy/FilterMenuItemView.swift
@@ -240,6 +240,10 @@ class FilterMenuItemView: NSView, NSTextFieldDelegate {
     if let chars = event.charactersIgnoringModifiers {
       if chars.count == 1 {
         focusQueryField()
+        if UserDefaults.standard.avoidTakingFocus {
+            setQuery("\(queryField.stringValue)\(chars)")
+            return true
+        }
       }
     }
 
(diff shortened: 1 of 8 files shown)


p0deje:
Thanks for the PR! I'll need to test it locally, but I can already mention that this preference needs localization (Russian, Spanish, Chinese). I can definitely add Russian, but if you add something simple for the other two, that would be great.

pwxn:
Okay! Is google translate good enough? I don't know any Chinese

p0deje:
Hope you don't mind - I did some code and text changes. Merged in 4102d7b.

Thanks for PR!

pwxn:
Thanks!

PR #213, opened 2020-11-24
Fix icon size for Big Sur

moriwang:
Hi Maccy team,

I found the icon is larger than other apps in Big Sur. So I used [the official icon template](https://developer.apple.com/design/resources/#macos-apps) and `Designs/Icons.sketch` to resize icons. But they look a little different in color.

Before: 
截屏2020-11-25 上午12 27 21

After:
截屏2020-11-25 上午12 30 46

This commit contains these resized icons and the icon template sketch file.

Since I knew nothing about Xcode, so it's not built and tested yet.


moriwang:
The color difference may caused by color profile. The icon template provided by Apple is P3 by default. But the origin icon may be sRGB.
More info: https://www.sketch.com/support/troubleshooting/color-management/

p0deje:
Huh, I already used the icon template from Apple, guess I did something to make it export bigger than it should be. The colors need to be fixed because it looks very much different (shadows, lines thickness). Do you believe you make the icon look exactly as before but just smaller?

moriwang:
> Huh, I already used the icon template from Apple, guess I did something to make it export bigger than it should be. The colors need to be fixed because it looks very much different (shadows, lines thickness). Do you believe you make the icon look exactly as before but just smaller?

I just copied the icon from `Designs/Icons.sketch` and paste into the template. I'll figure out what happened.

moriwang:
@p0deje  

I tried my best, but the color is still not exactly the same. Please fix this if you have time.

What I tried this time:
1. switch the color profile to the same (tried Unmanaged, sRGB and P3).
2. correctly resize the icon.

image

Johnderate:
Is this still on the agenda? If the icon from this PR is not ready for merging, should I open a new issue for this?

p0deje:
It's fixed in 16ab3220cbc0c0cb195bd3fc4e8a87f109bb6a95 and released in 0.19.0

Johnderate:
Thank you, it is indeed fixed. I had to update manually, the auto-update didn't show the version, that's why I missed it.

PR #242, opened 2021-02-20
Private Mode Feature

onurpolattimur:
First of all, I would like to thank you for creating such a good open source project. I use Maccy very actively. But every once in a while I need to turn it off for privacy. That's why I wrote a shell script and assigned a key shortcut with automator. I have found it very useful so I want to add it as a feature to Maccy.  

How it works: 
- Default shortcut key is: cmd +  shift + x
- When pressed these keys it will enter private mode and for 60 seconds it won't record the copied items to history. Since you already have icon changing on turn off the user can see wether or not is in private mode. 
- If you press these keys again within 60 seconds it will disable the private mode without waiting for time to run out. 

By the way I don't have any Swift experience, I tried to follow your structure. If you want anything changed don't hesitate on telling me. 

maccy_private_mode

### DIFF
--- a/Maccy/AppDelegate.swift
+++ b/Maccy/AppDelegate.swift
@@ -18,7 +18,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     migrateUserDefaults()
 
     maccy = Maccy()
-    hotKey = GlobalHotKey(maccy.popUp)
+    hotKey = GlobalHotKey(maccy.popUp, maccy.togglePrivateMode)
   }
 
   func applicationShouldHandleReopen(_ sender: NSApplication, hasVisibleWindows flag: Bool) -> Bool {
--- a/Maccy/GlobalHotKey.swift
+++ b/Maccy/GlobalHotKey.swift
@@ -7,10 +7,14 @@ class GlobalHotKey {
   static public var key: KeyboardShortcuts.Key? { KeyboardShortcuts.Shortcut(name: .popup)?.key }
   static public var modifierFlags: NSEvent.ModifierFlags? { KeyboardShortcuts.Shortcut(name: .popup)?.modifiers }
 
-  private var handler: Handler
+  private var popupHandler: Handler
+  private var privateModeHandler: Handler
 
-  init(_ handler: @escaping Handler) {
-    self.handler = handler
-    KeyboardShortcuts.onKeyDown(for: .popup, action: handler)
+
+  init(_ popupHandler: @escaping Handler, _ privateModeHandler: @escaping Handler) {
+    self.popupHandler = popupHandler
+    self.privateModeHandler = privateModeHandler
+    KeyboardShortcuts.onKeyDown(for: .popup, action: popupHandler)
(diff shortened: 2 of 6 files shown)


jgoldhammer:
I like this feature. I would definitely like to see it in the app...

LiorElb:
Would love to have this feature - seems super cool!

p0deje:
Thank you very much for taking the time and proposing this PR. Sorry for me being not responsive - I've tried to approach reviewing it few times but always got distracted and never finished.

I like the idea overall, but I feel that we create 2 competing features within the app. First is ability to ignore all new copies, which can be activated using `defaults`, via preferences or by option+clicking the menubar icon. Second - this private mode feature. Both serve the same goal - ignore all new copies for some time. Maybe there is a way to merge these features anyhow? Would option+click be enough for most cases? Or maybe we can enable private mode by shift+option+click on the menubar icon?

As for the implementation, I'm concerned with using global hotkey. I would like to avoid messing with people's global hotkeys, so I think there should be no defaults at all and this feature should be opt-in.

60 seconds seems like an arbitrary value and maybe it should be configurable? Maybe we even need to create a new preference pane called "Private" that allows to disable Maccy, configure private mode hotkeys and delay, etc.

j-dogcoder:
> 60 seconds seems like an arbitrary value and maybe it should be configurable? Maybe we even need to create a new preference pane called "Private" that allows to disable Maccy, configure private mode hotkeys and delay, etc.

I agree.

pdf13:
IMO the solution proposed by @onurpolattimur has a key point that Maccy does not address with his current solution: a keyboard shortcut to ignore new copies.

I love the idea of being able to turn it on and off without having to using a mouse/trackpad to position my pointer on a small space and click it, do my secret copy/paste and having to click again to return. Using a keyboard is so much faster, that I think eliminates the need of having a timer to restore the previous state.

Is this make sense to you folks?

seivan:
Pasting here for brevity. 

There are a few issues.

- Time based stuff rarely works out, and people keep toggling to give themselves more time.
- `option+click` is the wrong method for disabling, it should be `cmd+click` - [Related](https://github.com/p0deje/Maccy/issues/355)
- There's a bug right now that you can't re-enable by doing the same `option+click` once it's been disabled. 

My suggestions are allow toggling enabled/disabled state via `cmd+click`
Add keyboard shortcut for disabling Macy **only** for the next item in the pasteboard, this has the added support for not just text but also things that come externally, like say `Screenshot to pasteboard` or (not sure if it works) stuff copied from iOS via handover.

p0deje:
In the new version of Maccy, there will be a way to disable the next copy with Option+Shift+Click on the menu icon (see dc2f33e1e3cfe73ffa414e795dc4a6b1e19ac759 for more details). I'd like to close this PR as we haven't really settled on any particular decision and I don't want stale PRs hanging around. We can create a new issue to discuss the idea of private mode and how it should be designed.

@onurpolattimur Once again, thank you for the proposal!

Issue #260, opened 2021-03-27
Maccy saves clipboard contents after disabling "ignoreEvents" in plist


LiorElb:
Hey,
I've noticed an issue with the "Ignore Copied Items" mechanism specified in the README.md file.

Scenario is:
1. set "ignoreEvents" to true
2. copy and paste something sensitive
3. set "ignoreEvents" to false

After step 3 the sensitive data you copied while Maccy was disabled will be **back in the Maccy storage**.

Also - Iv'e noticed there is a pull request open that introduces `Private Mode`.  #242 
 I believe the same issue will happen there after the private session times out.

Update:
This behaviour also reproduces when using the `Turn Off` button in the Preferences->Advanced tab

yunus776:
> Hey,
> 
> I've noticed an issue with the "Ignore Copied Items" mechanism specified in the README.md file.
> 
> 
> 
> Scenario is:
> 
> 1. set "ignoreEvents" to true
> 
> 2. copy and paste something sensitive
> 
> 3. set "ignoreEvents" to false
> 
> 
> 
> After step 3 the sensitive data you copied while Maccy was disabled will be **back in the Maccy storage**.
> 
> 
> 
> Also - Iv'e noticed there is a pull request open that introduces `Private Mode`.  #242 
> 
>  I wonder if the same issue happens there after the private session times out.
> 
> 
> 
>

p0deje:
Should be fixed in #261 and available in the next released.

PR #261, opened 2021-03-28
Fix saving contents that are copied when ignoreEvents is enabled

onurpolattimur:
Fix for issue: #260


p0deje:
Thank you for the fix! I wonder how tests didn't catch that, but it's been a while since they worked properly.

PR #275, opened 2021-04-12
Update Chinese Translation

Tai-Zhou:
Update Chinese translation and add a keyboard indicator to `README.md` to make it clearer. I didn't realize that the comma <kbd>,</kbd> should be pressed and have spend more than an hour to figure out why long press <kbd>⌘</kbd> doesn't work.😣

Thank you for your great app!


simonla:
Sorry about this. I translate it from "Press ⌘, to open preferences when footer is hidden". Should we fix the original text as well?

p0deje:
@simonla Sorry, I don't get what's wrong. Can you explain in more details?

Issue #198, opened 2020-11-14
Feature request: Clear history on shutdown/reboot


mandrasch:
Thanks again for providing this very useful software as open source!

I would really like the optional setting of clearing the history on shutdown or reboot. 

Background: Sometimes copied passwords or other sensitive data is still in my history, therefore clearing it on shutdown/reboot would be a good protection. (I know that there is the option of pasteboard types for password managers which can be ignored, great feature)

Thanks in advance!

Fongeme:
It sounds like the more appropriate solution would be to delete the data on app shutdown and not on pc shutdown?
If that sounds good, i'd like to get going on this :)

Fongeme:
I've created #292

mandrasch:
Nice, thanks very much @Fongeme! (Don't have the skills to test the PR, hope it all goes well :-))

PR #292, opened 2021-06-25
Delete history on app shutdown

Fongeme:
#198


p0deje:
Thank you for the PR! I've had to do plenty of changes to make it work, but I've co-authored you on the commit.

PR #333, opened 2021-11-20
Improve zh-Hans translation

kidonng:
There is some error and improvements to be made to the zh-Hans translation. The text is a little bit much so I'm doing this in parts.


p0deje:
Thank you very much for your contribution!

PR #363, opened 2022-01-17
Updated link for downloads icon

ozziekhoo:
Not sure if there was meant to be a link on the downloads icon


p0deje:
It was meant to be just a badge to show a number of downloads, but your changes make it even better. Thanks!

PR #370, opened 2022-02-09
Adding German Localization

pitschr:
BEWARE: I do not have MacOS development environment. But I translated the text based on English translation template. So it would be good if you can do basic check before doing it public. I can take a 2nd round of localization check before publishing to AppStore.


p0deje:
Thank you very much for the contribution! Merged in 7d0aaac.

PR #434, opened 2022-07-19
Fixed showing popup window on center of the screen when working with multiple monitors

boogiedimik:
This is my first contribution ever, hope i did it right.

I actively use the application, but when working with two monitors when using the settings for displaying a pop-up window on the center of screen, in my opinion, it worked illogically: the window did not open on the active monitor where i working, but on the one marked in the system settings as the main monitor.

Fixed this for myself and i would like to share the fix.

### DIFF
--- a/Maccy/Maccy.swift
+++ b/Maccy/Maccy.swift
@@ -88,7 +88,8 @@ class Maccy: NSObject {
     withFocus {
       switch UserDefaults.standard.popupPosition {
       case "center":
-        if let screen = NSScreen.main {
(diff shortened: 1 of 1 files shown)


p0deje:
[`NSScreen.main`](https://developer.apple.com/documentation/appkit/nsscreen/1388371-main) should return the window with the focused application. It does not track mouse cursor location so you can have one screen with cursor and another with the focused app. In this case, Maccy would be shown on the latter.

Given Maccy is keyboard-driven, I think it should stay the way it's implemented. Thank you for the proposal though!

PR #477, opened 2022-10-17
Improve handling of selected text in the query field

ChaosCoder:
This PR improves the handling of selected text in the query field in the following cases:

- When typing text, the selected query text will be replaced with the new text.
- When hitting the backspace key, the selected query text will be removed.
- When pasting text, the selected query text will be replaced by the new text.

This gives a bit more system feel to the query field.

### DIFF
--- a/Maccy/MenuHeader/MenuHeaderView.swift
+++ b/Maccy/MenuHeader/MenuHeaderView.swift
@@ -165,9 +165,8 @@ class MenuHeaderView: NSView, NSSearchFieldDelegate {
       performMenuItemAction(MenuFooter.clearAll.rawValue)
       return true
     case .deleteOneCharFromSearch:
-      if !queryField.stringValue.isEmpty {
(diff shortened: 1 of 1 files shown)


ChaosCoder:
Hey @p0deje, is there anything I could do to get this accepted?

p0deje:
@ChaosCoder I'm a bit slow to review now, but I'll look through it within the next couple of weeks.

ChaosCoder:
Sure, take your time! :+1: I just hoped it could be accepted for hacktoberfest, but no worries 🙂

p0deje:
I've tried pulling this to test, but I don't think it behaves correctly. For example, moving the input cursor and pressing delete does nothing. Also, when I select text and then remove it by pressing delete, the selection should be dropped, but it is not.

Could you please test it more thoroughly and address at least these two issues?

p0deje:
Closing since there is no response. Feel free to re-open once the comments are addressed.

PR #501, opened 2023-01-04
Translation correction

fabian19vs:
made a correction on the translation, in spanish it was "Borrar el historial al dejar de fumar" ("Clear history when quit smoking"), therefore the purpose of the button was not clear.


p0deje:
Thank you @fabian19vs !

PR #539, opened 2023-03-29
Simplify brew command for installation

marsavela:
Latest versions of `brew` allow you to skip the cask argument and will install just fine. 👍


p0deje:
Thank you!

marsavela:
Great app. Thanks for it.

Issue #566, opened 2023-06-22
On MacOS 14 Beta 2 when opened Maccy freezes and disables keyboard until force quit


victor-yunenko:
Firstly thank you for the awesome clipboard manager!

I'm creating this ticket mostly for other people's reference, since this beta has been quite popular. Understandably wouldn't expect it be fixed soon, given how unstable the beta is by itself.

Reproduction: triggering Maccy's shortcut or clicking on it in the Menu Bar stops most of the keys from working. Afterwards a restart or force quitting Maccy through Activity Monitor helps.

Versions:
- 14 Beta 2 (23A5276g), released on Jun 21st
- Maccy 0.26.4 from homebrew

notcgi:
also unable to pin

BSakura-Miku:
在MacOS 14 Beta 2、1 出现相同问题

howardellacott:
I think this is a macOS bug as the same occurs with Little Snitch, Tailscale and Amphetamine

p0deje:
I can reproduce the issue, but no idea how to fix it yet.

xezrunner:
I fixed the app hanging and crashing in https://github.com/p0deje/Maccy/pull/569.
However, at least during debugging, the clipboard contents don't appear to get recorded. I think that is unrelated - perhaps some clipboard API has changed with Sonoma Beta 2.

xezrunner:
> I can reproduce the issue, but no idea how to fix it yet.

Debugged it a bit, the crash appears to be linked to modifying the frame of the header `NSView` in the menubar panel in any way.
Seems to be an OS-side crash though, might even be specific to `NSView`s inside `NSMenuItem`s. Should probably report this to Apple.

philihub:
Thank you for working on this! Would there be the possibility for a quick fix for the basic functionality of Maccy so it is at least usable again despite some features not working?

xezrunner:
> Thank you for working on this! Would there be the possibility for a quick fix for the basic functionality of Maccy so it is at least usable again despite some features not working?

If you are relying on searching within Maccy, I suggest you wait until the next macOS Beta and/or until the Maccy developer releases an update fixing it.
Right now, it looks like searching by just typing after summoning Maccy is broken on Sonoma Beta 2.

If you still want the patch:
You can grab my fork and build the branch with the patch. If you don't have Xcode, here's a precompiled build:
[...]

PR #569, opened 2023-06-25
Fix macOS Sonoma Beta 2 menubar hang/crash

xezrunner:
This fixes the hang and crash that occurs since macOS Sonoma Beta 2. (https://github.com/p0deje/Maccy/issues/566)

Maccy used to hide the "Maccy" title and search bar within the menu by setting the height of the NSView to a small value.

In Beta 2, modifying the frame in any way results in a crash.

While this is likely an OS bug, since setting the height to 1 is a hack anyway, and the header only contains the title and search bar, not adding the header when both are disabled seems like a cleaner solution to me.

### DIFF
--- a/Maccy/Maccy.swift
+++ b/Maccy/Maccy.swift
@@ -174,9 +174,7 @@ class Maccy: NSObject {
     clipboard.onNewCopy(updateMenuTitle)
     clipboard.startListening()
 
-    populateHeader()
-    populateItems()
-    populateFooter()
+    rebuild();
 
     updateStatusItemEnabledness()
   }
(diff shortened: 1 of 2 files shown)


p0deje:
Does search still work fine with this patch when hideSearch is enabled?

xezrunner:
> Does search still work fine with this patch when hideSearch is enabled?

Am I supposed to be able to search just by summoning the app and typing, without focusing the search field first?
While I've been using Maccy, I haven't been needing the search feature, so I don't know if this worked prior to my change.

In both scenarios (search field on/off), I can't just type when I summon Maccy, through either the menubar, nor the shortcut.
Instead, it selects the item that starts with that letter.

xezrunner:
This breaks searching when the search field is hidden. Perhaps unlinking the searching functionality from the queryField itself would be beneficial.

Closing this, as the solution is incorrect.

xezrunner:
> Does search still work fine with this patch when hideSearch is enabled?

Logically, it unfortunately does not in that scenario :(

However: it actually does not work even if the search field is enabled. The panel selects the item that starts with the letter being pressed.
I can see that `processSingleCharacter()` is called by `processKeyDownEvent()` whenever a character is pressed, suggesting that all input should be forwarded to the `queryField` search field control.
For whatever reason, this does not get called on macOS Sonoma Beta 2 in my case for some reason.

PR #570, opened 2023-07-01
Add explain for pin item

Agenric:
The software is incredibly useful, but during my usage, I noticed that when I have multiple similar entries pinned, I occasionally forget what they originally represent. Hence, I have submitted this pull request to address the issue.
To assist you in better understanding, here are two relevant screenshots showcasing this feature in action:
Xnip2023-07-01_11-04-08
Xnip2023-07-01_11-03-29

### DIFF
--- a/Maccy/HistoryItem.swift
+++ b/Maccy/HistoryItem.swift
@@ -35,6 +35,7 @@ class HistoryItem: NSManagedObject {
   @NSManaged public var numberOfCopies: Int
   @NSManaged public var pin: String?
   @NSManaged public var title: String?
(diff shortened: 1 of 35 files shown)


p0deje:
Wouldn’t renaming the pin be helpful in this case? You can provide a new title for a pin in preferences without changing its contents.

Agenric:
Oh, apologies for the misunderstanding. I now realize that the title and content of the pinned entries are indeed separate. Given this clarification, the pull request loses its significance, and I will proceed to close it. 
Thank you. @p0deje

PR #586, opened 2023-07-23
Align popover with the selected menu item

weisJ:
This is done by finding the preceding visible view and computing from it the bounds of the selected menu item in window coordinates.

### DIFF
--- a/Maccy/Menu.swift
+++ b/Maccy/Menu.swift
@@ -93,9 +93,11 @@ class Menu: NSMenu, NSMenuDelegate {
     offloadCurrentPreview()
 
     guard let item = item as? HistoryMenuItem,
-          let indexedItem = indexedItems.first(where: { $0.menuItems.contains(item) }) else {
+          let itemIndex = indexedItems.firstIndex(where: { $0.menuItems.contains(item) }) else {
       return
     }
+
+    let indexedItem = indexedItems[itemIndex]
 
     if let previewMenuItemView = indexedItem.previewMenuItem.view,
        previewMenuItemView.frame.width != menu.size.width {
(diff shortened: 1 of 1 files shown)


p0deje:
This works nicely for small copies, but looks like the popup is misplaced when the preview is long, see this example:
Screenshot 2023-07-24 at 17 38 49

weisJ:
It's actually not a problem with too large items, but with filtered items not being ignored properly. Should be fixed now.

weisJ:
With the latest commit I have also added a small gap between the menu and the preview popover. I think it looks a bit nicer this way. If you don't think this is appropriate as a default I think this could be an user controllable setting.

p0deje (commented):
I suppose you accidentally removed all the icons - can you please revert? I will merge afterwards.

weisJ:
Yes that wasn't meant to be part of the PR :D I have some troubles with the settings icons not being properly sized on my external monitor and was testing some stuff to find out what is causing it (haven't found anything yet).

p0deje:
It looks really solid, there is just one small thing I am not sure about. If you disable the search field completely, and open Maccy via menu icon, the first preview points lower than it should. I suppose that's because the menu bar is above so it compensates for its offset, but I wonder if it can be fixed:

Screenshot 2023-07-25 at 08 17 07

weisJ:
I noticed the same thing. However I wasn’t able to find a good solution for this. The bounds are computed correctly in this case but the as you mentioned the menubar pushes the popover down as the request for the correct position can’t be fulfilled. An option would be to hide the arrow in this case, as I am not sure how one would force it to be located higher up.

p0deje:
In any event, it works like that in master too, so this PR definitely makes things much better, if not perfect. Thank you for the contribution!

PR #590, opened 2023-07-27
Prevent menu from moving while scrolling

weisJ:
If scrollbars are visible scrolling in the menu will cause it to move upwards which feels a bit weird.
This solution uses the `confinementRect(for:on:)` method of `NSMenuDelegate` to limit the menu to its original bounds it had when shown. To determine this size the menu is shown and immediately hidden just after querying the window bounds for the popup.

This technique is really dirty however I did not find a better solution without reinventing menu positioning. Therefore this is merely a suggestion :)

### DIFF
--- a/Maccy/Maccy.swift
+++ b/Maccy/Maccy.swift
@@ -103,7 +103,7 @@ class Maccy: NSObject {
       case "center":
         if let frame = NSScreen.forPopup?.visibleFrame {
           self.linkingMenuToStatusItem {
-            self.menu.popUp(positioning: nil, at: NSRect.centered(ofSize: self.menu.size, in: frame).origin, in: nil)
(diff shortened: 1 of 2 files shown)


p0deje:
I tried testing it and there is a small issue that sometimes you can see the first menu opening and hiding immediately. Is there any way to work it around?

weisJ:
I feared that that may be a problem. It didn’t happens for me but I was aware of the possibility. Sadly I don’t know yet how one would prevent this from happening. `NSMenu` blocking really is the blocker in this regard.

I won’t have time to work on this in the next month so I’ll close the PR. Once I find a different solution I’ll reopen :)

p0deje:
Thanks for trying!

Issue #599, opened 2023-08-13
Add support to configure update duration when searching in the clipboard


martikyan:
When typing in the search bar, the list with filtered items updates after about one second. Would be nice to have an option to configure that duration.

[Good first issue] [Feature request]

PR #601, opened 2023-08-25
Implemented configurable search delay

martikyan:
Implemented configurable search delay

#599


Maccy/MenuHeader/MenuHeaderView.swift
```diff
@@ -118,10 +118,15 @@ class MenuHeaderView: NSView, NSSearchFieldDelegate {
   }
 
   private func fireNotification() {
+    searchThrottler.minimumDelay = TimeInterval(Double(UserDefaults.standard.searchDelay) / 1000)
```
p0deje, line 121: Are you doing this to ensure new values are picked up when the user changes delay? If yes, I'd rather have an observer updating delay whenever the UserDefaults value changes. There are examples of this in `Maccy.initializeObservers()`.

p0deje (commented):
Thank you for the pull request. I haven't pulled this to test yet, but I can already see that localizations need to be added before this can be released. If you would like to add those, you will need to:

1. Install bartycrouch from Homebrew.
2. Update localization files by running 
```
bartycrouch update
```
3. It would generate changes in other files as well - please revert those.
4. Go though all new strings in Appearance preferences localization files and update them manually (I use Google Translate for that).

martikyan:
@p0deje thanks for the quick response
I have made the changes you requested. Translated strings using ChatGPT, hope this will work as expected.

That was a huge manual work

p0deje:
@martikyan I am overwhelmed currently at work, but I will definitely get to review/merge your PR, sorry for the delay.

p0deje:
Sorry for taking so long to review, I'm currently trying to test this, but I keep getting weird results when used with large values (e.g. 5000 ms). I type "hello" and some of the keys are just dropped so the search would contain just "el". Do you experience the same problem?

martikyan:
> I type "hello" and some of the keys are just dropped so the search would contain just "el". Do you experience the same problem?

Hmm, I couldn't reproduce this. What's your clipboard size? Do you have specific steps to be able to make it happen?

p0deje:
@martikyan It might be somehow related to Sonoma - can you try on 0.27.0?

martikyan:
> can you try on 0.27.0?

Sorry, I use my Mac mainly for non-Mac development so I don't install preview software :(

p0deje:
> Sorry, I use my Mac mainly for non-Mac development so I don't install preview software :(
I meant Maccy 0.27.0 - just released it.

martikyan:
Tried on 0.27.0, cannot reproduce.
Have you noticed that sometimes when you open Maccy, click on the search bar before typing, everything you type is selected. Is it the expected behavior? Can that be somehow related to this issue?

martikyan:
(Unrelated to the issue) @p0deje Before merging this, let's consider having a button cell to enable/disable instant search instead of text input and stepper. What do you think about that?

p0deje:
> Tried on 0.27.0, cannot reproduce.
> Have you noticed that sometimes when you open Maccy, click on the search bar before typing, everything you type is selected. Is it the expected behavior? Can that be somehow related to this issue?

Just noticed this - I think it's a bug but I don't think it's related.

> Sorry, I use my Mac mainly for non-Mac development so I don't install preview software :(

Sonoma should be out tomorrow - are you planning to upgrade?

martikyan:
> are you planning to upgrade

Sure will do

martikyan:
@p0deje I was able to reproduce that after upgrading to Sonoma and it got fixed as soon as I back-merged from master.

But now I experience other kind of issue, when setting the update interval to 0 and typing fast opening closing Maccy randomly, the results will disappear at some moment with no header

image

After this, keyboard shortcut to open Maccy's popup won't work anymore. I don't know what broke that. Any ideas what can cause that?

martikyan:
> ... typing fast opening closing Maccy randomly, the results will disappear at some moment with no header

@p0deje Reproduced this on 0.27.0 version as well. Will open a separate issue when there'll be more info

p0deje:
> But now I experience other kind of issue, when setting the update interval to 0 and typing fast opening closing Maccy randomly, the results will disappear at some moment with no header

Yeah, that's me trying to workaround Sonoma bugs https://github.com/p0deje/Maccy/commit/e71736adc911e7cd88ce287ee10544270bfa4388. I hope I'll find a better solution that prevents header from disappearing.

p0deje:
@martikyan Would you mind rebasing the PR onto the latest master? There are quite a lot of conflicts due to Preferences -> Settings package move.

martikyan:
> @martikyan Would you mind rebasing the PR onto the latest master? There are quite a lot of conflicts due to Preferences -> Settings package move.

@p0deje Let's have this PR abandoned, and just decrease the refresh time from 400ms to something like 100ms. I don't think configuring the refresh time will make this tool more handy rather than more complex. If set to <100ms, when typing in the search bar, the previous search requests won't be cancelled and the app will start lagging. 400ms is too slow for the updates.
What do you think?

p0deje:
> @p0deje Let's have this PR abandoned, and just decrease the refresh time from 400ms to something like 100ms. I don't think configuring the refresh time will make this tool more handy rather than more complex. If set to <100ms, when typing in the search bar, the previous search requests won't be cancelled and the app will start lagging. 400ms is too slow for the updates.
What do you think?

@martikyan With the recent update to Sonoma the search got noticeably slower and I am not sure 100ms would be a good choice. Can you try playing with the delay in Sonoma with 999 history items and see what is the smallest delay that doesn't create a noticeable log?

PR #618, opened 2023-10-02
Various fixes for Sonoma

weisJ:
Use accessibility api to find the bounds of a menuItem instead of relying on it being connected to the view hierarchy.

My first solution involved reaching into the view hierarchy and extracting the table view used for the menu (just leaving the code here for documentation):
````swift
    // NOTE: This is an ugly hack and makes heavy assumptions on the implementation of menu items.
    //       The view hierarchy is assumed to look like this:
    //       windowContentView -> NSMenuScrollView -> NSClipView -> NSTableViewBackedMenuRepresentation
    guard let tableView = windowContentView.subviews.first?.subviews.first?.subviews.first as? NSTableView else {
      return nil
    }

    let tableIndex = menu.accessibilityIndex(ofChild: item)
    guard tableIndex >= 0 && tableIndex < tableView.numberOfRows,
          let menuItemView = tableView.view(atColumn: 0, row: tableIndex, makeIfNecessary: false) else {
      return nil
    }

    return menuItemView.convert(menuItemView.bounds, to: windowContentView)
````

Though while trying to compute the index of the menu item I ended up using `menu.accessibilityIndex(ofChild: item)`
and looking further into the available methods I found that I could get the location of a menu item in screen coordinates.
From this the bounds relative to the window can be easily computed.
This solution should also work pre Sonoma (though I don't have the means to test it). It also feels like a less brittle solution
than my previous code.

Moreover this solution doesn't need the dummy menu items with an empty view in them any more. So if removing the old codepath altogether is ok with you I would also go ahead and remove those empty menu items as well.


weisJ:
New changes:
- Removed `PreviewMenuItem`
- Fixed some keyboard shortcuts causing the system to beep
- Update the filter in event tracking mode to avoid movement and flickering
- Avoid rebuilding the menu if the visiblity of the header/footer is changed. Makes the settings snappier
  This also avoids adding all items and removing them immediately afterwards if an item limit is set.
- Fixes the misplaced popup on Sonoma by adjusting the position immediately after the widow is made visible. I am intercepting `viewDidMoveToWindow()` on the header view and haven't encountered any flickering (i.e. the window being misplaced for a short amount of time)

p0deje:
I've also noticed that with [Update filter in event tracking mode](https://github.com/p0deje/Maccy/pull/618/commits/c30c635b1914dc5528825ecf970ccf5917420b60) we no longer need these hacky conditions and we can issue previous API (addItem/insertItem/removeItem).

https://github.com/p0deje/Maccy/commit/e71736adc911e7cd88ce287ee10544270bfa4388#diff-7a496c279774679ba987e2bec731bbb5ef31b207d4828dad6fd4ba8dc44e1b60L586-R625

p0deje:
Looking at CI, it did introduce a lot of failing tests, so I wonder if the changes are not working on older macOS. I can test, but do can you confirm UI tests are passing for you on Sonoma?

weisJ:
I’ll have a look at the tests.

weisJ:
I am having some troubles with the tests. For example `testCopyWithClick` fails even though performing the steps manually results in the expected behaviour. Placing some breakpoints it seems that `app.menuItems[copy2].firstMatch.click()` doesn't perform the click (even though it appears to do so). At least it doesn't call the `onSelect` call back of the menu item. Have you experienced this yourself?

p0deje:
Yeah, I remember it working really odd on Sonoma, which is why CI uses Ventura.

weisJ:
In this case I'll try to replicate the failing tests manually. Another issue I stumbled upon: `app.typeText(text)` appears to handle some characters incorrectly. Specifically characters which are bound to an action ('p', 'j', 'k') are typed with a `.command` modifier mask causing tests to fail as those events are swallowed (as they should be). For now I have switched to individually dispatching the single characters.

weisJ:
I think that most failures are related to event ordering. Due to the delayed nature of updating the filter the states aren't up to date when the next step in the test is executed. I am working on fixing it.

weisJ:
Closing in favour of #623. I'll PR the remaining changes once I have ironed out the issues.

Issue #612, opened 2023-09-27
Maccy not appear in screen center sometimes on macOS 14.0 (23A344)


rea1shane:
Sometimes it appear in top center. When I press `cmd` + `shift` + `C` again, a beep will appear.

Maccy version 0.27.0

samtsevich:
That depends on the active window when you try to run Maccy.
For example, in Safari, Cmd+Shift+C is a shortcut for "Start element selection" in Developer Mode.
Thus, I just changed the shortcut that runs Maccy to Cmd-Option-C.

p0deje:
2. Can you please record a screen video showing the problem? You can do this using QuickTime.app.
3. Can you please share the output of running `defaults read org.p0deje.Maccy` in Terminal.app?

weisJ:
I have had this happen as well. I think it occurs if a maximum number of history is specified. Somehow too many items (probably all) are added to the menu causing it to have maximum size (hence the popup being pushed up to the top). Because the item count is limited immediately before the popup is shown one doesn’t see that the menu was actually too large at time of opening.

rea1shane:
![录屏2023-09-28 10 37 15](https://github.com/p0deje/Maccy/assets/93363751/a3f43da5-f53c-4845-927b-33cda67c81a4)

When I press `CMD` + `Shift` + `C` to close the window quickly, there'll be a bell.

Rarely it appear correctly in the center.

```yaml
{
    "KeyboardShortcuts_popup" = "{\\"carbonModifiers\\":768,\\"carbonKeyCode\\":8}";
    "LaunchAtLogin__hasMigrated" = 1;
    "NSStatusItem Preferred Position Item-0" = 5675;
    "NSWindow Frame SUStatusFrame" = "700 743 400 134 0 0 1800 1125 ";
    "NSWindow Frame SUUpdateAlert" = "590 545 620 398 0 0 1800 1125 ";
    "NSWindow Frame com.sindresorhus.Preferences.FrameAutosaveName" = "679 150 383 646 0 0 1920 1055 ";
    SUAutomaticallyUpdate = 1;
    SUEnableAutomaticChecks = 1;
    SUHasLaunchedBefore = 1;
    SULastCheckTime = "2023-09-27 06:59:21 +0000";
    SUUpdateGroupIdentifier = 3766769327;
    WebKitDefaultFontSize = 13;
    WebKitJavaEnabled = 0;
    WebKitJavaScriptEnabled = 0;
[...]

PR #623, opened 2023-10-04
Fix popover location on Sonoma

### LINKED ISSUE #615: Popover location is broken in Sonoma
In Sonoma the popover is always shown in the centre of the menu. This is due to the fact that apparently a menu items
view is no longer part of the menus window view hierarchy?! So far I haven't found a way to determine the location of
a menu item relative to anything. It just reports its own size located at `(0,0)` every time.
I tried to go back to the old implementation for the popover but then the application crashes as popovers have can only be shown
relative to views which are part of the view hierarchy (this was the confirmation that indeed the view is no longer attached to the window).

At this moment I don't think this can be solved without a macOS fix (not even sure they consider this a bug, I definitely do).

weisJ:
I have am still working on resolving the timing issues arising from updating the filter in event tracking mode and the faster header/footer settings changes. In the meantime here are the changes independent of those.

Fixes #615 #612


weisJ:
Failing tests are:
 -  `testNewCopyIsAdded `: This test seems incorrect to me. It tests that the newly copied item doesn't exist, despite it clearly being created once the item is added. I have changed the test to check whether the item is not visible (current behaviour). Alternatively changes to the history could be buffered until the menu is closed. What would be the intended behaviour be here? 
- `testSearchAndCopyWithCommandShortcut `: This test is flaky. The `waitForSearch()` function doesn't seem to wait long enough in this circumstance. Relying on timings with sleep is a bit hacky and should probably be replaced with a wait on a proper condition to check that search has settled down. Not sure how to approach it though.

I have adjusted both tests so they should pass now (hopefully).

Maccy/Clipboard.swift
```diff
@@ -6,9 +6,12 @@ class Clipboard {
 
   typealias OnNewCopyHook = (HistoryItem) -> Void
 
-  var onNewCopyHooks: [OnNewCopyHook] = []
+  private var onNewCopyHooks: [OnNewCopyHook] = []
+  private var syncOnNewCopyHooks: [OnNewCopyHook] = []
```
p0deje, line 10: Can you explain the need for sync hook and event queue?

Maccy/Menu.swift
```diff
@@ -80,124 +78,112 @@ class Menu: NSMenu, NSMenuDelegate {
...
     setKeyEquivalents(historyMenuItems)
+  }
+
+  func menuWillOpen(_ menu: NSMenu) {
+    isVisible = true
+    previewThrottle.minimumDelay = initialPreviewDelay
     highlight(firstVisibleUnpinnedHistoryMenuItem ?? historyMenuItems.first)
   }
 
+  internal func adjustMenuWindowPosition() {
```
p0deje, line 98: I am unsure if that's the reason for the change, but, the popup would jump to the top of the screen/window during the search.

p0deje: It also looks like the popup is not shown at all whenever the Maccy popup is too big (e.g. 999 items in the history).

p0deje (commented):
First of all, @weisJ thank you for this massive PR. It does address plenty of issues on Sonoma, I really appreciate you working on that. `Menu.swift` code is definitely not the best code I've ever written 😆 

It looks like there are plenty of changes however and some of them introduce new issues. For example, now when you type something that is not present in search, hit Enter - it creates a new copy entry in Maccy. Yet, it creates the same entry every time you pop up Maccy. I believe it has something to do with the new sync/async event queue.

Likewise, the popup positioning code prevents showing the popup when it's huge and still doesn't address the popup jumping to to screen/window top during the search.

I have yet to test this in older versions of macOS (Mojave, Ventura) but I think it should be fine there. 

Now to the point - is there any way to decompose this PR into smaller pieces that can be tested/merged one by one? This PR for example could just fix the preview location (which is definitely working great now, thank you). The rest could be:
- PR for fixing the popup location
- PR for fixing event processing
- etc

If that's too much work for you (you've already done so much) - I will try to extract the pieces into separate commits and cherry-pick those into the main branch. I would appreciate it if you could at least reply to my comments so I could understand some of your code.

In any event, thank you so much for all the code and fixes, I learned a lot from it already.

Maccy/Clipboard.swift
weisJ: The test `testCreatesNewCopyOnEnterWhenSearchResultsAreEmpty` expects the added entry to only exist after the popup is reopened (before it wouldn't be visible but the test doesn't check for it). Therefore these events are delayed until the menu is closed. This can be reverted but then the test needs to be adjusted to check the proper behaviour namely either:
- The menu item is added but not visible
- The menu item is added and immediately visible (currently not the case, but could be accomplished)

Maccy/Menu.swift
weisJ: As for changing location during search: This method is only called once after the popup has been shown the first time. It jumps to the top of the screen because that is the location AppKit wants to place the menu. As mentioned before the menu reports a rather large height before being visible (I suppose that menu item alternatives are incorrectly also accounted for in the height). Therefore in theory the menu would have to take up the whole screen placing it at the top. When searching the window is repositioned relative to this initial location on screen.

I will investigate the issue for a large amount of items. Probably skipping adjusting the menu location if the menu is too large should do the trick.

weisJ:
> It looks like there are plenty of changes however and some of them introduce new issues. For example, now when you type something that is not present in search, hit Enter - it creates a new copy entry in Maccy. Yet, it creates the same entry every time you pop up Maccy. I believe it has something to do with the new sync/async event queue.

Oh yes, I know what's causing this. Will adjust the commit.

> Likewise, the popup positioning code prevents showing the popup when it's huge and still doesn't address the popup jumping to to screen/window top during the search.

See my comment above.

> Now to the point - is there any way to decompose this PR into smaller pieces that can be tested/merged one by one? This PR for example could just fix the preview location (which is definitely working great now, thank you). The rest could be:
> 
> * PR for fixing the popup location
> * PR for fixing event processing
> * etc
> 
> If that's too much work for you (you've already done so much) - I will try to extract the pieces into separate commits and cherry-pick those into the main branch. I would appreciate it if you could at least reply to my comments so I could understand some of your code.
> 

I can absolutely do that :)

weisJ:
Closed in favour of #629, #630 and #631

PR #629, opened 2023-10-09
Fix popover location on Sonoma

### LINKED ISSUE #615: Popover location is broken in Sonoma

weisJ:
Fixes #615


weisJ:
Failing test is `testNewCopyIsAdded` as is expected as the test currently doesn't match the implementation (It expects the added entry to not exists, while the entry is added immediately currently).

PR #630, opened 2023-10-09
Some fixes to event handling in order to avoid system beeps

weisJ:
This avoids a sound from being played when the menu is closed or the settings are opened.


weisJ:
Failing tests are `testNewCopyIsAdded` (see #629) and `testCreatesNewCopyOnEnterWhenSearchResultsAreEmpty`. The latter is flaky for me and I can't imagine these changes to cause it to fail.

PR #631, opened 2023-10-09
Fix menu location on Sonoma

### LINKED ISSUE #619: Window jumps while searching
After chasing the content of the search field and the filtering has taken place the window moves slightly down.
Seems to do bigger jumps if the window is opened in the screen centre instead of at the cursor location. Don't know whether
this also happens on versions before Sonoma (haven't really used the search functionality that much).

weisJ:
This depends on #623 

Currently WIP:
- With a large amount of items the menu doesn't show up
- The menu jumps during searches

Fixes #619 #612

### DIFF
--- a/Maccy/Menu.swift
+++ b/Maccy/Menu.swift
@@ -247,6 +247,10 @@ class Menu: NSMenu, NSMenuDelegate {
   }
 
   func updateFilter(filter: String) {
+    let window = menuWindow()
+    var savedTopLeft = window?.frame.origin ?? NSPoint()
+    savedTopLeft.y += window?.frame.height ?? 0.0
+
     var results = search.search(string: filter, within: indexedItems)
 
     // Strip the results that are longer than visible items.
@@ -284,6 +288,11 @@ class Menu: NSMenu, NSMenuDelegate {
 
     setKeyEquivalents(historyMenuItems)
     highlight(filter.isEmpty ? firstUnpinnedHistoryMenuItem : historyMenuItems.first)
+
+    ensureInEventTrackingModeIfVisible(dispatchLater: true) {
+      let window = self.menuWindow()
+      window?.setFrameTopLeftPoint(savedTopLeft)
+    }
   }
 
   func select(_ searchQuery: String) {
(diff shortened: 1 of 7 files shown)


weisJ:
Menu jumps and large menus should be fixed now, however I can imagine that the handling of large menus isn't quite 100% percent correct. I will need to do some more testing to ensure it behaves as expected.

p0deje (commented):
I suppose this is now ready to be reviewed and it looks great. My only comment would be if we could revert the part related to fixing `testCreatesNewCopyOnEnterWhenSearchResultsAreEmpty` that introduces synced hooks and delayed processing. As far as I understand, this is not related to menu jumping and can be fixed in a separate PR or the test could be updated to match a new behavior.

weisJ:
I have reverted the changes and updated the test to reflect the actual behaviour: The menu item may exist but isn't visible

Maccy/Menu.swift
```diff
@@ -78,15 +78,41 @@ class Menu: NSMenu, NSMenuDelegate {
...
     highlight(firstVisibleUnpinnedHistoryMenuItem ?? historyMenuItems.first)
   }
 
+  internal func adjustMenuWindowPosition() {
+    let frame : NSRect?
+    switch UserDefaults.standard.popupPosition {
+    case "center":
+      frame = NSScreen.forPopup?.visibleFrame
+    case "window":
+      frame = NSWorkspace.shared.frontmostApplication?.windowFrame
```
p0deje, line 104: I suppose that's not being properly reported when called from `viewDidMoveToWindow()` so when set to `window` the popup is not positioned properly on the screen. Maybe instead of duplicating the logic across `Menu. adjustMenuWindowPosition` and `Maccy.popUp` we can find a way to pass the frame into menu upon popup and store it as long as the menu is not closed?

p0deje (commented):
Can't wait to merge it as it fixes so many problems, but there is an issue with popping up at the window center. Any chance you could fix it or do you want me to give it a try?

p0deje:
I'll go ahead and merge this - I have prepared a fix for the window center popup. Once again, thank you so much for this - it's amazing how better it behaves on Sonoma now. I'll try testing on older versions later.

Issue #540, opened 2023-03-31
Feature Request: Ignore based on text content?


mmiller42:
Since you have the ability to ignore entries by source application and pasteboard type, could you also add the ability to filter based on text content matching a pattern, such as a regular expression? Obviously these rules would only apply to textual content.

For example, a user might frequently copy a password to their clipboard from various sources, but they frequently screen share while working and don't want to accidentally display their password when opening up Maccy. They might want to pattern match on the first few characters of their password "PASSWORD" e.g. `/^\s*PA.+$/`.  Or they might find themselves copying AWS access keys from the AWS web console to their clipboard frequently and want to ignore them without excluding all pasteboard events from their browser.

p0deje:
Can you elaborate about what kind of regexps are you planning to use to exclude content? The example with PASSWORD/AWS won't likely work because those are unique so I'm not sure how useful this is going to be.

mmiller42:
Well what I mean is that it would be a user-configurable area like the way I can configure ignore source applications and pasteboard types:

<img width="400" src="https://user-images.githubusercontent.com/11927375/229580728-07a2146d-33cb-446e-bfb4-0402fb074ebc.png" />

Then a user could add regular expressions (or globs or whatever you prefer) to another section on this panel, "Text Patterns" or something.

So for example, as an individual user, I might want to add these to my ignore list:

|Text Pattern|Description|
|:---|:---|
|`A(K\|S)IA[A-Z0-9]{16}`|match `AKIA`/`ASIA` followed by 16 uppercase letters and/or numbers, i.e. the format of an AWS access key ID|
|`XX[A-Za-z0-9/+=]{38}`|AWS secret access keys are 40 characters long and contain alphanumerics, `/`, `+`, `=`. Unfortunately sha hashes such as those used in git also match this pattern. But I would not feel horrible about storing the first two characters of the key in order to add it to an ignore list. (So in this example my access key starts with `XX`)|
|`XX[^\s]{30}`|Imagining I commonly copy a long password to my clipboard -- again, putting the first two characters in my ignore list, then simply excluding any string of a certain length and not containing whitespace|
|`https:\/\/privnote\.com`|Some web services offer the ability to create private shareable content by URL -- ideally we could match and exclude URLs so I can copy/paste a URL from a particular website without it adding it to history. Say for example, I use a service to share a private password with coworker 1 via Slack. Before they open it, I accidentally expose the URL via screen share to coworker 2, who accesses it first.|

[...]

PR #632, opened 2023-10-12
Regex ignore

GeorgeKandamkolathy:
Implemented regex ignore pattern for pasteboard items.

#540


weisJ:
This does not compile, did you forget to commit a file? Also the project uses 2 space indentation instead of 4 spaces.

GeorgeKandamkolathy:
Yep, sorry there was a file missing. I've added it back and fixed the indentation as well.

p0deje:
Thank you!

PR #635, opened 2023-10-13
Use Sparkle Swift package

weisJ:
Sparkle provides a Swift package. Is there a reason to check in the complete framework to the repository instead of using the package? If yes feel free to close the PR as this is just something I noticed :).

Note that I am not entirely sure this branch works correctly (as I don't know how to test it). However I follow all the steps from the Sparkle 2.* documentation.


p0deje:
Thanks for submitting, I was reluctant to migrate because I would need to find a way to test this properly to make sure the autoupdate keeps working. I'll give it a shot later.

p0deje:
Thank you, merged in https://github.com/p0deje/Maccy/commit/a0ee02e00db1fb4939d8ed88a8a8a2fada8b7269

Issue #384, opened 2022-03-25
Maccy After being used in the background for many times, the memory occupation increases linearly


iHzx:
## Operating environment
* Device: MacBook Pro 14`2021
* macOS: Monterey 12.3
* Maccy Version: 0.22.2

## Describe the problem
After running Maccy for a long time and using it for many times, the memory occupation of the application will gradually increase (it seems to grow linearly and there is no sign of release). It can be recovered after restarting the application.
I hope the author can fix this problem, thank you.

cemetary:
Device: MacBook Air (M1)
macOS: Monterey 12.3.1.
Maccy Version: 0.22.2

In my case Maccy takes up 1.38 GB.

<img width="1072" alt="Captura de Pantalla 2022-04-10 a la(s) 18 33 03" src="https://user-images.githubusercontent.com/10538213/162647020-554b1203-5999-44bc-ab12-b49df1275f29.png">

p0deje:
The way Maccy uses Core Data could probably be improved to avoid memory bloat - right now every new copy is kept in-memory so if you copy large images (for example) - it can increase over time. I need to see if there is a way to make Core Data work in a different way by offloading unused items from memory.

pbond630:
Alex thx for this.

From the behaviour I've seen on my machine, it appears that references even to deleted items are being retained. The memory usage remains the same even after deleting all items by doing a 'Clear'. 
Shutting down & restarting Maccy is the only way to reclaim the memory.

Divaltor:
What's the status of this issue?

p0deje:
No progress at the moment.

Laicure:
Is it the same how Microsoft Windows store (_Windows+V_) clipboard items, via memory?

I'm thinking if there should be an option to store locally or via memory.

chunyang-wen:
Maybe users have set a large history size? And a lot of images copied.(Just random guess)

nicolaerario:
> Maybe users have set a large history size? And a lot of images copied.(Just random guess)

It's not the case: I can reproduce with small text entries also. Just wait some time and the memory increase progressively

avichou:
<img width="808" alt="image" src="https://github.com/p0deje/Maccy/assets/86660671/a644b9c5-4096-487f-a292-bac86ece3ac1">
it often gets dangerously close to 2gb...

tony-swg:
[...]

PR #647, opened 2023-10-18
Refactor Menu.swift

weisJ:
This works towards refactoring Menu.swift to be easier to understand and less reliant on implicit assumptions and behaviours. The current changes should be rather uncontroversial. Further goals would be:

- Move to a single source of truth about pinned/unpinned. Currently this is checked through both `HistoryMenuItem` and `HistoryItem`.
- More specifically extract a data model which is reflected by the menu on the ui side. Essentially the goal is to make the state of the menu independent of the ui side and be the single source of truth of the state. Of course this model needs to make sure that this is always reflected by the UI.
- Move heavy computations to a background thread which calls back into the main ui thread to update the model/ui.

What do you think about this proposal?


weisJ:
Changes:
- Rebased on #652
- Extracted removal/insertion of all menu items associated to an `IndexedItem` into separate functions.

weisJ:
Not sure why the tests are failing
````
✗ testAllowsToFocusSearchField, failed - Maccy did not pop up
✗ testAllowsToFocusSearchField, Failed to Error Domain=com.apple.dt.xctest.ui-testing.error Code=10001 "Application org.p0deje.Maccy is not running" UserInfo={NSLocalizedDescription=Application org.p0deje.Maccy is not running}: Application org.p0deje.Maccy is not running
````
None of the changes should affect starting up Maccy (Tests pass fine locally).

**Comment by weisJ** [comment]
Converted back to draft to bisect the CI failure

**Comment by p0deje** [comment]
I suppose it needs another rebase now that #652 is merged.

**Comment by p0deje** [comment]
> If I understand it correctly there is a mechanism for the searchable content (value) of a IndexedItem to be different from its title, however this isn't utilised currently i.e. value == title always holds. Are there plans to make use of it?

`value` is a full string while `title` is shortened to fit the menu width. Allowing to searching within `title` first speeds search up because it often contains the search pattern.

**Comment by weisJ** [comment]
> > If I understand it correctly there is a mechanism for the searchable content (value) of a IndexedItem to be different from its title, however this isn't utilised currently i.e. value == title always holds. Are there plans to make use of it?
> 
> `value` is a full string while `title` is shortened to fit the menu width. Allowing to searching within `title` first speeds search up because it often contains the search pattern.

Ah got it, thank you for the clarification. Totally missed that :D. I have adjusted the commit to reflect the earlier behaviour. I couldn't make the `searchableContent` member `@NSManaged`. Not even sure whether  it should be. How should this be handled. I think it makes more sense to have the content in the `HistoryItem` instead of `IndexedItem`.

File: Maccy/Search.swift [hunk 1]
```diff
@@ -40,8 +40,8 @@ class Search {
   private func fuzzySearch(string: String, within: [Searchable]) -> [SearchResult] {
     let pattern = fuse.createPattern(from: string)
     let searchResults: [SearchResult] = within.compactMap({ item in
-      fuzzySearch(for: pattern, in: item.title, of: item) ??
-        fuzzySearch(for: pattern, in: item.value, of: item)
+      fuzzySearch(for: pattern, in: item.title, of: item)
+      ?? fuzzySearch(for: pattern, in: item.item.searchableContent, of: item)
```
p0deje, line 44: To be honest, I don't see a lot of value in replacing `value` to `item.searchableContent`. The current implementation doesn't let the search access internals of `item` so changes in `HistoryItem` are less likely to break search.

Maccy/Menu/Menu.swift
```diff
@@ -119,123 +104,35 @@ class Menu: NSMenu, NSMenuDelegate {
...
-        if popoverWindow.frame.minX < previewWindow.frame.minX {
-          popoverWindow.setFrameOrigin(
-            NSPoint(x: popoverWindow.frame.minX - Menu.popoverGap, y: popoverWindow.frame.minY)
-          )
-        } else {
-          popoverWindow.setFrameOrigin(
-            NSPoint(x: popoverWindow.frame.minX + Menu.popoverGap, y: popoverWindow.frame.minY)
-          )
-        }
+    if let headerView = menuHeader {
```
p0deje, line 109: This is no longer needed since I've moved it back to `MenuHeaderView`.

p0deje (commented):
There are quite a lot of changes and I am not sure I followed all the proposals:

1. I like how you extracted preview logic into its own controller.
2. I don't quite follow the refactoring of `IndexedItem` - what does it bring compared to existing code? It's been added as an internal struct in Menu to support quick search/manipulation of items and their menu items.
3. I don't follow the changes in HistoryItem/HistoryMenuItem. Frankly, the storage implementation probably needs to be re-written to support #384, so I'd rather not touch it at this point.

Maccy/Menu/Menu.swift
weisJ: Fixed

Maccy/Search.swift
weisJ: I have moved the member to `IndexedItem`, so only content of `IndexedItem` is accessed during search.

weisJ:
> 2. I don't quite follow the refactoring of `IndexedItem` - what does it bring compared to existing code? It's been added as an internal struct in Menu to support quick search/manipulation of items and their menu items.

`Menu.swift` is rather large and overwhelming to navigate. I just felt that it makes the code easier to follow. Changed it to now live inside an `extensions Menu` so it is clear that is internal to `Menu`.

> 3. I don't follow the changes in HistoryItem/HistoryMenuItem. Frankly, the storage implementation probably needs to be rewritten to support [Maccy After being used in the background for many times, the memory occupation increases linearly #384](https://github.com/p0deje/Maccy/issues/384), so I'd rather not touch it at this point.

With the latest changes only the `value` member from `HistoryMenuItem` is removed. I was only used to provide the value for `IndexedItem`, which now reuses code from `HistoryItem` to compute the `value` member. Effectively this just - deduplicates the code for turning the content of `HistoryItem` into a string (`HistoryMenuItem` duplicated the pre-truncation code paths from `HistoryItem.generateTitle`)
- and removes the `value` member from `HistoryMenuItem` as it is then no longer needed.

weisJ:
> 2. Preview is show only after the first menu popup - afterwards the preview popover is not shown at all.

This should be fixed now.

 > 1. With 999 history size, Menu shows noticeably slower on the first and subsequent popups. I haven't profiled yet, might as well just try to bisect the commit where it's introduced.

Currently looking into this. Should have a fix ready by tomorrow.

weisJ:
Will try to finish this on the weekend :)

weisJ:
I will close this in favour of upcoming smaller PRs containing smaller batches of these changes.

PR #652, opened 2023-10-20
Deduplicate menu position calculation

weisJ:


p0deje:
Would you mind rebasing the PR onto the latest master?

weisJ:
Rebased and extracted the handling of showing the popup into its own class. Declutters `Maccy.swift` a bit. Also found a member variable (`window`) that wasn't used.

PR #676, opened 2023-11-16
add example password regex to settings description

qpwo:
Sorry I don't have xcode so I'm not certain I did this right. Feel free ofc to make any changes as you see fit, or delete and recreate or whatever, no need to check with me.

Regex explanation:

```
^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)[^\s]{8,}$
(?=.*[a-z])   --  at least one lowercase
(?=.*[A-Z])   --  at least one uppercase
(?=.*\d)      --  at least one digit
[^\s]{8,}     -- 8 or more non-whitespace characters
```


p0deje:
The problem is you now need to translate it in every 16 languages supported. :) Not sure it's worth it.

qpwo:
Damn haha

PR #705, opened 2024-01-12
fix: concealed and transient are incorrectly configured

le0zh0u:
Switched these two places for `concealed` and `transient` in `NSPasteboard.PasteboardType `


p0deje:
Good catch, thank you!

PR #710, opened 2024-01-22
Refactor Menu v2

weisJ:
I will PR in batches the changes from #647. Some of the changes will need some adjustment. Lets start with extracting the preview logic into its own controller class.


Maccy/Menu/Menu.swift
```diff
@@ -635,12 +594,6 @@ class Menu: NSMenu, NSMenuDelegate {
     removeItem(item)
   }
 
-  private func offloadCurrentPreview() {
-    previewThrottle.cancel()
-    previewPopover?.close()
-    previewPopover = nil
-  }
-
   private func boundsOfMenuItem(_ item: NSMenuItem, _ windowContentView: NSView) -> NSRect? {
```
p0deje, line 597: I think you forgot to remove this one as it's now in controller.

weisJ:
Good catch. Looks like I forgot to stage this change. Will amend it later today.

p0deje (approved):
LGTM, I am not sure why CI is failing but given it's green on master, can you please double-check that you haven't accidentally broken anything with this change?

weisJ:
Should be good now

PR #713, opened 2024-01-24
Update GeneralSettingsViewController.strings

dennis-benzinger-hybris:
Fix the translation for "open"

The old translation was for the adjective "open" but not for the verb.


p0deje:
Thank you!

PR #714, opened 2024-01-25
Fix copying multiple files in Finder

### LINKED ISSUE #695: Copy multiple files shows as a single file

- [X] I have verified that there isn't already an issue reporting the same bug to prevent duplication.
- [X] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

_No response_

### macOS Version

_No response_

### Maccy Settings

_No response_

### Description

_No response_

### Steps to Reproduce

Just copy multiple files at once in Finder - it currently shows as a single file in history.

kthchew:
Fixes #695.

The UI currently looks like this:
Screenshot 2024-01-25 at 12 40 51 AM

In some places, I sometimes used `.first` as a shortcut to support the changes in `HistoryItem` from `Data?` to `[Data]`. I'm not sure if this is the best solution, but I believe it would match the behavior before this PR for those data types.

As for commit 00d41fc, my changes to fix the bug uncovered an issue where `HistoryItemContent.value` is sometimes `nil`, which caused a crash with images copied on another iCloud device with Universal Clipboard when Maccy is restarted. Without this PR, you can also observe this crash by trying to copy the affected image item (which appears blank after restarting Maccy). I thus had to remove the force unwrap.

### DIFF
--- a/Maccy/History/HistoryItem.swift
+++ b/Maccy/History/HistoryItem.swift
@@ -207,6 +207,6 @@ class HistoryItem: NSManagedObject {
     let contents = getContents()
     return contents
       .filter { types.contains(NSPasteboard.PasteboardType($0.type)) }
-      .map { $0.value }
+      .compactMap { $0.value }
   }
 }
--- a/Maccy/History/HistoryItem.swift
+++ b/Maccy/History/HistoryItem.swift
@@ -53,45 +53,57 @@ class HistoryItem: NSManagedObject {
   @NSManaged public var pin: String?
   @NSManaged public var title: String?
 
-  var fileURL: [URL] {
+  var fileURLs: [URL] {
     guard !universalClipboardText else {
       return []
     }
 
-    return contentData(filePasteboardTypes)
+    return allContentData(filePasteboardTypes)
       .compactMap { URL(dataRepresentation: $0, relativeTo: nil, isAbsolute: true) }
   }
 
-  var htmlData: [Data] { contentData(htmlPasteboardTypes) }
-  var html: [NSAttributedString] {
-    htmlData
-      .compactMap { NSAttributedString(html: $0, documentAttributes: nil) }
-  }
-
-  var image: [NSImage] {
-    var data = contentData(imagePasteboardTypes)
-      .compactMap { NSImage(data: $0) }
-    
-    if data.isEmpty && universalClipboardImage {
-      return fileURL
-        .compactMap { try? Data(contentsOf: $0) }
-        .compactMap { NSImage(data: $0) }
-    } else {
-      return data
+  var htmlData: Data? { contentData(htmlPasteboardTypes) }
+  var html: NSAttributedString? {
+    guard let data = htmlData else {
+      return nil
     }
+
+    return NSAttributedString(html: data, documentAttributes: nil)
   }
 
-  var rtfData: [Data] { contentData(rtfPasteboardTypes) }
-  var rtf: [NSAttributedString] {
-    return rtfData.compactMap { NSAttributedString(rtf: $0, documentAttributes: nil) }
+  var image: NSImage? {
+    var data: Data?
+    data = contentData(imagePasteboardTypes)
+    if data == nil, universalClipboardImage, let url = fileURLs.first {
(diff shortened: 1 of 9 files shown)


p0deje:
This is quite a change in how storage is organized, do you think it might be possible to keep the current contract and only change `fileURL` to `fileURLs` so that it always returns a list of files? That's so far the first case in years where we might want to represent pasteboard items in a single history item.

kthchew:
I believe that would be possible. I think `contentData` might still need to return an array, but `.first` can be used everywhere except file URLs. I should be able to experiment with this by tomorrow.

kthchew:
I pushed some less significant changes that only change how file URLs are handled. Let me know what you think.

p0deje:
> As for commit https://github.com/p0deje/Maccy/commit/00d41fc1da1cf9a26f8c2edd4f9a7623cc106d5e, my changes to fix the bug uncovered an issue where HistoryItemContent.value is sometimes nil, which caused a crash with images copied on another iCloud device with Universal Clipboard when Maccy is restarted. Without this PR, you can also observe this crash by trying to copy the affected image item (which appears blank after restarting Maccy). I thus had to remove the force unwrap.

Can you provide the exact steps to reproduce the issue? I tried copying a random image on my iPhone and ii unwrapped properly in Maccy.

Maccy/History/HistoryItem.swift
```diff
@@ -183,8 +181,10 @@ class HistoryItem: NSManagedObject {
       return title
     }
 
-    if let fileURL = fileURL {
-      title = fileURL.absoluteString.removingPercentEncoding ?? ""
+    if !fileURLs.isEmpty {
+      title = fileURLs
+        .compactMap { $0.absoluteString.removingPercentEncoding }
+        .joined(separator: "; ")
```
p0deje, line 187: Shall we use `\n` here?

kthchew:
> Can you provide the exact steps to reproduce the issue? I tried copying a random image on my iPhone and ii unwrapped properly in Maccy.

On Maccy 0.29.4:

1. Copy some image on the iPhone.
2. Check Maccy to see that it appears in the clipboard history (it works correctly at this point)
3. Copy something else on the phone (like text for example). I did not have to do this originally (maybe a wait period will also trigger the crash), but it seems like this crash doesn't happen 100% of the time unless you do.
4. Quit Maccy
5. Reopen Maccy
6. Check the history again, the entry that used to have the image is blank. Clicking it causes a crash.

Maccy/History/HistoryItem.swift
kthchew: Should it appear as a literal newline in the history, or as a "⏎" like other newlines when copying text?

p0deje:
I am not in front of a computer, but I think the code below would replace newlines with special symbol. If that’s the case, you can just use \n. On Jan 27, 2024, at 16:00, Kenneth Chew ***@***.***> wrote:﻿
@kthchew commented on this pull request.



In Maccy/History/HistoryItem.swift:
> @@ -183,8 +181,10 @@ class HistoryItem: NSManagedObject {
       return title
     }
 
-    if let fileURL = fileURL {
-      title = fileURL.absoluteString.removingPercentEncoding ?? ""
+    if !fileURLs.isEmpty {
+      title = fileURLs
+        .compactMap { $0.absoluteString.removingPercentEncoding }
+        .joined(separator: "; ")

Should it appear as a literal newline in the history, or as a "⏎" like other newlines when copying text?

p0deje:
I still couldn't reproduce the crash, but the changes make storage access safer in any event. Thank you!

p0deje:
Not sure why exactly, but looks like this also introduced a bug where copied formatted text could be duplicated. For example, when I copy something from Safari and then select it in history, it's pasted in 3 copies. @kthchew Would you like to try and implement a fix?

kthchew:
It seems like this is happening due to the change from `setData` to `writeObjects` to write to the pasteboard. But for some reason, the old method of using `setData` meant that only the first file was copied if selecting it from the history. I'll make a new PR that only uses `writeObjects` for files.

PR #755, opened 2024-03-27
Remove header underline in README

phlpsong:
Before:

Screenshot 2024-03-27 at 22 18 39

After:
Screenshot 2024-03-27 at 22 18 30


phlpsong:
@p0deje  apologies, this does not work well in Github README but works well locally.

PR #760, opened 2024-04-04
Translated Maccy to Dutch (nl)

yasingunaydiin:
I have translated Maccy to Dutch and will be maintaining it. Feel free to let me know of any errors!

PR #764, opened 2024-04-18
add arabic Localizable.strings

columbusux:
we need RTL to make it readable


p0deje (commented):
Thank you for preparing the PR! However, this is only a part of localization that needs to be done for Maccy, there are much to translate for preferences. Any chance you could also add those?

An example of complete translation PR is https://github.com/p0deje/Maccy/pull/760.

p0deje:
> Yes, I stopped temporarily after I found the letters being reversed because I asked a friend and he said that the project must support the RTL feature, and I make this translation initially to verify compatibility and that it appears correctly.

Maccy fully supports RTL, you can proceed.

Issue #826, opened 2024-08-02
[V2] Remember position when moving the popup if mode is set to center of screen


weisJ:
If the positioning mode is set to "Screen center" after moving the Maccy-window manually the position should be saved just as for spotlight. Of course a "reset position" button in the settings would then be beneficial.

weisJ:
It needs to be considered how the custom position should be interpreted on different screens. Either have it be in percentage of the width height of the screen or remember the position per screen. WDYT?

p0deje:
I suppose the implementation might look similar to screen center.

1. The user chooses "Last position". 
2. If there are multiple screens, the option has a submenu allow to either select "Active screen" or specify an exact screen to choose. 
3. For "Active screen", we need to somehow calculate the position of the window on the current screen relative to its resolution. For example, if window `origin` is (500,500) and resolution is (1000,1000), the relative origin would be saved in a format that ensures it will be interpreted as (1000,1000) on a screen with resolution of (2000,2000). This probably can be expressed as a multiplier for screen resolution (e.g. for the example we'll save `originMultiplier=(0.5,0.5)`. Later, when Maccy pops up it would place window on the active screen and will use the multiplier to offset accordingly.
4. For specific screen, we can still keep the multiplier but place window on the configured screen.

p0deje:
As I am writing this, I realized that 2.0 misses "Active screen" in "Screen center" - it's a bug that needs to be fixed.

p0deje:
I've fixed screen center logic in https://github.com/p0deje/Maccy/commit/7c844aa250fdf64466e7ce9391cfb1ceca7499d3, you can implement a similar SwiftUI logic for "Last position".

weisJ:
#829

Issue #712, opened 2024-01-23
Add option to position menu like spotlight


weisJ:
### Before Submitting Your Feature Request

- [X] Check that there isn't already a similar feature request to avoid creating a duplicate.
- [X] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Problem

I would like to have an option to position the menu such that its top edge aligns with the top edge default spotlight location.
This would make it more predictable where the menu pops up.

### Solution

This could just replace the center in window mode or be a separate setting.

I couldn't quite figure out how to add a new option in the corresponding settings combobox, so if the second option is preferable I would need some assistance in implementing this.

p0deje:
> I would like to have an option to position the menu such that its top edge aligns with the top edge default spotlight location.
This would make it more predictable where the menu pops up.

Spotlight window is movable, so I am not sure I follow how this makes location more predictable. Can you elaborate a bit more?

weisJ:
I don't intend to actually follow the spotlight position. Currently depending on the number of history items the menu is larger or taller. Because the entering modes center the whole menu window this causes the search bar to be at different positions, which feels a bit unintuitive e.g. if I have only 1 entry in the history the search bar will be close to the center of the screen while with e.g. 100 entries it will be at the top of the screen.

p0deje:
Ok, I see what you mean now. This could be useful, and I wouldn't mind merging a PR. 

However, right now I feel like we might need to get rid of the custom logic that adjusts menu height/position as it seems that macOS 14.3 finally fixed its issues around that and our logic is still error-prone in many cases (#29). I have a build where I removed all that custom logic and so far Maccy seems to position just fine by cursor and no longer jumping around as it used to.

Do you think we'd be able to implement this feature w/o the code for custom adjustment of menu height or enabled only for it?

p0deje:
Actually, 14.3 still reports an incorrect menu height initially so the code would need remain. I will adapt it to only run when centering a menu in a screen/window.

p0deje:
@weisJ Would you mind looking at https://github.com/p0deje/Maccy/commit/84422077275fc2f5e6f17b85c686e426af731e49 and see if you can spot any issues? This seems to work fine in my basic testing scenarios.

PR #827, opened 2024-08-02
Add option to position like spotlight search

weisJ:
See #712 

Maybe this could also be consolidated with the `center` option using #826 by adding some "snap" spots for known locations.

Also should I add all other translations as well?

### DIFF
--- a/Maccy/PopupPosition.swift
+++ b/Maccy/PopupPosition.swift
@@ -7,6 +7,7 @@ enum PopupPosition: String, CaseIterable, Identifiable, CustomStringConvertible,
   case statusItem
   case window
   case center
+  case spotlight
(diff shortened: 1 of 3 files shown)


p0deje:
What do you think if instead of doing this custom positioning, we simply let users position/resize window wherever they want and remember the position? Since the window now hides extra items under a scroll menu, the search bar would be positioned in the same place regardless of the number of items in history.

weisJ:
Yes that is what I proposed in #826 👍🏼

p0deje:
Do you want to replace this PR with a new implementation or create a new one?

weisJ:
Feel free to close it. I'll submit a new one

PR #829, opened 2024-08-02
Add option to remember the last window position

weisJ:
Haven't yet updated all the translations. 

Also made it such that the window is no longer movable when "Last position" isn't selected as it may be confusing if moving the window has no effect (this may not be desirable, if so it can be reverted).

Some additional thought:

The window center method could be converted to "Presets" for `.lastPosition` together with the current default position of `.lastPosition`. I imagine this being an additional dropdown with Options `["Default", "Center", "Custom"]`, where `Custom` is selected if the window is moved. This way the reset button could also be removed as resetting is a matter of changing it back to `Default`.

image

### DIFF
--- a/Maccy/Extensions/Defaults.Keys+Names.swift
+++ b/Maccy/Extensions/Defaults.Keys+Names.swift
@@ -48,4 +48,5 @@ extension Defaults.Keys {
   static let sortBy = Key<Sorter.By>("sortBy", default: .lastCopiedAt)
   static let suppressClearAlert = Key<Bool>("suppressClearAlert", default: false)
   static let windowSize = Key<NSSize>("windowSize", default: NSSize(width: 450, height: 800))
+  static let windowPosition = Key<NSSize>("windowPosition", default: NSSize(width: 0.5, height: 0.8))
 }
--- a/Maccy/Extensions/Defaults.Keys+Names.swift
+++ b/Maccy/Extensions/Defaults.Keys+Names.swift
@@ -48,5 +48,5 @@ extension Defaults.Keys {
   static let sortBy = Key<Sorter.By>("sortBy", default: .lastCopiedAt)
   static let suppressClearAlert = Key<Bool>("suppressClearAlert", default: false)
   static let windowSize = Key<NSSize>("windowSize", default: NSSize(width: 450, height: 800))
-  static let windowPosition = Key<NSSize>("windowPosition", default: NSSize(width: 0.5, height: 0.8))
+  static let windowPosition = Key<NSPoint>("windowPosition", default: NSPoint(x: 0.5, y: 0.8))
 }
--- a/Maccy/FloatingPanel.swift
+++ b/Maccy/FloatingPanel.swift
@@ -99,7 +99,9 @@ class FloatingPanel<Content: View>: NSPanel, NSWindowDelegate {
   }
 
   func windowDidMove(_ notification: Notification) {
+    // Only update position if the window is visible to avoid accumulating floating point errors.
     guard isVisible else { return }
+
     if let screenFrame = screen?.visibleFrame {
       let anchorX = frame.minX + frame.width / 2 - screenFrame.minX
       let anchorY = frame.maxY - screenFrame.minY
--- a/Maccy/FloatingPanel.swift
+++ b/Maccy/FloatingPanel.swift
@@ -55,7 +55,7 @@ class FloatingPanel<Content: View>: NSPanel, NSWindowDelegate {
       rootView: view()
         .ignoresSafeArea()
         .gesture(DragGesture()
-          .onChanged { _ in
+          .onEnded { _ in
             if let screenFrame = self.screen?.visibleFrame {
               let anchorX = self.frame.minX + self.frame.width / 2 - screenFrame.minX
               let anchorY = self.frame.maxY - screenFrame.minY
--- a/Maccy/FloatingPanel.swift
+++ b/Maccy/FloatingPanel.swift
@@ -91,6 +91,22 @@ class FloatingPanel<Content: View>: NSPanel, NSWindowDelegate {
     return frameSize
   }
 
+  override var isMovable: Bool {
+    get {
+      return Defaults[.popupPosition] == .lastPosition
+    }
+    set {}
+  }
+
+  func windowDidMove(_ notification: Notification) {
+    guard isVisible else { return }
+    if let screenFrame = screen?.visibleFrame {
+      let anchorX = frame.minX + frame.width / 2 - screenFrame.minX
+      let anchorY = frame.maxY - screenFrame.minY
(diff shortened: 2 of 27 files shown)


Maccy/Extensions/Defaults.Keys+Names.swift
```diff
@@ -48,4 +48,5 @@ extension Defaults.Keys {
   static let sortBy = Key<Sorter.By>("sortBy", default: .lastCopiedAt)
   static let suppressClearAlert = Key<Bool>("suppressClearAlert", default: false)
   static let windowSize = Key<NSSize>("windowSize", default: NSSize(width: 450, height: 800))
+  static let windowPosition = Key<NSSize>("windowPosition", default: NSSize(width: 0.5, height: 0.8))
```
p0deje, line 51: `NSSize` doesn't seem to be a good choice to store this. Maybe keep a simple dictionary with `x/y` or `NSPoint`?

Maccy/FloatingPanel.swift
```diff
@@ -91,6 +91,22 @@ class FloatingPanel<Content: View>: NSPanel, NSWindowDelegate {
     return frameSize
   }
 
+  override var isMovable: Bool {
```
p0deje, line 94: I think it's okay to let window be moved even without last position. Sometimes it might appear and hide something that user would want to see, so we let the user to move the window and then paste.

Maccy/FloatingPanel.swift
```diff
@@ -91,6 +91,22 @@ class FloatingPanel<Content: View>: NSPanel, NSWindowDelegate {
...
+    get {
+      return Defaults[.popupPosition] == .lastPosition
+    }
+    set {}
+  }
+
+  func windowDidMove(_ notification: Notification) {
+    guard isVisible else { return }
```
p0deje, line 102: Is this guard necessary? What happens when it's removed?

Maccy/Settings/AppearanceSettingsPane.swift
```diff
@@ -49,33 +50,44 @@ struct AppearanceSettingsPane: View {
...
               }
-            } else {
-              Text(position.description)
             }
           }
+          .labelsHidden()
+          .frame(width: 141)
+          .help(Text("PopupAtTooltip", tableName: "AppearanceSettings"))
+
+          if popupAt == .lastPosition {
```
p0deje, line 82: Can you add a screenshot of the new settings pane?

Maccy/Extensions/Defaults.Keys+Names.swift
weisJ: I wanted to use a `NSPoint` but swift complaint that it somehow isn't default serialisable.

Maccy/FloatingPanel.swift
weisJ: Sure thing, I'll revert this part.

Maccy/FloatingPanel.swift
weisJ: Somehow the method is fired while still hidden. Repeatedly opening/closing caused some noticeable movement due to floating point errors, so I made it such that the value is only changed if the user actively moves the window (i.e. while it is visible).

Maccy/Settings/AppearanceSettingsPane.swift
weisJ: Attached an image. The button is only enabled if the position isn't the default value. It also is only shown if `.lastPosition` is active.

Maccy/Extensions/Defaults.Keys+Names.swift
p0deje: You likely need to extend similar to how I did NSSize - https://github.com/p0deje/Maccy/blob/07aac99e746d083bba3a2a02033849c7fcf673c7/Maccy/Extensions/NSSize+DefaultsSerializable.swift

Maccy/Settings/AppearanceSettingsPane.swift
```diff
@@ -49,33 +50,44 @@ struct AppearanceSettingsPane: View {
...
+            Button(action: {
```
p0deje, line 83: You can omit `action` here:

```swift
Button {
  _windowPosition.reset()
} label: {
  Text("PopupAtLastLocationReset", tableName: "AppearanceSettings")
}
```

Maccy/Settings/AppearanceSettingsPane.swift
p0deje: Thanks, what do you think about making UI more compact and using a simple "x" button:

1. Can be clicked any time even if it's already reset.
2. Uses `x.circle` SF symbol.
3. Uses borderless button style.

Maccy/Extensions/Defaults.Keys+Names.swift
weisJ: Done

Maccy/FloatingPanel.swift
weisJ: Changed it back. Though I have kept it disabled for `.statusItem`. Being able to move a menubar menu feels weird.

Maccy/FloatingPanel.swift
weisJ: Added a comment for clarification

Maccy/Settings/AppearanceSettingsPane.swift
weisJ: Done. Though I have opted to use `arrow.uturn.backward` as it is more commonly associated with a "revert" action.

image

p0deje: What do you think about `arrow.uturn.backward.circle` or `arrow.uturn.backward.circle.fill`?

weisJ: Ah I was searching for this but couldn't find them. `arrow.uturn.backward.circle.fill` is what I wanted to do initially.

weisJ:
Changes: More reliable detection of deliberate window movement

Maccy/FloatingPanel.swift
```diff
@@ -52,7 +52,16 @@ class FloatingPanel<Content: View>: NSPanel, NSWindowDelegate {
     /// Set the content view.
     /// The safe area is ignored because the title bar still interferes with the geometry
     contentView = NSHostingView(
-      rootView: view().ignoresSafeArea()
+      rootView: view()
+        .ignoresSafeArea()
+        .gesture(DragGesture()
+          .onChanged { _ in
```
p0deje, line 58: Should you use `.onEnded` instead to skip intermediate dragging steps?

weisJ: Yes, good catch

weisJ:
Added the remaining translations using Deepl and Google Translate.

Suborna88mc:
Im waiting for last 5 hours but no any one i find here..help me pls

Issue #912, opened 2024-10-04
Display the app icon, from which the text was copied 


Romachamp:
### Before Submitting Your Feature Request

- [X] Check that there isn't already a similar feature request to avoid creating a duplicate.
- [X] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Problem

Hello! I’ve used Maccy for a long time, but the main issue I faced was not being able to instantly see, from where I copied the item. Later, I discovered I can see the app, from where I copied, but it isn’t displayed next to it, so I must click on the item to get this information. 

### Solution

The solution I want is the app icon, from where the item is copied, right next to it. In this way, I can instantly see, from where I copied anything. 

Alternatively, you could provide an option in the settings to display app icon next to the copied item, instead of the keyboard shortcut to paste that item. It would be more handy for me personally.

p0deje:
You can see the source app in the preview, which can be shown instantly by changing Preview Delay. Is this enough for your needs?

Romachamp:
Well, I’ve used this feature. However, in the preview the app is shown by the text and it still isn’t that handy. With app icon right next to the copied icon, I would be able to see it instantly and it would be more handy to me.

Romachamp:
@p0deje you could add an option to the settings to show the app icon, instead of shortcuts, so people can decide what’s more handy for them. What do you think about it?

p0deje:
I am open to merge PRs adding support for this, but I don't have capacity to implement it myself at the moment.

weisJ:
I have this actually implemented in the branch where I keep my custom changes: ab9fb0aa666b64bfcf17964d3397d34297520798

I don't have a lot of time to clean it up and connect it to the settings, however it is a good starting point @Romachamp

p0deje:
> I have this actually implemented in the branch where I keep my custom changes: https://github.com/p0deje/Maccy/commit/ab9fb0aa666b64bfcf17964d3397d34297520798

Screenshots or didn't happen.

Romachamp:
Awesome, this is exactly what I wanted. It would be awesome, if it were in the app.

Romachamp:
@p0deje Will you implement this feature now, since there is a pull request?

p0deje:
@Romachamp There is no PR yet.
@weisJ Are you up to create a PR with this? We'd need a checkbox in Appearance setting to turn it on (with off by default) and localizations for it in all languages (I can help with that).

[...]

PR #942, opened 2024-11-04
Add option to show application icon

weisJ:
Add option to show app icon in the history list (see #912). The settings toggle is missing translations to languages other than english.
The app icon is also shown in the popover. 

Icons are kept for 24 hours (so they are updated if the application icon changes).

I tried to match the icon size to the visual height of the text next to it (for both the history list and the popover).

image

image


weisJ:
Changed this to draft as I have a better idea for how to do icon invalidation.

weisJ:
I have changed invalidation to attach a file system event handler to listen for changes to the app file. If it is modified the icon is updated. If it is deleted a fallback icon will be shown and every hours a new attempt at retrieving the icon is made.

weisJ:
Fixed the SwiftLint issues.

Related to this feature: As I have been using it enabled the last two weeks I noticed that due to a lot of copies coming from the web the icons didn't add much context for me in this case. I have now a sort of good working mechanism which shows me the favicon of the webpage the copy originated from instead of the app icon.
Crucially this of course requires making requests to those webpages, which broadens the sandbox requirements for Maccy. Would this feature be something worth while preparing a PR for? (I totally understand if the privacy concerns here outweigh the potential benefits)

Romachamp:
I think just app icons would be enough. Website icons aren’t all that good, and I still use a lot of other apps too. So, everything, which is done currently is pretty good and suits my needs.

p0deje:
Merged in 9f96adc with some smaller changes (refactoring, vertically aligning icon for image copies, extra leading padding).

Thanks again @weisJ, you rock!

weisJ:
Just noticed: If you want the icon to be vertically centred then you could have simply removed the `VStack`. I put it there as I thought having the icon top aligned would look more natural.

PR #964, opened 2024-11-25
Adjust paddings to make history list edge to edge

weisJ:
Adjusts some paddings to make the history list more "edge to edge" removing some whitespace cutting of history items half way.

**Before:**
SCR-20241125-kbep
SCR-20241125-kbkh
SCR-20241125-kbna

**After:**
SCR-20241125-kccp
SCR-20241125-kceb
SCR-20241125-kbxu


p0deje (commented):
I modeled paddings around the context menu visual representation and older version of Maccy. I think removing the horizontal paddings makes it look less consistent compared to regular macOS menus.

weisJ:
Makes sense 👍🏼

Issue #96, opened 2020-04-14
Subsequent press of hotkey selects next item


jeffbowen:
For the longest time I've used [LaunchBar](https://www.obdev.at/products/launchbar/index.html) as my clipboard manager of choice all because of one feature. By pressing the hotkey that invokes the clipboard manager subsequent times, you can select the next item in the list (described under "Show clipboard history" [here](https://www.obdev.at/resources/launchbar/help/ClipboardHistory.html)).

It would be amazing if the same functionality could be available in Maccy as I don't use any other features of LaunchBar, am tired of it's "pay for a license" nagging, and have never found another clipboard manager that does the same thing.

To further clarify, the user would press the hotkey which would display Maccy. While continuing to hold down the modifier key(s) they used, they can press the non-modifier key again to select the next item in the list. Upon releasing all of the keys, whichever item was selected, pastes. Try out LaunchBar (free) to get a better idea of how it works or just ask and I can explain further.

A few months ago I took a crack at trying to add it to Maccy myself but wasn't able to get it working.

Also, for what it's worth, my hotkey of choice is command + option + v. It's always made sense to me logically as it's the paste command, plus the option/alt modifier.

p0deje:
During the last few months I've taken mulitple attempts trying to implement this feature. However, I cannot figure out a way to trigger event when modifier flags are released. Current implementation uses Carbon event handlers to intercept system key down/up events because NSMenu doesn't allow to do that using regular view-based methods. The solution Maccy uses is based on https://github.com/mikekazakov/MGKMenuWithFilter. However, I can't find a combination of EventTypeSpec that would detect releasing of modifier keys.

georgesnow:
I believe you have to use either a global event monitor or event tap.

p0deje:
It's now possible to implement this because https://github.com/p0deje/Maccy/blob/master/Maccy/RunLoopLocalEventMonitor.swift would allow us to get the `.flagsChanged` event. If anyone wants to take a shot at implementing this, PRs are welcome and I can guide you.

7zism3:
@p0deje Can you guide me, I would love to try this! Thank you

p0deje:
[...]

PR #1003, opened 2024-12-25
Add repeated-press logic to cycle popup selection

alexisgaziello:
Hey!

I was hoping to contribute to Maccy.

I'd like to add an option to cycle over the clipboard option on repeated "open" shortcut stroke. 

Before merge, there are probably a couple of things needed to be done, such as adding tests.

I was looking to get your thoughts before spending more time on it. Couple of questions:

- Would you like to get this feature merged?
- Do you use any formatter/linter? (It's my first time working with Swift/Xcode) I tried using [swiftformat](https://github.com/nicklockwood/SwiftFormat) and most of the files changed.
- Does the current logic look OK?
- Any general requirements to create a PR/commit?

### DIFF
--- a/Maccy.xcodeproj/project.pbxproj
+++ b/Maccy.xcodeproj/project.pbxproj
@@ -125,6 +125,9 @@
 		DAFE2DDA268A521B00990986 /* String+Shortened.swift in Sources */ = {isa = PBXBuildFile; fileRef = DAFE2DD9268A521A00990986 /* String+Shortened.swift */; };
 		DAFE2DE9268A9B1B00990986 /* HistoryItemTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DAFE2DE8268A9B1B00990986 /* HistoryItemTests.swift */; };
 		DAFEF0B8249D7DEE006029E8 /* KeyboardShortcuts.Name+Shortcuts.swift in Sources */ = {isa = PBXBuildFile; fileRef = DAFEF0B7249D7DEE006029E8 /* KeyboardShortcuts.Name+Shortcuts.swift */; };
+		FE5B372C2D26274E00A9BC20 /* CycleSelection.swift in Sources */ = {isa = PBXBuildFile; fileRef = FE5B372B2D26274E00A9BC20 /* CycleSelection.swift */; };
+		FE5B37302D26BC6C00A9BC20 /* CycleSelectionUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = FE5B372F2D26BC6C00A9BC20 /* CycleSelectionUITests.swift */; };
+		FED48BDC2D2A1AF800FC57AF /* BaseTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = FED48BDB2D2A1AF800FC57AF /* BaseTest.swift */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
@@ -483,6 +486,9 @@
 		DAFE2DD9268A521A00990986 /* String+Shortened.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "String+Shortened.swift"; sourceTree = "<group>"; };
 		DAFE2DE8268A9B1B00990986 /* HistoryItemTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = HistoryItemTests.swift; sourceTree = "<group>"; };
 		DAFEF0B7249D7DEE006029E8 /* KeyboardShortcuts.Name+Shortcuts.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "KeyboardShortcuts.Name+Shortcuts.swift"; sourceTree = "<group>"; };
+		FE5B372B2D26274E00A9BC20 /* CycleSelection.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CycleSelection.swift; sourceTree = "<group>"; };
+		FE5B372F2D26BC6C00A9BC20 /* CycleSelectionUITests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CycleSelectionUITests.swift; sourceTree = "<group>"; };
+		FED48BDB2D2A1AF800FC57AF /* BaseTest.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BaseTest.swift; sourceTree = "<group>"; };
 /* End PBXFileReference section */
 
 /* Begin PBXFrameworksBuildPhase section */
@@ -568,6 +574,8 @@
 			isa = PBXGroup;
 			children = (
 				DA0EE7B8204657830025FC60 /* MaccyUITests.swift */,
+				FED48BDB2D2A1AF800FC57AF /* BaseTest.swift */,
+				FE5B372F2D26BC6C00A9BC20 /* CycleSelectionUITests.swift */,
 			);
 			path = MaccyUITests;
 			sourceTree = "<group>";
@@ -740,6 +748,7 @@
 				DA20FA712B082DD600056DD5 /* Notifier.swift */,
 				DA689FC72C1D15140009B887 /* PinsPosition.swift */,
 				DA689FC52C1D14F10009B887 /* PopupPosition.swift */,
+				FE5B372B2D26274E00A9BC20 /* CycleSelection.swift */,
 				DAC14123232367B200FCFA30 /* Search.swift */,
 				2F1A79BF2C6DFB7800C98EBD /* SearchVisibility.swift */,
 				DAA5ACC92C1BEE8A00B58513 /* SoftwareUpdater.swift */,
@@ -952,6 +961,8 @@
 			buildActionMask = 2147483647;
 			files = (
 				DA0EE7B9204657830025FC60 /* MaccyUITests.swift in Sources */,
+				FE5B37302D26BC6C00A9BC20 /* CycleSelectionUITests.swift in Sources */,
+				FED48BDC2D2A1AF800FC57AF /* BaseTest.swift in Sources */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
@@ -993,6 +1004,7 @@
 				DA689FC82C1D15140009B887 /* PinsPosition.swift in Sources */,
 				DA13D7D92C1A223E00FA9E23 /* Get.swift in Sources */,
 				DA1969182C3F327500258481 /* SearchFieldView.swift in Sources */,
+				FE5B372C2D26274E00A9BC20 /* CycleSelection.swift in Sources */,
 				DAE8F5D42C43262B00851CA9 /* Popup.swift in Sources */,
 				DA13D7DA2C1A223E00FA9E23 /* Clear.swift in Sources */,
 				DA13D7DB2C1A223E00FA9E23 /* Delete.swift in Sources */,
--- a/Maccy/CycleSelection.swift
+++ b/Maccy/CycleSelection.swift
@@ -1,0 +1,162 @@
+import AppKit
+import KeyboardShortcuts
+
+/// Represents a popup action that cycles through clipboard history items.
+final class CycleSelection {
+
+  /// The event tap reference.
+  private var eventTap: CFMachPort?
+  /// The event tap's run loop source.
+  private var runLoopSource: CFRunLoopSource?
+  /// Pointer to callback context data (holding keyCode and modifiers).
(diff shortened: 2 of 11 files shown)


p0deje:
Thank you for the pull request! 

> Would you like to get this feature merged?
> Does the current logic look OK?

I very much like the idea of cycle hotkey, but I believe it should be implemented as the oldest open issue in Maccy (#96) requests:

1. You press COMMAND+SHIFT+C (or other shortcut) to popup Maccy.
2. You release С but keep COMMAND+SHIFT pressed.
3. You press С as many times as you need. Each press navigates to the next item in history.
4. You release COMMAND+SHIFT which confirms the current selection and copies it.
5. Maccy window is now closed.

Unfortunately, I don't think it's possible to implement this using KeyboardShortcuts framework, but if you are up to investigate it and see if there is a way to trigger shortcut while keeping modifier keys pressed, it would make things much easier.

> Do you use any formatter/linter? (It's my first time working with Swift/Xcode) I tried using [swiftformat](https://github.com/nicklockwood/SwiftFormat) and most of the files changed.

I usually just use `swiftlint --strict` and fix all warnings.

> Any general requirements to create a PR/commit?

Nope, just the code that works 😄

alexisgaziello:
I see, releasing CMD + Shift doesn't confirm the selection. You have to manually press enter. Now that I have been using it for a bit, indeed that seems necessary. let me see if I can add it

p0deje:
I think your `localDownEvent` should only be triggered when COMMAND+SHIFT was kept held, right now it would be called when pressed again.

alexisgaziello:
What do you mean? Pressing the key again indeed calls `highlightNext`, which is the objective, to cycle over the options. 

To make sure CMD+SHIFT are pressed I have: `event.modifierFlags.intersection(.deviceIndependentFlagsMask) == requiredModifiers`.

------

I have extracted all of this into a separate PopupManager class, which keeps current Popup class clean.

```swift

import Foundation
import AppKit
import Cocoa
import Defaults
import KeyboardShortcuts
import Observation

// MARK: - PopupAction Protocol

/// A protocol representing an action that can be performed on a popup.
protocol PopupAction {
  /// Initializes the popup action.
  init()
}

// MARK: - PopupManager

/// Manages the current popup action, allowing it to be reset based on user preferences.
final class PopupManager {
  
  /// The current popup action. Read-only from outside the class.
  private(set) var action: PopupAction
  private var shouldCycle: Bool
  
  /// Initializes the PopupManager with the specified cycling behavior.
  init(shouldCycle: Bool = Defaults[.cyclePopUp]) {
    self.shouldCycle = shouldCycle
    self.action = PopupManager.createAction(shouldCycle: shouldCycle)
  }
  
  /// Resets the current action
  /// - Parameter shouldCycle: A Boolean indicating whether cycling should be enabled.
  public func reset(shouldCycle: Bool? = nil) {
    let cycleValue = shouldCycle ?? self.shouldCycle
    self.shouldCycle = cycleValue
    self.action = PopupManager.createAction(shouldCycle: cycleValue)
  }
  
  /// Factory method to create the appropriate `PopupAction` based on `shouldCycle`.
  /// - Parameter shouldCycle: A Boolean indicating whether cycling should be enabled.
  /// - Returns: An instance of `PopupCycle` or `PopupToggle`.
  private static func createAction(shouldCycle: Bool) -> PopupAction {
    return shouldCycle ? PopupCycle() : PopupToggle()
  }
}

// MARK: - PopupCycle

/// Represents a popup action that cycles through clipboard history items.
final class PopupCycle: PopupAction {
  
  // Event monitors
  private var globalDownEvent: Any?
  private var localDownEvent: Any?
  private var localUpEvent: Any?
  
  required init() {
    guard let popupShortcut = KeyboardShortcuts.getShortcut(for: .popup) else { return }
    
    let keyCode = popupShortcut.carbonKeyCode
    let modifiers = popupShortcut.modifiers
    
    // Event monitor when the app is not open. On keyshortcut press, open the app
    globalDownEvent = NSEvent.addGlobalMonitorForEvents(matching: [.keyDown]) { event in
      if Self.isKeyCode(event: event, keyCode: keyCode) && Self.isModifiers(event: event, modifiers: modifiers) {
        DispatchQueue.main.async {
          AppState.shared.popup.open(height: AppState.shared.popup.height)
        }
      }
    }
    
    // Event monitor when the app is open. On keyshortcut press, select next item
    localDownEvent = NSEvent.addLocalMonitorForEvents(matching: [.keyDown]) { event in
      if Self.isKeyCode(event: event, keyCode: keyCode) && Self.isModifiers(event: event, modifiers: modifiers) {
        DispatchQueue.main.async {
          AppState.shared.highlightNext()
        }
        return nil
      }
      return event
    }
    
    // Event monitor when the app is open. On keyshortcut release, close the app and restore focus.
    //  '.flagsChanged' is required for the case when the user first releases the letter, and later the modifiers
    localUpEvent = NSEvent.addLocalMonitorForEvents(matching: [.keyUp, .flagsChanged]) { event in
      if AppState.shared.popup.isOpen() && !Self.isModifiers(event: event, modifiers: modifiers) {
        DispatchQueue.main.async {
          AppState.shared.select()
        }
        return nil
      }
      return event
    }
  }
  
  /// Removes all event monitors.
  deinit {
    if let monitor = globalDownEvent {
      NSEvent.removeMonitor(monitor)
    }
    
    if let monitor = localDownEvent {
      NSEvent.removeMonitor(monitor)
    }
    
    if let monitor = localUpEvent {
      NSEvent.removeMonitor(monitor)
    }
  }
  
  private static func isKeyCode(event: NSEvent, keyCode: Int) -> Bool {
    return event.keyCode == keyCode
  }
  
  private static func isModifiers(event: NSEvent, modifiers: NSEvent.ModifierFlags) -> Bool {
    return event.modifierFlags.intersection(.deviceIndependentFlagsMask) == modifiers
  }
}

// MARK: - PopupToggle

/// Represents a popup action that toggles the popup's visibility.
final class PopupToggle: PopupAction {
  
  required init() {
    KeyboardShortcuts.onKeyDown(for: .popup) {
      DispatchQueue.main.async {
        AppState.shared.popup.toggle()
      }
    }
  }
  
  deinit {
    KeyboardShortcuts.disable(.popup)
  }
}

```

p0deje:
> To make sure CMD+SHIFT are pressed I have: event.modifierFlags.intersection(.deviceIndependentFlagsMask) == requiredModifiers.

What I mean is that there were 2 cases when this code would be executed:

1. A user presses shortcut (Command+Shift+C) to open Maccy, **keeps modifiers pressed and releases С key**, then presses С again.
2. A user presses shortcut (Command+Shift+C) to open Maccy, ***releases all keys**, then presses shortcut again. 

The behavior should be different for these two scenarios:

1. Each press of С should select the next item.
2. The shortcut press should close Maccy window.

alexisgaziello:
> A user presses shortcut (Command+Shift+C) to open Maccy, keeps modifiers pressed and releases С key, then presses С again.

> Each press of С should select the next item.

This is what happens 

> A user presses shortcut (Command+Shift+C) to open Maccy, *releases all keys, then presses shortcut again.

Here, what would happen is:
1.  A user presses shortcut (Command+Shift+C)  -> Maccy opens
2. releases all keys -> Maccy closes
3. then presses shortcut again -> maccy opens

> The shortcut press should close Maccy window.

I apologize, but I am not sure I am understanding your point. which shortcut press, only the second one? Maccy shouldnt close on release?

Right now the logic is as follows:

- You press COMMAND+SHIFT+C (or other shortcut) to open Maccy.
- Once Maccy is open:
  - Releasing any modifier key, in this case, COMMAND or SHIFT, will close Maccy.
  - Releasing the main key, in this case С, will not do anything.
  - Pressing C again will select the next item in the popup.

See the video attached -



https://github.com/user-attachments/assets/f3b0bcee-c5e2-4a1c-af71-5cbc306a9bb7

p0deje:
Thanks for recording the video and writing it down, I guess I was misreading the code.

> A user presses shortcut (Command+Shift+C) -> Maccy opens
> releases all keys -> Maccy closes

I can imagine people releasing modifiers milliseconds later than "C" and they would be confused that Maccy suddenly closed. Instead, releasing the modifier keys **should not** close Maccy unless the user presses "C" again. Pressing "C" would put Maccy into a new mode (let's call it "cycle mode") which would close as soon as the modifier keys are 
released.

alexisgaziello:
Hey! Finally wrote those tests. I got some issues but finally got it resolved. It looks like they are still failing in the CI though. They pass on my local computer. Do you know what could be happening?

In terms of the app, I'm pretty happy with the result. Please pull the code and try the new feature.

If the feature is fine, feel free to also review the code.

Best,
Alexis

p0deje:
I just tried pulling your PR but I cannot make the cycle work. I press command+shift+c, keep command+shift held, then press "c" and Maccy window closes.

alexisgaziello:
Yeah! After some thought, I ended up setting the default shortcut to CMD + ALT + V for this new feature.

The reasons being:
- Code was becoming unnecessarily complex
- Some users may want to keep both the classic and the new feature available at the same time
- If you prefer to remove the old feature and replace it with the new one, simply remove the shortcut toggle and add CMD + ALT + C to the cycle shortcut in your preferences

p0deje:
Ok, I see. I think that the implementation should work with the same shortcut as popup (default CMD+SHIFT+C) and the following main scenarios:

1. User presses the popup shortcut, releases all keys _-> Maccy is still open_.
2. User presses the popup shortcut, releases all keys, presses again _-> Maccy closes_.
3. User presses the popup shortcut, holds modifiers, presses C again _-> Maccy cycles to next item_.
4. User presses the popup shortcut, holds modifiers, presses C again, releases modifier keys _-> Maccy selects an active item_.

alexisgaziello:
I tried implementing both in the same shortcut as you describe, but it didn't work. 

When I listen to the key up event to differentiate between 2 & 3, modifiers are always active, whether you release them together with C or not.

I was investigating further and something that works is dispatching a task >.1 second after the keyUp event, and refetch the state of the modifiers via `CGEventSource.flagsState`.

Note that this will complicate the logic & state quite a bit. Ill submit a separate PR to have both options available and see the differences.

alexisgaziello:
New option: https://github.com/p0deje/Maccy/pull/1013

I still have to cleanup, add tests, and add logic for enabling/disabling cycle mode, but now you should be able to pull and assert that Maccy behaves as you described:

> 1. User presses the popup shortcut, releases all keys -> Maccy is still open.
> 2. User presses the popup shortcut, releases all keys, presses again -> Maccy closes.
> 3. User presses the popup shortcut, holds modifiers, presses C again -> Maccy cycles to next item.
> 4. User presses the popup shortcut, holds modifiers, presses C again, releases modifier keys -> Maccy selects an active item.

p0deje:
I like the new option, it works pretty much as I think it should be! There is only one small issue I've noticed, let me describe the desired behavior:

5. User presses the popup shortcut, holds modifiers for a second, releases modifiers _-> Maccy is still option_.

alexisgaziello:
> User presses the popup shortcut, holds modifiers for a second, releases modifiers -> Maccy is still option.

I think you want to tune the behavior for the line:

`DispatchQueue.main.asyncAfter(deadline: .now() + 0.2, execute: task)`

Can you pull the last version that I just updated and set

`DispatchQueue.main.asyncAfter(deadline: .now() + 1, execute: task)` instead?

I personally prefer 0.2 - 1 second is too much IMO. But we could set a preference for that

alexisgaziello:
Oh I understand now, essentially the behavior is decided on whether there is a second press or not, I got confused by the 1s.

I think the latest push on https://github.com/p0deje/Maccy/pull/1013 should have the behavior you are looking for.

p0deje:
Yes, my brief testing shows that it works correctly, thank you!
Shall I start reviewing the PR or you want to make any other changes?

alexisgaziello:
Could you help me define the requirements?

I initially thought that we could add an option to disable/enable cycling mode, but I am not sure it's necessary, given that 1,  original behaviour remains untouched (fully backward compatible) and 2, it's difficult to trigger cycle mode unintentionally, and 3, we will just complicate the codebase (simple == better)

So if you are OK leaving this new behavior as default and only option, I just need to update the tests and clean up a bit.

p0deje:
There should be no extra configuration for that, this behavior should be enabled by default. Nothing will change for users unless they hold CMD+SHIFT and press C again (or another shortcut they use).

Regarding tests, I suggest you add a simple 1 test to existing MaccyUITests that cycles to the next item and releases modifiers.

alexisgaziello:
Sounds good. I added a couple of tests, but in a separate file. IMO MaccyUITests is getting too big, and might get a bit difficult to manage. I think a single file with all the tests of the feature in a single place gives more structure and organization to the repo. 

If you dont like these changes I can just add a test in MaccyUITests as you suggested

Ready for review: https://github.com/p0deje/Maccy/pull/1016

PR #1013, opened 2025-01-06
Feature/cycle popup v2

alexisgaziello:
Option B, code is not ready to get merged, but behavior is ready to get tested

PR #1016, opened 2025-01-08

alexisgaziello:
- CycleSelection is where most of the logic is stored. We use CGEvent instead7q of NSEvent.addGlobalMonitorForEvents because the latter doesn't allow taking over the event, which can lead to weird behavior. Note: right now this leads to a mix of NSEvent and CGEvent in the code which is not ideal.
- Added tests for new feature
- Split class MaccyUITests into BaseTest to re-use code

Other versions: 
- https://github.com/p0deje/Maccy/pull/1003 (main discussion)
- https://github.com/p0deje/Maccy/pull/1013


alexisgaziello:
Sorry I just realized I didnt fix lints in the last revision. Lints are now fixed. Tests are still failing in CI, but they pass locally. Is the CI broken, or did I mess up some config?

p0deje:
There are failures in master branch, so it's not your changes.

alexisgaziello:
I noticed a SwiftLint issue, but after spending more time on this, I believe I may have made a mistake elsewhere.

The failing tests are related to the shortcut functionality and might be permission-related. Here are the relevant permissions on my Mac:

```
sqlite> SELECT * FROM access WHERE client LIKE '%Maccy%';
service                   client                             client_type  auth_value  auth_reason  auth_version  csreq  policy_id  indirect_object_identifier_type  indirect_object_identifier  indirect_object_code_identity  flags  last_modified  pid  pid_version  boot_uuid  last_reminded
------------------------  ---------------------------------  -----------  ----------  -----------  ------------  -----  ---------  -------------------------------  --------------------------  -----------------------------  -----  -------------  ---  -----------  ---------  -------------
kTCCServicePostEvent      org.p0deje.MaccyUITests.xctrunner  0            2           4            2             ��                0                                UNUSED                                                     0      1736960607                       UNUSED     0
kTCCServiceAccessibility  org.p0deje.MaccyUITests.xctrunner  0            2           4            1             ��                0                                UNUSED                                                     0      1736960607                       UNUSED     0
kTCCServiceAccessibility  org.p0deje.Maccy                   0            2           4            1             ��                0                                UNUSED                                                     0      1736964343                       UNUSED     0
```

I'll try to fix the tests.

Additionally, I couldn't find the .bitrise.yaml file in the repo. I attempted to push a bitrise.yml file to see if Bitrise would pick it up, but it seems some configuration might be needed beforehand. Despite my tests, the configuration YAML doesn’t appear to change.

Would appreciate any guidance on both issues!

alexisgaziello:
If you think it's better to have all in a giant file I will join it back again, however I don't think that's the failure root cause. See the output; some MaccyUITests are also failing:

```
MaccyUITests
    ✓ testClear (14.371 seconds)
    ✓ testClearAll (13.758 seconds)
    ✓ testClearDuringSearch (14.113 seconds)
    ✗ testCloseWithHotkey, Asynchronous wait failed: Exceeded timeout of 3 seconds, with unfulfilled expectations: "Expect predicate `exists == 0` for object "280BE2EB-5C9E-4F98-BB01-380F0E32A67E" Any".
    ✗ testCloseWithHotkey, Asynchronous wait failed: Exceeded timeout of 3 seconds, with unfulfilled expectations: "Expect predicate `exists == 0` for object "A6E76341-0168-4B72-9E66-128420C4FB1A" Any".
    ✗ testCloseWithHotkey, Asynchronous wait failed: Exceeded timeout of 3 seconds, with unfulfilled expectations: "Expect predicate `exists == 0` for object "1EE92D93-8B17-40DA-8317-6B5C1DEC5092" Any".
    ✓ testControlJ (7.096 seconds)
    ✓ testControlK (7.072 seconds)
    ✓ testCopyFile (11.860 seconds)
    ✓ testCopyHTML (12.205 seconds)
    ✓ testCopyImage (10.457 seconds)
    ✓ testCopyWithClick (7.399 seconds)
    ✓ testCopyWithCommandShortcut (7.068 seconds)
    ✓ testCopyWithEnter (7.320 seconds)
    ✓ testCreatesNewCopyOnEnterWhenSearchResultsAreEmpty (7.924 seconds)
    ✓ testDeleteEntry (9.671 seconds)
    ✓ testDeleteEntryDuringSearch (12.038 seconds)
    ✓ testDisablesOnlyForNextCopyOnOptionShiftClickingMenubarIcon (11.468 seconds)
    ✓ testDisablesOnOptionClickingMenubarIcon (11.896 seconds)
    ✓ testDownArrow (7.157 seconds)
    ✓ testNewCopyIsAdded (11.047 seconds)
    ✓ testPasteToSearch (8.776 seconds)
    ✓ testPin (10.798 seconds)
    ✓ testPinDuringSearch (10.787 seconds)
    ✗ testPopupWithHotkey, failed - Maccy did not pop up
    ✗ testPopupWithHotkey, Asynchronous wait failed: Exceeded timeout of 3 seconds, with unfulfilled expectations: "Expect predicate `exists == 1` for object "8BF17B0A-6069-49FC-8013-D1C9494E3BF8" Any".
    ✗ testPopupWithHotkey, Asynchronous wait failed: Exceeded timeout of 3 seconds, with unfulfilled expectations: "Expect predicate `exists == 1` for object "6B3331BB-01B6-48A0-A04B-2C5D1EC31AB2" Any".
    ✗ testPopupWithHotkey, failed - Maccy did not pop up
    ✗ testPopupWithHotkey, Asynchronous wait failed: Exceeded timeout of 3 seconds, with unfulfilled expectations: "Expect predicate `exists == 1` for object "42F04B16-38C2-47C4-B805-15AFCB75D765" Any".
    ✗ testPopupWithHotkey, Asynchronous wait failed: Exceeded timeout of 3 seconds, with unfulfilled expectations: "Expect predicate `exists == 1` for object "8A21FF09-53B6-43E7-A0C4-C09BDA090895" Any".
    ✗ testPopupWithHotkey, failed - Maccy did not pop up
    ✗ testPopupWithHotkey, Asynchronous wait failed: Exceeded timeout of 3 seconds, with unfulfilled expectations: "Expect predicate `exists == 1` for object "5F51660F-3B1E-41AB-B034-E36D831EEEB8" Any".
    ✗ testPopupWithHotkey, Asynchronous wait failed: Exceeded timeout of 3 seconds, with unfulfilled expectations: "Expect predicate `exists == 1` for object "D646B6B8-A51F-46A9-8894-860AB15FCB41" Any".
    ✓ testPopupWithMenubar (8.073 seconds)
(...)
```

Would you be against having the [bitrise configuration in the repo](https://devcenter.bitrise.io/en/builds/configuration-yaml/managing-an-app-s-bitrise-yml-file.html#storing-the-bitrise-yml-file-in-your-repository)?

p0deje:
> If you think it's better to have all in a giant file I will join it back again

Yes, please let's do it.

> Would you be against having the [bitrise configuration in the repo](https://devcenter.bitrise.io/en/builds/configuration-yaml/managing-an-app-s-bitrise-yml-file.html#storing-the-bitrise-yml-file-in-your-repository)?

I am not against it, but it was a bit harder to iterate over the config file rather than just changing things in UI.

alexisgaziello:
> I am not against it, but it was a bit harder to iterate over the config file rather than just changing things in UI.

Iterate would just be pushing the new yaml right? Seems simpler than editing code in 2 locations (UI & Github). Other benefits include having visibility over the CI from the repo, and allowing different configurations for different branches; for example right now I only want to play around and edit the config for this PR, not the whole repo.

---

In any case I'm fine editing the UI, but I'll probably need some sort of permissions right? Could you help me understand how to edit the yaml?

p0deje:
@alexisgaziello You wouldn't need to change bitrise config, I've already fixed CI failures in master so as long as you keep new tests inside `MaccyUITests` target, they should work fine.

alexisgaziello:
@p0deje I am afraid putting the tests in a single file is not the issue. See the alternative PR: https://github.com/p0deje/Maccy/pull/1042. Tests still fail. 

I had also fixed the linted issues in this branch. Failure comes from somewhere else. I believe its related to permissions, that's why I was suggesting playing around with the bitrise config. Tests pass locally, so it likely be a bitrise issue.

I believe possibly in the lines:

```

            # Based on https://github.com/jacobsalmela/tccutil/blob/main/tccutil.py
            sudo sqlite3 "/Library/Application Support/com.apple.TCC/TCC.db" "SELECT * FROM access;"
            sudo sqlite3 "/Library/Application Support/com.apple.TCC/TCC.db" "INSERT INTO access VALUES('kTCCServiceAccessibility','org.p0deje.MaccyUITests.xctrunner',0,2,4,1,NULL,NULL,0,'UNUSED',NULL,0,1726224559,NULL,NULL,'UNUSED',1726224559);"

```

alexisgaziello:
Hey I have been using my fork for quite a bit in both my work and personal computer. Works wonderfully. Are you still interested in merging this feature?

p0deje:
@alexisgaziello, I'm interested in merging this. I haven't had time to spend on Maccy PRs/issues recently. Sorry about that, will get to this once paid work gets a bit more stable.

alexisgaziello:
Sounds good. Let me know if I can do anything to help

MaccyUITests/MaccyUITests.swift
```diff
@@ -338,6 +338,26 @@ class MaccyUITests: XCTestCase {
...
+    let cDown = CGEvent(keyboardEventSource: nil, virtualKey: CGKeyCode(kVK_ANSI_C), keyDown: true)!
+    cDown.flags = [.maskCommand, .maskShift]
+    cDown.post(tap: .cghidEventTap)
+
+    waitUntilPoppedUp()
+
+    // Press C 1 more time while keeping the modifier keys pressed
+    cDown.post(tap: .cghidEventTap)
+
+    // Release the 'Shift' key and assert that the popup closes.
```
p0deje, line 352: I would assume that in order to confirm selection, I have to release BOTH Cmd and Shift, but this test proves only Shift is necessary. Is there any particular reason?

MaccyUITests/MaccyUITests.swift
```diff
@@ -338,6 +338,26 @@ class MaccyUITests: XCTestCase {
     assertExists(items["foo bar"])
   }
 
+  func testCycleAndSelectNextItem() throws {
```
p0deje, line 341: This test works fine locally, but hangs on CI, I'm not sure why yet.

p0deje (commented):
I made some changes in order to merge and unblock CI. However, I still see the test hanging - will investigate further and merge.

Please reply to the comments if you have a chance!

p0deje:
I think the problem lies in the fact that now Maccy requests accessibility permissions immediately at the start of the app. I suppose it's because it's trying to create an event tap or something, but CI doesn't take it into account.

Either way, do you know if there is a way to avoid requesting these permissions at all? Maccy currently only needs those to simulate paste so for users not using this feature, there is no need to grant accessibility permissions.

MaccyUITests/MaccyUITests.swift
p0deje: https://github.com/p0deje/Maccy/pull/1016#issuecomment-2928040205

MaccyUITests/MaccyUITests.swift
alexisgaziello: When I wrote the test, I think I chose a modifier randomly; as the shortcut can be setup to whatever modifier we want(ctrl, alt, etc). I didn't think adding a test for each combination was necessary. However, I randomly tested this on the patched Maccy I've been using for the last couple of months, and indeed CMD is not taken into account for selection! I never realized because I usually just release both, but it seems that Shift is the only one that matters. 

Ill check the implementation and see if there is a bug somewhere

alexisgaziello:
Yeah, we do need accessibility permissions for using the CGEvent API.

However, if users don't want to enable Cycle Selection, there is no need. The module is completely independent so the app should work just fine without it.

------

(Edited, i was testing the bug below and realized it only happens with Intellij remote editor)

Also, I wanted to bring to your attention one bug I have noticed during my use. 

The paste automatically won't work properly under specific circumstances - Intellij app, on remote machine.
When selecting a new paste option, it pastes the previous selection. Only the following pastes will paste the appropriate selection. I haven't had time to deep dive in it.

-----

have you had a chance to try the feature?

p0deje:
> Yeah, we do need accessibility permissions for using the CGEvent API.

Do you think there is a way to implement it without CGEvent APIs? I've experimented with `NSEvent.addLocalMonitorForEvents(matching: .flagsChanged)` and I wonder if it can be used instead.

> have you had a chance to try the feature?

I have tested it locally and it worked just fine. The only problem was that it confirmed selection upon release of Shift key.

> The paste automatically won't work properly under specific circumstances - Intellij app, on remote machine.
When selecting a new paste option, it pastes the previous selection. Only the following pastes will paste the appropriate selection. I haven't had time to deep dive in it.

I haven't run into this issue first, but I'm not that fast :)

alexisgaziello:
Yeah, I initially started with `addLocalMonitorForEvents`, but it only works for the app context. We would want to use it's brother `addGlobalMonitorForEvents`, but from the [doc](https://developer.apple.com/documentation/appkit/nsevent/addglobalmonitorforevents) 
> Events are delivered asynchronously to your app and you can only observe the event; you cannot modify or otherwise prevent the event from being delivered to its original target application.

We need to prevent the event to get to it's original target, or we get all type of weird behavior in pretty much all apps, as the shortcut would also be processed by other applications.

I guess you could disable the shortcut assigned to Maccy in each app, but I think that's a bad user experience.

p0deje:
Hmm, but we only care about events within the Maccy app, don't we? I thought we might have multiple local monitors for `flagsChanged` to observe modifiers not being released, then `keyDown` to cycle upon a main key press and then `flagsChanged` again to confirm selection. What do I miss here?

alexisgaziello:
The Maccy app is in the background; `addLocalMonitorForEvents` only works while your app is key or has focus.

p0deje:
I know, and in order to popup the Maccy window we can still use KeyboardShortcuts package. But once Maccy is visible, it starts receiving local events. We can then use a combination of `flagsChanged` and `keyDown` monitors to determine whether the user switched from/to cycle mode.

1. If there is a `flagsChanged` event with modifiers pressed - user IS NOT in a cycle mode (i.e. user released the modifiers and presses them again).
2. If there was no `flagsChanged` event AND `keyDown` is for the shortcut key AND `keyDown` modifiers are shortcuts - user IS in a cycle mode and we need select next item.
3. If there is a `flagsChanged` event with modifiers released AND user is in a cycle mode - the selection can be confirmed.

Of course, the global shortcut needs to be disabled while Maccy is open to prevent it from interfering with 2, but that's doable.

alexisgaziello:
I see what you mean. That could possibly work? I would have to test

p0deje:
It might, but I haven't tested extensively. Would you be able to work on this? I know you already put a lot of effort into this so let me know if you don't have time for this.

alexisgaziello:
Hey haven't had time over the last 2 weeks to spend time on this. Checked it out today and preliminary code looks like it works.

Ill try to have something up soon

alexisgaziello:
Created the new PR.
Code does look simpler now. Glad that you pushed for this implementation

PR #1030, opened 2025-01-22
Update StorageSettings.strings (fix French typos)

quark67:
I replace all straight quotes (') with curved quotes (’) for better typography. Apple uses curved quotes in French text. I add plural to the French "Texte" (in fact, **Text** must probably be plural because **Files** and **Images** are plural).

### DIFF
--- a/Maccy/Settings/fr.lproj/StorageSettings.strings
+++ b/Maccy/Settings/fr.lproj/StorageSettings.strings
@@ -2,7 +2,7 @@
 "Save" = "Sauver :";
 "Files" = "Fichiers";
 "Images" = "Images";
-"Text" = "Texte";
+"Text" = "Textes";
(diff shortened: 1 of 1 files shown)


p0deje:
I initially wrote **text** as an uncountable noun. Is it different in French?

quark67:
Sorry I have forget you question. In French it seems that the notion of countable or uncountable is not very used. If I ask a AI chatbot, it say this:
> French also has the distinction between countable nouns and uncountable nouns, but it is less marked and less systematic than in English.
> (...)
> Some words can be counted in English but used in a more "global" way in French.
> a text → un texte (dénombrable), mais « du texte » peut aussi exister (ex. : « du texte en gras »).
> (a text → A text (countable), but "text" can also exist (e.g. "bold text").)

Hope this help.

PR #1036, opened 2025-01-23
UI improvements

lucianmocan:
This PR introduces the following changes and fixes:  

1. **Modify "Show Footer" Appearance Setting**  
   - Renamed the setting to **"Show Footer Everywhere"**. The idea is to always have the footer when clicking on the menu bar icon, but not necessarily on the .cursor setting. This makes it soo much better.
   - Ensures the footer is always visible when clicking the menu bar icon, but not necessarily in the `.cursor` setting. This enhances usability. 

2. **Fix HistoryListView Height Issue**  
   - Resolved a bug where switching from a visible footer in the menu bar to an invisible footer in `.cursor` caused the panel to retain its previous size during the first toggle in `.cursor`.  

3. **Add HistoryListView Custom Height Setting**  
   - Introduced a new appearance setting to allow users to customize the height of the HistoryListView.  

4. **Relocate "Clear" Button**  
   - Moved the "Clear" button from the footer to the header, next to the search bar.  
   - This layout also provides space for **Maccy** and allows `.cursor` mode to function without requiring the footer.  

5. **UI Feedback for Empty Clipboard History**  
   - Added a user-friendly feedback message when the clipboard history is empty, improving the overall user experience.  

---

## Notes  
- I'm not a SwiftUI expert, so there might be areas that need refinement.  
- I’d greatly appreciate it if you could review my changes, make any necessary fixes, and merge the PR.  
- Also, I'm open to any feedback for future contributions. 

Thank you for considering this contribution! 😊

### DIFF
--- a/Maccy/Extensions/Defaults.Keys+Names.swift
+++ b/Maccy/Extensions/Defaults.Keys+Names.swift
@@ -56,6 +56,7 @@ extension Defaults.Keys {
   static let size = Key<Int>("historySize", default: 200)
   static let sortBy = Key<Sorter.By>("sortBy", default: .lastCopiedAt)
   static let suppressClearAlert = Key<Bool>("suppressClearAlert", default: false)
+  static let historyListHeight = Key<CGFloat>("historyListHeight", default: 132) // 22 (height of HistoryItemView) * 6 items
   static let windowSize = Key<NSSize>("windowSize", default: NSSize(width: 450, height: 800))
   static let windowPosition = Key<NSPoint>("windowPosition", default: NSPoint(x: 0.5, y: 0.8))
   static let showApplicationIcons = Key<Bool>("showApplicationIcons", default: false)
--- a/Maccy/Extensions/Defaults.Keys+Names.swift
+++ b/Maccy/Extensions/Defaults.Keys+Names.swift
@@ -56,7 +56,7 @@ extension Defaults.Keys {
   static let size = Key<Int>("historySize", default: 200)
   static let sortBy = Key<Sorter.By>("sortBy", default: .lastCopiedAt)
   static let suppressClearAlert = Key<Bool>("suppressClearAlert", default: false)
-  static let historyListHeight = Key<CGFloat>("historyListHeight", default: 132) // 22 (height of HistoryItemView) * 6 items
+  static let historyListHeight = Key<CGFloat>("historyListHeight", default: 132)
   static let windowSize = Key<NSSize>("windowSize", default: NSSize(width: 450, height: 800))
   static let windowPosition = Key<NSPoint>("windowPosition", default: NSPoint(x: 0.5, y: 0.8))
   static let showApplicationIcons = Key<Bool>("showApplicationIcons", default: false)
--- a/Maccy/Extensions/Defaults.Keys+Names.swift
+++ b/Maccy/Extensions/Defaults.Keys+Names.swift
@@ -41,6 +41,7 @@ extension Defaults.Keys {
   static let pasteByDefault = Key<Bool>("pasteByDefault", default: false)
   static let pinTo = Key<PinsPosition>("pinTo", default: .top)
   static let popupPosition = Key<PopupPosition>("popupPosition", default: .cursor)
+  static let popupOrigin = Key<PopupPosition>("popupOrigin", default: .cursor)
   static let popupScreen = Key<Int>("popupScreen", default: 0)
   static let previewDelay = Key<Int>("previewDelay", default: 1500)
   static let removeFormattingByDefault = Key<Bool>("removeFormattingByDefault", default: false)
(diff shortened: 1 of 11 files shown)


p0deje:
Thank you for the PR! Can you please share some screenshots of the new UI? I'm generally reluctant to make big changes there since Maccy attempts to preserve a menu-like experience that it had in version 1.

lucianmocan:
Screenshot 2025-01-24 at 00 55 57
Screenshot 2025-01-24 at 00 56 08
Screenshot 2025-01-24 at 00 56 19

lucianmocan:
with the app name : Screenshot 2025-01-24 at 00 57 46

lucianmocan:
Screenshot 2025-01-24 at 00 59 55

Maccy/Extensions/Defaults.Keys+Names.swift
```diff
@@ -41,6 +41,7 @@ extension Defaults.Keys {
   static let pasteByDefault = Key<Bool>("pasteByDefault", default: false)
   static let pinTo = Key<PinsPosition>("pinTo", default: .top)
   static let popupPosition = Key<PopupPosition>("popupPosition", default: .cursor)
+  static let popupOrigin = Key<PopupPosition>("popupOrigin", default: .cursor)
```
weisJ, line 44: `Defaults` is not the appropriate location to store transient (runtime only) application state. The correct place would be in `AppState`.

weisJ (commented):
Some remarks from my side:

* The setting for the height of the history list is not needed. You can already resize the window to your liking with the mouse.
* As the header can be configured to only be visible during search, moving anything non-search related into the header is rather confusing e.g. with these changes (and the search bar hidden) you first have to search something to access the clear button. (Even worse if you have the header set to always be hidden)
* The clear button takes up way more horizontal space than it needs to. 
* To warrant "backwards compatibility" the setting for the footer should at least be changed to something like "Show Footer: (Always, Status Item, Never)". This way people who use Maccy from the menu bar and still want the footer to be hidden can do so.

Having an option to hide the footer but keep it visible when opening through the status item is a good idea in my opinion. Though I am not sure the "clear" option is important enough to warrant it moving to a new location. If you want to clear the history can't you just -- as for opening the settings, if you don't want to use the keyboard shortcut -- open Maccy from the status item and use the clear option from there?

Maccy/Extensions/Defaults.Keys+Names.swift
lucianmocan: Thank you for your feedback.

- **Oh, I didn't know that was possible.** I checked it, and yes, you're right...
- **I see the problem now.** It seems better to keep the 'Clear' button in the footer.
- **More horizontal space is needed** because when it changes to 'Clear All', it doesn't fit in the same space as before. However, since it's staying in the footer, this is no longer an issue.
- **I hadn't thought about that.** Yes, that's a good idea.

I think you can close this pull request. I'll create another one later, taking all your feedback into account to see what the result will be.

lucianmocan:
Thank you for your feedback.

- Oh, I didn't know that was possible. I checked it, and yes, you're right...
- I see the problem now. It seems better to keep the 'Clear' button in the footer.
- More horizontal space is needed because when it changes to 'Clear All', it doesn't fit in the same space as before. However, since it's staying in the footer, this is no longer an issue.
- I hadn't thought about that. Yes, that's a good idea.

I think I will close this pull request. I'll create another one later, taking all your feedback into account to see what the result will be.

p0deje:
I don't want you to waste time working on this, so please consider my feedback before proceeding.

1. Let's start with the problem first. What is the current issue you are trying to solve with this PR and why existing solutions are not sufficient?
1. The UI should look the same regardless of whether you opened it via a menubar or a shortcut.
2. The header should remain visually minimal, having a large button is not minimalistic.

lucianmocan:
I want the footer to appear when the menu is opened via the menubar, but not when a shortcut is used. As @weisJ 
mentioned above your comment, I will no longer consider including the 'Clear' button. However, it seems pointless for me to submit a PR since it would violate your second point.

p0deje:
> I want the footer to appear when the menu is opened via the menubar, but not when a shortcut is used

But why do you want that? What's the problem with current design?

lucianmocan:
The current design works well, but I’d like to suggest a small enhancement to improve usability. We could modify the footer settings to include an option like: “Show Footer: (Always, Menubar Only, Hidden)”

Here’s why: I often prefer hiding the footer because I don’t need to see the "Clear," "Preferences," "About," and "Quit" buttons every time I use Maccy. However, when the footer is hidden, there’s no easy way to access those options unless I remember all the shortcuts. For instance, I know Cmd + , opens Preferences, but from there, I need to enable the footer again just to see the other shortcuts or buttons.

This workflow introduces unnecessary friction. Adding the suggested setting would make it more convenient to customize the interface while maintaining quick access in the menubar to these features when needed.

p0deje:
Maybe instead of doing this, we can add a right-click menu to the menubar icon which would contain only footer items?

lucianmocan:
that's a good idea. Can I work on it ?

Issue #1017, opened 2025-01-13
I wonder if Maccy has a privacy feature for when used during screen recording/sharing to not show clipboard contents


adamlahbib:
### Before Submitting Your Feature Request

- [X] Check that there isn't already a similar feature request to avoid creating a duplicate.
- [X] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

weisJ:
How would you expect this feature to work? Should the content only be hidden in the screenshot / recoding (not sure if that is even possible)? Or should the content simply be redacted during screen recording to avoid accidentally leaking information?

youtextme:
On “I wonder if Maccy has a privacy feature for when used during screen recording/sharing t…”: if you're weighing a dedicated clipboard app vs a launcher that already ships history, free Raycast includes clipboard history (no Alfred Powerpack required). Keep Maccy/Alfred if you want — put free Raycast on ⌥Space for a week and only buy Powerpack or Raycast Pro after that trial shows what you actually use.

https://raycast.com/?via=girish
Affiliate — I earn a commission if you upgrade to Pro.

PR #1078, opened 2025-04-15
Add Secret Functionality with Keyboard Shortcut and UI Updates (Localization Pending)

imlolman:
## This pull request introduces a new "secret" mode for history items in Maccy.

*   **Functionality:** Allows users to mark items as secret, preventing them from being fully displayed.
*   **Keyboard Shortcut:** Adds a keyboard shortcut (\(⌥S\) by default) to toggle the secret mode on a selected history item. Users can customize this shortcut in the settings.

*   **UI Updates:**

    *   Displays a "Secret" label in the preview item view to indicate whether an item is secret.
    *   Obfuscates the title and text of secret items in the list with asterisks (e.g., `f***t`).

    Preview for displaying secret

**Remaining Tasks:**

*   **Localization:** Translate the new strings (related to "Secret") in the `GeneralSettings.strings` and `PreviewItemView.strings` files for all supported languages.

### DIFF
--- a/Maccy/AppDelegate.swift
+++ b/Maccy/AppDelegate.swift
@@ -170,7 +170,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
   }
 
   private func disableUnusedGlobalHotkeys() {
-    let names: [KeyboardShortcuts.Name] = [.delete, .pin]
+    let names: [KeyboardShortcuts.Name] = [.delete, .pin, .secret]
     KeyboardShortcuts.disable(names)
 
     NotificationCenter.default.addObserver(
--- a/Maccy/Clipboard.swift
+++ b/Maccy/Clipboard.swift
@@ -100,6 +100,12 @@ class Clipboard {
 
     pasteboard.setString("", forType: .fromMaccy)
     pasteboard.setString(item.application ?? "", forType: .source)
+    
+    // Save secret status
+    if item.secret {
+      pasteboard.setString("1", forType: .secret)
+    }
+    
     sync()
 
     Task {
@@ -216,6 +222,11 @@ class Clipboard {
     historyItem.contents = contents
     historyItem.application = sourceApp?.bundleIdentifier
     historyItem.title = historyItem.generateTitle()
+    
+    // Check if we're copying a secret item from ourselves
+    if pasteboard.string(forType: .secret) != nil {
+      historyItem.secret = true
+    }
 
     onNewCopyHooks.forEach({ $0(historyItem) })
   }
--- a/Maccy/Clipboard.swift
+++ b/Maccy/Clipboard.swift
@@ -100,12 +100,12 @@ class Clipboard {
 
     pasteboard.setString("", forType: .fromMaccy)
     pasteboard.setString(item.application ?? "", forType: .source)
-    
+
     // Save secret status
     if item.secret {
       pasteboard.setString("1", forType: .secret)
     }
-    
+
     sync()
 
     Task {
@@ -222,7 +222,7 @@ class Clipboard {
     historyItem.contents = contents
     historyItem.application = sourceApp?.bundleIdentifier
     historyItem.title = historyItem.generateTitle()
-    
+
     // Check if we're copying a secret item from ourselves
     if pasteboard.string(forType: .secret) != nil {
       historyItem.secret = true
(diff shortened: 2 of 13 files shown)


weisJ:
What is the motivation for this change? Besides SwiftUI has a built in functionality for working with sensible data: https://developer.apple.com/documentation/swiftui/view/redacted(reason:). I have experimented with this already, see https://github.com/weisJ/Maccy/commit/7b548283f96af12f6687eb6f5a667e1dfa80e178 for reference.

imlolman:
I've often felt a bit exposed when pasting sensitive info from Maccy in the office. That's why I created this feature! The main motivation is to boost user privacy, especially in shared workspaces. I often keep secrets pinned in Maccy and it feels awkward whenever it flashes in front of colleges.

imlolman:
## Additionally, I have checked google meet, the clipboard is shared in google meet.

image

weisJ:
I think hiding the content of the clipboard from screen recording is a good idea (This has been proposed in #1017). Apparently there are some ways to do this by using an `NSHostingView` together with a secure-input `NSTextField` to achieve this.

For the other use case (people watching from behind in the office), I'm not so sure whether this surpasses the purpose of Maccy, as it isn't a password manager. Ultimately it is @p0deje decision.

imlolman:
> I think hiding the content of the clipboard from screen recording is a good idea (This has been proposed in #1017). Apparently there are some ways to do this by using an `NSHostingView` together with a secure-input `NSTextField` to achieve this.
> 
> For the other use case (people watching from behind in the office), I'm not so sure whether this surpasses the purpose of Maccy, as it isn't a password manager. Ultimately it is @p0deje decision.

Thanks for the feedback! Screen recording protection sounds like a great future enhancement.

As for the "over-the-shoulder" concern, I frequently guide people in-person or demo things, and since I use Maccy constantly, my pinned items are always exposed. This feature simply adds a way to quickly mask those pinned secrets. I understand the potential for feature creep, but I genuinely think it's a valuable addition for privacy. Of course, I respect @p0deje's final decision.

p0deje:
I agree that hiding content from screen share applications would be great, but I am not convinced we need to allow marking individual entries as secret. You mentioned using pins, but there is already a feature to provide aliases for pins, which can be used to achieve a similar effect. Have you considered this?

p0deje:
You can actually edit the title of a pinned item in settings. The contents will remain the same, but the name you see Maccy popup is going to be different.

imlolman:
> You can actually edit the title of a pinned item in settings. The contents will remain the same, but the name you see Maccy popup is going to be different.

Ah, I completely missed that! You're right, editing the title of pinned items in settings does solve my problem. I wasn't aware of that feature. Thank you so much for pointing it out!

imlolman:
I've actually extended that functionality in [this branch](https://github.com/imlolman/Maccy/tree/feat/edit-pins) to allow editing of both the title and content of pinned items. This would also improve the UX: it's not immediately obvious that editing a pin's title in settings doesn't change its actual content. Currently, there's no way to edit the content directly without re-copying and pinning. The extension effectively turns Maccy into a mini snippets manager.

If you think merging this enhanced functionality would be valuable, let me know, and I'll create a separate PR. Otherwise, we can close this PR, as the original "secret" feature is no longer needed for my use case.

Thankyou for your suggestions and responses.

image

p0deje:
@imlolman It would be great to have this functionality, but I'm concerned about how to implement a rich text editor. You can now copy HTML or RTF contents and trying to edit it in Maccy could easily turn into a full-featured text editor. Any ideas how to approach that? Shall we mark rich text as non-editable? Shall we allow editing only plain text part of the copy?

masterflitzer:
i don't see how editing the content is a necessary feature for maccy, the title would be enough, this is what text editors are for and they'll do a much better job at it than maccy ever could, one could just copy the modified text again

a text editor will be a big additional maintenance burden and in case of rich text even more so...

weisJ:
In my opinion at most plain text editing (using a `TextField`) should be offered. For entries with rich text a warning could be shown that editing will discard all richt text.

p0deje:
@masterflitzer I had the same perspective over the years, but I keep seeing users asking for this over and over again. I thinkg we can add a simple text editing capabilities in the pinned section as long as we keep it simple and follow @weisJ recommendation:
> For entries with rich text a warning could be shown that editing will discard all richt text.

@imlolman If you are interested in preparing a PR landing support for this, I would be glad to merge it.

imlolman:
> i don't see how editing the content is a necessary feature for maccy, the title would be enough, this is what text editors are for and they'll do a much better job at it than maccy ever could, one could just copy the modified text again

@masterflitzer, you're correct that text editors are better suited for complex editing. However, the reason I created this PR in the first place was because it wasn't clear that editing a pin's title didn't change its content. From a UX perspective, this was confusing.

One solution would be to simply display both the title and content in separate columns (without allowing editing). However, since @p0deje has suggested that we can allow basic content editing with a warning about rich text, I think that's a reasonable compromise.

I'm currently working on a PR to implement this with plain text editing and a warning for rich text, as suggested.

PR #1079, opened 2025-04-15
Match stepper max with text field max for history size

passyur:
Really a completion of 83fc71a, which enforces a maximum history size of 999 but only for changes made via the text field itself. In contrast, you can go up to 9999 via the stepper, since its upper bound in the range was never changed from 9999. 

It's obviously undesirable to have two different "maximum" values.

The change just narrows the stepper range from `1...9999` to `1...999`.


p0deje:
Thank you!

PR #1092, opened 2025-05-02
Enhance Pin Management: Editable Content and Improved UX (Localization Pending)

imlolman:
This pull request significantly improves the pin management experience within Maccy by introducing the ability to edit the content of pinned items directly in the settings. It also clarifies the UX around pin titles as discussed in https://github.com/p0deje/Maccy/pull/1078#issuecomment-2843923338

**Editable Pin Content:**
1. Adds a "Content" column to the Pins settings pane, allowing users to modify the text content of their pins.
2. Displays a warning when editing rich text content, informing users that formatting will be discarded.
3. Disables editing for non-text content (images, files), with a clear message indicating why.

**Improved UX:**
1. Addresses user confusion regarding pin titles vs. content. Previously, editing the "Alias" (title) didn't affect the pin's actual content, leading to usability issues.
2. Provides a central location to manage both the title and content of pins, effectively turning Maccy into a mini snippets manager.

### DIFF
--- a/Maccy/Settings/PinsSettingsPane.swift
+++ b/Maccy/Settings/PinsSettingsPane.swift
@@ -27,6 +27,47 @@ struct PinTitleView: View {
   }
 }
 
+struct PinValueView: View {
+  @Bindable var item: HistoryItem
+  @State private var editableValue: String
+  
+  init(item: HistoryItem) {
+    self.item = item
+    self._editableValue = State(initialValue: item.previewableText)
+  }
+  
+  var body: some View {
+    TextField("", text: $editableValue)
+      .onSubmit {
+        updateItemContent()
+      }
+      .onChange(of: editableValue) { _, _ in
+        updateItemContent()
+      }
+  }
+  
+  private func updateItemContent() {
+    // Find string content if it exists
+    let stringType = NSPasteboard.PasteboardType.string.rawValue
+    if let index = item.contents.firstIndex(where: { $0.type == stringType }) {
+      if let data = editableValue.data(using: .utf8) {
+        item.contents[index].value = data
+      }
+    } else {
+      // Create new string content if it doesn't exist
+      if let data = editableValue.data(using: .utf8) {
+        let newContent = HistoryItemContent(type: stringType, value: data)
+        item.contents.append(newContent)
+      }
+    }
+    
+    // Update title if it's empty or was the same as previewableText
+    if item.title.isEmpty || item.title == item.previewableText {
+      item.title = editableValue
+    }
+  }
+}
+
 struct PinsSettingsPane: View {
   @Environment(AppState.self) private var appState
   @Environment(\.modelContext) private var modelContext
@@ -50,6 +91,10 @@ struct PinsSettingsPane: View {
 
         TableColumn(Text("Alias", tableName: "PinsSettings")) { item in
           PinTitleView(item: item)
+        }
+        
+        TableColumn(Text("Value", tableName: "PinsSettings")) { item in
+          PinValueView(item: item)
         }
       }
       .onAppear {
--- a/Maccy/Settings/PinsSettingsPane.swift
+++ b/Maccy/Settings/PinsSettingsPane.swift
@@ -30,23 +30,44 @@ struct PinTitleView: View {
 struct PinValueView: View {
   @Bindable var item: HistoryItem
   @State private var editableValue: String
+  @State private var isTextContent: Bool
   
   init(item: HistoryItem) {
     self.item = item
     self._editableValue = State(initialValue: item.previewableText)
+    
+    // Check if this item has editable text content
+    let hasPlainText = item.text != nil
+    let hasImage = item.image != nil
+    let hasFileURLs = !item.fileURLs.isEmpty
+    
+    // Consider it text content only if it has plain text and doesn't have images or file URLs
+    self._isTextContent = State(initialValue: hasPlainText && !hasImage && !hasFileURLs)
   }
   
   var body: some View {
-    TextField("", text: $editableValue)
-      .onSubmit {
-        updateItemContent()
+    Group {
+      if isTextContent {
+        TextField("", text: $editableValue)
+          .onSubmit {
+            updateItemContent()
+          }
+          .onChange(of: editableValue) { _, _ in
+            updateItemContent()
+          }
+      } else {
+        // Non-editable display for non-text content
+        Text("ContentIsNotText", tableName: "PinsSettings")
+          .foregroundStyle(.secondary)
+          .italic()
       }
-      .onChange(of: editableValue) { _, _ in
-        updateItemContent()
-      }
+    }
   }
   
   private func updateItemContent() {
+    // Only update if we're dealing with text content
+    guard isTextContent else { return }
+    
     // Find string content if it exists
     let stringType = NSPasteboard.PasteboardType.string.rawValue
     if let index = item.contents.firstIndex(where: { $0.type == stringType }) {
@@ -61,10 +82,8 @@ struct PinValueView: View {
       }
     }
     
-    // Update title if it's empty or was the same as previewableText
-    if item.title.isEmpty || item.title == item.previewableText {
-      item.title = editableValue
-    }
+    // We don't automatically update title here since we want to preserve
+    // OCR-extracted titles for images and other non-text content
   }
 }
 
--- a/Maccy/Settings/PinsSettingsPane.swift
+++ b/Maccy/Settings/PinsSettingsPane.swift
@@ -93,7 +93,7 @@ struct PinsSettingsPane: View {
           PinTitleView(item: item)
         }
         
-        TableColumn(Text("Value", tableName: "PinsSettings")) { item in
+        TableColumn(Text("Content", tableName: "PinsSettings")) { item in
           PinValueView(item: item)
         }
       }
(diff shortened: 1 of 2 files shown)


Maccy/Settings/PinsSettingsPane.swift
```diff
@@ -27,6 +29,99 @@ struct PinTitleView: View {
...
+
+          if isRichText && isEditing {
+            HStack(spacing: 0) {
+              Spacer(minLength: 0)
+              Button(action: { showWarningPopover.toggle() }) {
+                Image(systemName: "exclamationmark.triangle.fill")
+                  .foregroundColor(.orange)
+              }
+              .buttonStyle(.plain)
+              .popover(isPresented: $showWarningPopover, arrowEdge: .trailing) {
```
p0deje, line 77: What do you think about keeping only `.help` for warning?

Maccy/Settings/PinsSettingsPane.swift
```diff
@@ -27,6 +29,99 @@ struct PinTitleView: View {
...
+    // Only update if we're dealing with text or rich text content
+    guard isTextContent || isRichText else { return }
+
+    // Remove all non-plain-text content
+    let stringType = NSPasteboard.PasteboardType.string.rawValue
+    item.contents.removeAll { $0.type != stringType }
+
+    // Update or add the plain text content
+    if let index = item.contents.firstIndex(where: { $0.type == stringType }) {
+        if let data = editableValue.data(using: .utf8) {
```
p0deje, line 111: Can you please use 2-spaces indentation everywhere>

Maccy/Settings/PinsSettingsPane.swift
```diff
@@ -27,6 +29,99 @@ struct PinTitleView: View {
...
+    // Consider it text content only if it has plain text and doesn't have images or file URLs
+    self._isTextContent = State(initialValue: hasPlainText && !hasImage && !hasFileURLs)
+    self._isRichText = State(initialValue: hasRichText && !hasImage && !hasFileURLs)
+  }
+
+  var body: some View {
+    Group {
+      if isTextContent || isRichText {
+        ZStack(alignment: .trailing) {
+          TextField("", text: $editableValue)
```
p0deje, line 59: Do you know if there is a way to vertically expand the multiline text field so it can show full text? I could imagine a lot of people would use this as a simple snippet manager and editing multiline text would be necessary.

Maccy/Settings/PinsSettingsPane.swift
```diff
@@ -7,8 +7,10 @@ struct PinPickerView: View {
 
   var body: some View {
     if let pin = item.pin {
+      // Ensure unique pins for ForEach
+      let uniquePins = Array(Set(availablePins + [pin])).sorted()
```
p0deje, line 11: Any particular reason for this change?

Maccy/Settings/en.lproj/PinsSettings.strings
```diff
@@ -1,4 +1,7 @@
 "Title" = "Pins";
 "Key" = "Key";
 "Alias" = "Title";
-"PinCustomizationDescription" = "You can customize the title and hotkey of every pinned item.\nTo edit, double-click it and enter new value.";
\ No newline at end of file
+"Content" = "Content";
+"ContentIsNotText" = "Non-editable content (image or file)";
+"RichTextEditWarning" = "Editing will discard all the formatting";
```
p0deje, line 6:
```suggestion
"RichTextEditWarning" = "Editing will discard all formatting.";
```

Maccy/Settings/en.lproj/PinsSettings.strings
```diff
@@ -1,4 +1,7 @@
...
+"PinCustomizationDescription" = "You can customize the title and hotkey of every pinned item.\nOnly text content can be edited directly. To edit, double-click it and enter new value.";
```
p0deje, line 7:
```suggestion
"PinCustomizationDescription" = "You can customize the hotkey, title, and content of any pinned item. To edit, double-click the item and enter a new value.\nPlease note that only plain text can be changed";
```

Maccy/Settings/PinsSettingsPane.swift
weisJ: I too think it is confusing that the popover appears when hovering and also is toggled through clicking the icon. The icon with `.help(...)` is more idiomatic in my opinion.

weisJ (commented):
I think in addition to the warning icon the user should be requested to explicitly confirm discarding the rich text content with a popup. Possibly with a "don't ask again checkmark" for power users, who don't care about it. (Of course there should be a way to reset it in the settings).

weisJ:
Another thing to consider is, that once editing content of pins landed I can see people asking why they cannot edit any item. Right now they'd have to pin the item, go to the settings. edit it there, and then unpin the item. 

A possibility to streamline this experience, would be to make item editing into a modal dialog, which can be opened
- through the pins setting
- or through a keyboard shortcut/button in the info popover while it is open.

In both cases the option to edit the title of the item would only appear for pinned items. This would go hand in hand with my proposal (#1073) for making the info popover less intrusive.
One challenge here would probably be to prevent Maccy from dismissing itself while the dialog is open.

Opinions on this?

Maccy/Settings/PinsSettingsPane.swift
imlolman: Sure. I have updated the current indentation issues, will keep in mind for future.

imlolman:
> Another thing to consider is, that once editing content of pins landed I can see people asking why they cannot edit any item. Right now they'd have to pin the item, go to the settings. edit it there, and then unpin the item.
> 
> A possibility to streamline this experience, would be to make item editing into a modal dialog, which can be opened
> 
> * through the pins setting
> * or through a keyboard shortcut/button in the info popover while it is open.
> 
> In both cases the option to edit the title of the item would only appear for pinned items. This would go hand in hand with my proposal (#1073) for making the info popover less intrusive. One challenge here would probably be to prevent Maccy from dismissing itself while the dialog is open.
> 
> Opinions on this?

A modal dialog for editing clipboard history could be cool, but honestly, I'm not sure it fits the primary use case for Maccy. I see pins as the items people use repeatedly, so making them easy to manage in the settings pane seems like the most practical approach. We can always explore the ideas in #1073 separately, but I want to keep this PR focused on pins.

p0deje:
First of all, sorry for taking so long to review the PR - I know you've put a lot of effort into it! 

I've been testing this locally and I am on the fence whether the current solution is the most user-friendly. Let's say you pinned a new item. Now you open a pins settings pane and you see something like this:

Screenshot 2025-05-31 at 22 08 16

It's a bit confusing at first because both title and content show the same thing, so you don't even know which one needs to be edited if you want to change the pin. Eventually, you understand that you need to double-click the content cell and change it. Now, what's going to happen to a title? Should it be updated as well? I suppose it should since you had not changed it before. What if you first changed the title and then edited the contents? In this case, I feel like it should remain the same regardless of what you to the contents. 

Another problem is actually editing. There is currently no way to render multiline text which is very limiting, especially if your pin contents are programming code. 

There is also no way to relate image/file contents to pins in the table. This is not related to the PR, but the general issue of the current "Pins" settings approach.

With all that mind, I lean towards completely removing "Pins" section and actually moving all these controls to the "Preview" popup of the pin. So when you inspect the pin with a preview, there would be:

1. pin/unpin button (when pinned, it would also show a pin letter selector)
2. edit button that switches text preview to an editable text field
3. pin letter selector
4. title editable field

Screenshot 2025-05-31 at 22 18 33

We can even allow editing contents for unpinned items too, since there is no reason to restrict it anymore! I feel like this new approach would solve all the mentioned problems and make UX much smoother.

What are your thoughts on this?

imlolman:
Let's revisit the original goal here: the reason I created this PR was that the UI gave me no clue whether editing the title would also modify the content. That confusion was the core issue.

My initial solution was simply to add a column displaying the content, as I've already done. I'm perfectly fine with only making the title editable, as that alone would resolve the initial confusion.

Regarding your suggestion to add buttons to the preview box, I'm hesitant. The preview box currently toggles on hover and contains no editable elements. Introducing a multi-line image and textarea WYSIWYG editor there feels like scope creep that detracts from our primary goal.

I think a simple box displaying the content alongside an editable title (perhaps with an edit icon in the title column) would be a more focused and effective solution.

p0deje:
Good points, I'm going to merge it as-is so that the title/content is editable as of today. It's not going to be the best UX, but I'll improve it with a new preview implementation that will include buttons for pin/delete along with editing capabilities.

p0deje:
Thank you for the conribution!

moutella:
@imlolman Just wanted to say thank you, this feature is really helpful! I hadn't realized before that it was possible to edit the pins previously, until I tried implementing it myself

I only found out about it because I tried implementing this feature myself, and when I cloned the repository I realized that the repository was in a different state than the latest build, and found this PR. It really does improve the UX

@p0deje I looked through the repository but couldn’t find any information about when new releases are made. I’m using a local build for now to take advantage of this feature, but I think end-users would really benefit from having it in an official release

p0deje:
@moutella New releases are made on a case-by-case basis. I think it's been a while since the last one, so I should probably cut a new version with all new features.

p0deje:
I was hoping to get #1123 merged before the release, but it has some bugs and I'm not sure it's worth waiting

Issue #1075, opened 2025-04-03
clipboard item preview overflow not visible


hichemfantar:
### Before Submitting Your Bug Report

- [x] I have verified that there isn't already an issue reporting the same bug to prevent duplication.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Maccy Version (see 'About' window)

2.3.0

### macOS Version

14.3.1 

### Maccy Settings

```Shell
{
    "KeyboardShortcuts_delete" = "{\\"carbonModifiers\\":2048,\\"carbonKeyCode\\":51}";
    "KeyboardShortcuts_pin" = "{\\"carbonKeyCode\\":35,\\"carbonModifiers\\":2048}";
    "KeyboardShortcuts_popup" = "{\\"carbonKeyCode\\":8,\\"carbonModifiers\\":768}";
    "LaunchAtLogin__hasMigrated" = 1;
    "NSStatusItem Visible Item-0" = 1;
    "NSStatusItem Visible Item-1" = 0;
    "NSWindow Frame SUUpdateAlert" = "487 409 620 398 0 0 1512 944 ";
    "NSWindow Frame com.sindresorhus.Settings.FrameAutosaveName" = "542 607 510 232 0 0 1512 944 ";
    SUAutomaticallyUpdate = 0;
    SUEnableAutomaticChecks = 1;
    SUHasLaunchedBefore = 1;
    SULastCheckTime = "2025-01-14 19:03:36 +0000";
    SUSendProfileInfo = 0;
    SUUpdateGroupIdentifier = 1642517698;
    enabledPasteboardTypes =     (
        "public.rtf",
        "public.html",
        "public.file-url",
        "public.utf8-plain-text",
        "public.tiff",
        "public.png"
    );
    historySize = 100;
    ignoredPasteboardTypes =     (
        "com.typeit4me.clipping",
        "Pasteboard generator type",
        "com.agilebits.onepassword",
        "net.antelle.keeweb",
        "de.petermaurer.TransientPasteboardType"
    );
    menuIcon = clipboard;
    migrations =     {
        "2024-07-01-version-2" = 1;
    };
    showFooter = 1;
    showInStatusBar = 1;
    showRecentCopyInMenuBar = 0;
    showSearch = 1;
    showTitle = 1;
    windowSize = "[450,800]";
}
```

### Description

[...]

PR #1110, opened 2025-06-08
Customize preview item character limit and size

harishgonnabattula:
This PR addresses the bug/feature request in https://github.com/p0deje/Maccy/issues/1075.

## Improvement
We set a hard limit on preview item character limit to 10K but this causes some large texts to be cut off. Moving this to setting to allow for customization.

Screenshot 2025-06-08 at 12 01 42 AM

## Bug

We set the preview height to be `min(textSize.height, maxHeight)`  causing the bug observed above. I dont completely understand why we limit it to a minimum. Assigning direct to textSize.height seem to be working properly.
**Will need feedback on if this acceptable.**

### DIFF
--- a/Maccy/Extensions/Defaults.Keys+Names.swift
+++ b/Maccy/Extensions/Defaults.Keys+Names.swift
@@ -43,6 +43,7 @@ extension Defaults.Keys {
   static let popupPosition = Key<PopupPosition>("popupPosition", default: .cursor)
   static let popupScreen = Key<Int>("popupScreen", default: 0)
   static let previewDelay = Key<Int>("previewDelay", default: 1500)
+  static let characterLimit = Key<Int>("characterLimit", default: 10_000)
   static let removeFormattingByDefault = Key<Bool>("removeFormattingByDefault", default: false)
   static let searchMode = Key<Search.Mode>("searchMode", default: .exact)
   static let showFooter = Key<Bool>("showFooter", default: true)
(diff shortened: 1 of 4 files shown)


harishgonnabattula:
@p0deje can you take a look. Thanks!

alexisgaziello:
I would be hesitant to merge this change. It's additional config that makes the interface more complex and I wonder if a significant number of users would ever use the option.

What was the original reason for the character limit? I assume it was for performance. I tested a build after [removing the limit entirely](https://github.com/alexisgaziello/Maccy/tree/feature/no-char-limit) and found that everything worked smoothly, though I acknowledge I have a powerful computer.

-------

Perhaps a better alternative is to offer this as an advanced option that users can modify with a `defaults` command:

```
defaults write org.p0deje.Maccy previewCharacterLimit 20000
```

This approach would provide the flexibility for users who need it without cluttering the UI.

-----

Video with a preview containing >60k chars. 

https://github.com/user-attachments/assets/fafb9b54-e371-4b54-b68a-c2e822fca1bd

p0deje:
> What was the original reason for the character limit? I assume it was for performance. I tested a build after [removing the limit entirely](https://github.com/alexisgaziello/Maccy/tree/feature/no-char-limit) and found that everything worked smoothly, though I acknowledge I have a powerful computer.

Yeah, it was for performance reasons originally. I think there is an opportunity to remove this limit completely (or set to a very high number), but I'm trying to completely rework the preview UI, so I don't focus on improving what's currently implemented.

harishgonnabattula:
The character limit setting is fine to not be included, but what about the bug fix? @p0deje if you are completely changing the preview then I will close this PR otherwise does it make sense to have the bug fix merged?

p0deje:
@harishgonnabattula if you could open a separate PR with a bugfix only, I'd be happy to merge it and relaese in the next version.

weisJ:
I am looking into removing the character limit completely within #1296.

PR #1121, opened 2025-06-22

alexisgaziello:
Rewrite of https://github.com/p0deje/Maccy/pull/1016 using KeyboardShortcut API and `addLocalMonitorForEvents`


alexisgaziello:
Got it, yeah we can do it - opening a separate PR

Issue #1038, opened 2025-01-27
Empty space when app bottom is hidden


ad-angelo:
### Before Submitting Your Bug Report

- [x] I have verified that there isn't already an issue reporting the same bug to prevent duplication.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Maccy Version (see 'About' window)

Since 2.0.1

### macOS Version

15.0.1

### Maccy Settings

```Shell
{
    "KeyboardShortcuts_delete" = "{\\"carbonKeyCode\\":51,\\"carbonModifiers\\":2048}";
    "KeyboardShortcuts_pin" = "{\\"carbonKeyCode\\":35,\\"carbonModifiers\\":2048}";
    "KeyboardShortcuts_popup" = "{\\"carbonKeyCode\\":9,\\"carbonModifiers\\":768}";
    NSNavPanelExpandedSizeForOpenMode = "{800, 448}";
    NSOSPLastRootDirectory = {length = 588, bytes = 0x626f6f6b 4c020000 00000410 30000000 ... 10010000 00000000 };
    "NSStatusItem Visible Item-0" = 0;
    "NSStatusItem Visible Item-1" = 0;
    "NSWindow Frame NSNavPanelAutosaveName" = "331 284 800 448 0 0 1440 875 ";
    "NSWindow Frame SUUpdateAlert" = "401 379 637 398 0 0 1440 875 ";
    "NSWindow Frame com.sindresorhus.Settings.FrameAutosaveName" = "479 279 510 487 0 0 1440 875 ";
    SUAutomaticallyUpdate = 1;
    SUEnableAutomaticChecks = 1;
    SUHasLaunchedBefore = 1;
    SULastCheckTime = "2025-01-27 17:20:47 +0000";
    SUUpdateGroupIdentifier = 3299539765;
    clearOnQuit = 1;
    clearSystemClipboard = 1;
    enabledPasteboardTypes =     (
        "public.html",
        "public.utf8-plain-text",
        "public.rtf",
        "public.tiff",
        "public.png"
    );
    ignoreEvents = 0;
    ignoredApps =     (
        "com.apple.Passwords"
    );
    migrations =     {
        "2024-07-01-version-2" = 1;
    };
    pasteByDefault = 1;
    removeFormattingByDefault = 0;
    searchVisibility = duringSearch;
    showApplicationIcons = 0;
    showFooter = 0;
    showInStatusBar = 0;
[...]

PR #1123, opened 2025-06-23

alexisgaziello:
Rewrite of https://github.com/p0deje/Maccy/pull/1121 using KeyboardShortcut API and addLocalMonitorForEvents, and taking account repeated events.


alexisgaziello:
Im gonna use this for the next week an ensure everything works fine before merging

alexisgaziello:
Notes;
- the alternative still uses onKeyDown, would you rather use onKeyUp?
- I don't remove/restart the localMonitor everytime - it shouldn't be necessary. Only on keyboard shortcut removal.
- No need to use isRepeat, functionality works out of the box

p0deje:
> the alternative still uses onKeyDown, would you rather use onKeyUp?

I would prefer `keyUp` because that's what the user does, but it's not a strong opinion. 

> I don't remove/restart the localMonitor everytime - it shouldn't be necessary. Only on keyboard shortcut removal.

That makes sense.

> No need to use isRepeat, functionality works out of the box

Even better!

p0deje:
There seems to be a bug that when you close the popup by clicking outside, the shortcut is not registered back. You might want to somehow integrate the logic to enable/disable the keyboard shortcut within `FloatingPanel` as it's the only place that handles open/close. Alternatively, you can use `onChange(of: scenePhase)` callback in `ContentView` and handle the shortcut based on the `active`/`background` scene.

alexisgaziello:
> I would prefer keyUp because that's what the user does, but it's not a strong opinion.

What do you mean? I chose `onKeyDown` mainly because it's more responsive and will bring Maccy faster than onKeyUp

> There seems to be a bug that when you close the popup by clicking outside, the shortcut is not registered back.

Good catch. Adding `KeyboardShortcuts.enable(.popup)  // reset` to the close() function of the FloatingPanel class instead of Popup solves the issue. 

Makes me thing, why do we have these 2 separate classes? Should I actually move all this logic to the FloatingPanel class?

p0deje:
`FloatingPanel` is more abstract and doesn't really know anything about the contents of what's being rendered - it's just an `NSPanel` after all. `Popup` is, however, more specific to Maccy and is tied to SwifUI views more by being `Observable`. This allows me to run some tasks when, for example, the popup is being resized.

I'd rather keep `FloatingPanel` as simple as possible to make sure it only bridges SwiftUI to NSPanel. If there is a way to use views and view modifiers to implement the functionality, let's do that. If not, let's write an extension to `FloatingPanel`.

alexisgaziello:
Hey sorry, I haven't had time to spend on this over the last couple of weeks. Ill fix the last little bug this weekend.

p0deje:
@alexisgaziello No worries, take your time!

alexisgaziello:
Ok I think bug is solved hopefully in an elegant-ish manner. Feel free to make any modifications. Essentially added an `onClose` function to floating panel that can be filled with whatever.

Also added an additional test to validate the bug.

alexisgaziello:
One more small commit to cycle to the beginning when reaching the end of footer.

Might have to edit the logic a bit when footer is hidden, but it looks like the feature to hide the footer is kinda broken right now (https://github.com/p0deje/Maccy/issues/1038). We can address cycling to the beginning when footer is hidden when fixing (https://github.com/p0deje/Maccy/issues/1038).

p0deje:
Testing this now and I can see there is a bug where I would release the modifiers, then press Shift+C and it confirms selection.

p0deje:
> One more small commit to cycle to the beginning when reaching the end of footer.

It used to work like that, but I intentionally removed this behavior in Maccy 2.0 because that's not how macOS menus work. When you reach the end of the menu, pressing the down arrow doesn't cycle back to the top.

> Testing this now and I can see there is a bug where I would release the modifiers, then press Shift+C and it confirms selection.

I think I have it working, mind if I push a commit to your PR?

alexisgaziello:
oops didnt see your had a fix and pushed a fix, let me know your thoughts.

Do you want me to remove the "cycle back to the beginning then?" Its mainly because during my usage I sometimes go to the end and the last option is Quit. Getting stuck there is not ideal as you can Quit Maccy without really wanting and avoiding this require some mouse movement.

alexisgaziello:
Also you are a great tester lol. I have been using this feature for weeks and never noticed the bug. I guess I use Maccy differently

p0deje:
> oops didnt see your had a fix and pushed a fix, let me know your thoughts.

I haven't pushed anything yet, no worries.

> Do you want me to remove the "cycle back to the beginning then?" Its mainly because during my usage I sometimes go to the end and the last option is Quit. Getting stuck there is not ideal as you can Quit Maccy without really wanting and avoiding this require some mouse movement.

Ugh, that's unfortunate. Can you please change `highlightNext` to have an extra arg `allowCycle` that is `false` by default? You can then pass `true` to it from cycle mode only, where it makes total sense.

> Also you are a great tester lol. 

Lol I've been doing QA for 18 years professionally.

alexisgaziello:
All right let's do it that way

alexisgaziello:
Excited to close the oldest Maccy issue!! https://github.com/p0deje/Maccy/issues/96

p0deje:
> oops didnt see your had a fix and pushed a fix, let me know your thoughts.

hmm, I can still reproduce, let me reapply my patch and push

p0deje:
@alexisgaziello Mind pulling and taking a quick test? This seems to work fine in my limited scenarios.

alexisgaziello:
mhmhm weird Im pretty sure I fixed it. Maybe I am not understanding the issue?

In any case just tested your branch and looks good to me!

Note; you removed the other commit for allowing cycle to the first item. Might want to add that back?

p0deje:
Poor merge, sorry. Returned it.

alexisgaziello:
Also I noticed you removed

    if state == .opening **&& allModifiersReleased(event) {**

I prefer it as it allows me to keep one modifier pressed and start cycling (I usually just keep cmd as pinky gets tired so release it from shift).

This is my personal preference though, we can decide whatever behavior you think makes more sense

p0deje:
> I prefer it as it allows me to keep one modifier pressed and start cycling (I usually just keep cmd as pinky gets tired so release it from shift).

Right, but the downside of this is that you cannot release CMD and press C to enter uppercase C letter. The implementation is a bit blurry, but now you can:

1. Keep all modifiers pressed.
2. Press C to cycle to the next item.
3. Release SHIFT.
4. Press C to cycle through other items.
5. Release CMD to confirm.

Is this an acceptable workaround for you? You can still release SHIFT, just not on the first press.

alexisgaziello:
Yeah your workarounds works.

But the way I use it is if I want to type something, I enter toggle mode which for me is just releasing everything. The I start typing whatever I need to type, which might be upper case C or whatever I want to type.

Keeping the shift from the shortcut pressed because you want to type an uppercase C seems like a rare scenario IMO, specially given that the search ignores casing.

Something I do a lot is open Maccy, inspect which position I am targeting, and then send a couple of Cs to get there. While I stay in the initial opening  for a couple of seconds searching my target, I usually release the shift.

I would suggest trying both during the next couple of days and you can decide which one you want to implement based of which option you think is more comfortable.

p0deje:
I reverted the change back, you are the first real user of this feature so I'll trust your judgement.

p0deje:
Anything else you want to add/change here or shall I merge and prepare for a new release next week?

alexisgaziello:
Looks good to me!!

Issue #1126, opened 2025-07-11
Can we have a privacy toggle so that the Maccy popup window doesnt show up at all on screen recordings and sharings


adamlahbib:
### Before Submitting Your Feature Request

- [x] Check that there isn't already a similar feature request to avoid creating a duplicate.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

freebrowser1:
Can be a nicxe idea, but how do blind you select a clipping ?

weisJ:
It is possible to hide an application window from screen recordings by setting the windows [sharingType](https://developer.apple.com/documentation/appkit/nswindow/sharingtype-swift.property) to `.none`. You will still be able to see the window yourself.

adamlahbib:
@p0deje can you check this PR https://github.com/p0deje/Maccy/pull/1130

p0deje:
merged in 789d2814f7b78b9be28e87b4caaed737e7cc1fcf

p0deje:
I think we need to iterate on this for a bit. 

1. We need to get preview hidden too (#1133 should address this).
2. This now prevents screenshots from capturing Maccy window complete. I am afraid it's going to complicate a lot of bug reports :) I think there should be a way to determine when screen sharing is starting in macOS (https://www.reddit.com/r/swift/comments/1jysl8g/how_does_raycast_detect_that_screen_sharing_is/). When that happens, we can automatically enable private mode and turn it off when sharing stops.

Any thoughts?

moutella:
About 2. Unfortunately the method posted on reddit does not work, since UIKit is not compatible with macOS. Someone sent an link at the end of the thread that may have some other method, but it's currently down. 

I found another app that claims to be able to detect if your screen is being shared, called "Stealthly", but it explicitly states that the Mac App Store version of the app doesn't have this feature, so it's definitely something "not intended" by Apple, which means it's probably not easy to achieve such feature. Raycast also isn't on the Mac App Store.

With that in mind, I think a partial workaround would be:

1. Add a hidden setting that decides whether or not to change the sharingType
2. Set the default to hide Maccy. 
3. Then, add in the issue template the commands to change the settings, to help people with bug reports
    ```sh
    defaults write org.p0deje.Maccy privacyMode false # default is true
    ```

This may be a bit complicated, but I believe most people that are sending bug reports would be able to do this.

[...]

PR #1130, opened 2025-07-15
Set window sharing type to .none to block screen capture

adamlahbib:
## What

This PR sets the sharingType of the FloatingPanel window to .none, effectively preventing the window from being captured in screen recordings or screenshots by other apps.

## Why

Privacy & Security: This ensures that sensitive UI content shown in the floating panel cannot be recorded by screen capture tools.

User trust: Helps reinforce that the app respects user privacy and does not expose clipboard contents or UI overlays unintentionally.

macOS Support: The NSWindow.sharingType property is fully supported on macOS and allows a per-window decision on whether it can be shared.

## Resources
Feature Request: https://github.com/p0deje/Maccy/issues/1126
Apple Developer Docs: https://developer.apple.com/documentation/appkit/nswindow/sharingtype-swift.property

### DIFF
--- a/Maccy/FloatingPanel.swift
+++ b/Maccy/FloatingPanel.swift
@@ -27,6 +27,7 @@ class FloatingPanel<Content: View>: NSPanel, NSWindowDelegate {
 
     self.statusBarButton = statusBarButton
     self.identifier = NSUserInterfaceItemIdentifier(identifier)
+    self.sharingType = .none
 
     Defaults[.windowSize] = contentRect.size
     delegate = self


p0deje:
Looking at the docs for https://developer.apple.com/documentation/appkit/nswindow/sharingtype-swift.enum/none

> NSWindowSharingNone can cause content to not be available in certain sharing situations. Don’t use this value to hide or omit content from being captured. Instead, use FairPlay Streaming (FPS). For more information, read FairPlay Streaming.

p0deje:
I am not sure if FairPlay is the solution too, but does `none` work on recent macOS? The docs say:

> A legacy constant that macOS no longer uses.

weisJ:
I think the value still works and the comment refers to the fact that macOS itself will not create any window with .none sharingType. The FairPlay api seams to specifically be for media playback.

I once tried to mask the content (using https://developer.apple.com/documentation/SwiftUI/View/redacted(reason:)) but failed at reliably detecting when the screen is recorded.

p0deje:
Tested it in Google Meet, and indeed, the window is not present on the shared screen. Thanks!

adamlahbib:
@p0deje The sub popup with the value when you hover is still showing, ill adjust that... Which component is responsible for it?

PR #1133, opened 2025-07-20
Fixes unexpected behavior of previews still showing up while sharing the screen

moutella:
After using #1130 locally, I realized that the preview pop-up are still showing when sharing the screen. This could lead to the user unkowingly sharing private info with others.

Since we can't add ".sharingType" to the preview, I fixed it by disabling the preview when the user chooses to have privacy mode. Tested it in google meet with myself and it's working.

Since this disables a feature when activated, I made the decision to make the default value to false, so by default there would be no privacy.

### DIFF
--- a/Maccy/Views/ContentView.swift
+++ b/Maccy/Views/ContentView.swift
@@ -66,6 +66,8 @@ struct ContentView: View {
         popover.animates = false
         // Prevent NSPopover from becoming first responder.
         popover.behavior = .semitransient
+        // Prevent popover from showing while sharing the screen
+        popover.contentViewController?.view.window?.sharingType = .none
       }
     }
   }


p0deje:
Have you tried adding a sharing type to the NSPopover window in ContentView.swift, something like this:

```swift
popover.contentViewController?.view.window?.sharingType = .none
```

moutella:
@p0deje Thanks for the suggestion, tried it and it works perfectly now!

This was my first time contributing to something made with Swift, still learning a lot 😅

p0deje:
Do you think we can merge this without an extra setting? I think most people would prefer to hide their Maccy window during screen sharing, I'd rather wait for anyone to request such a configuration option. Any new UI changes requires a lot of work with localization, which is why I'm hesitant to add it.

moutella:
I believe the priority should be removing the confusing behavior Maccy has currently, so I definitely agree we should merge without the extra setting. 

However, I am personally a fan of having control over these kind of settings, for special occasions like either teaching someone about Maccy, tutorials about certain features or workflow videos. So maybe this could be considered in another feature request, along with some other options about privacy. Will probably open a feature request after completing this explaining what I am planning to do

With that in mind, I have decided to remove the settings from this PR, and renamed the PR to avoid confusion.

PR #1137, opened 2025-07-22
Fixes screenshot behavior when not sharing screen

moutella:
Solves the issues mentioned by @p0deje [here](https://github.com/p0deje/Maccy/issues/1126#issuecomment-3098020892) 

This means that Maccy will only hide it's windows while the screen is being shared.


p0deje:
Looks like we cannot use sharingType because it effectively makes Maccy windows invisible on external monitors connected via DisplayPort. I've had to revert [96d0dd6](https://github.com/p0deje/Maccy/commit/96d0dd61b095d3f10638d68f6237ea6b27f7c742).

Issue #1136, opened 2025-07-21
Maccy won't open in any monitor other than the Main display


bscheffer13:
### Before Submitting Your Bug Report

- [x] I have verified that there isn't already an issue reporting the same bug to prevent duplication.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Maccy Version (see 'About' window)

2.4.0

### macOS Version

15.5

### Maccy Settings

```Shell
{
    "KeyboardShortcuts_delete" = "{\\"carbonModifiers\\":2048,\\"carbonKeyCode\\":51}";
    "KeyboardShortcuts_pin" = "{\\"carbonModifiers\\":2048,\\"carbonKeyCode\\":35}";
    "KeyboardShortcuts_popup" = "{\\"carbonKeyCode\\":8,\\"carbonModifiers\\":768}";
    "LaunchAtLogin__hasMigrated" = 1;
    "NSStatusItem Visible Item-1" = 0;
    "NSWindow Frame SUStatusFrame" = "1520 975 400 134 0 0 3440 1415 ";
    "NSWindow Frame SUUpdateAlert" = "1410 777 620 398 0 0 3440 1415 ";
    "NSWindow Frame com.sindresorhus.Preferences.FrameAutosaveName" = "1445 659 542 435 0 0 3440 1415 ";
    "NSWindow Frame com.sindresorhus.Settings.FrameAutosaveName" = "1365 886 482 305 0 0 3440 1415 ";
    SUEnableAutomaticChecks = 1;
    SUHasLaunchedBefore = 1;
    SULastCheckTime = "2025-07-21 21:05:59 +0000";
    SUUpdateGroupIdentifier = 759539049;
    WebKitDefaultFontSize = 13;
    WebKitJavaScriptEnabled = 0;
    WebKitPluginsEnabled = 0;
    WebKitStandardFont = "-apple-system-font";
    WebKitUserStyleSheetEnabledPreferenceKey = 1;
    WebKitUserStyleSheetLocationPreferenceKey = "/Applications/Maccy.app/Contents/Frameworks/Sparkle.framework/Resources/ReleaseNotesColorStyle.css";
    enabledPasteboardTypes =     (
        "public.utf8-plain-text",
        "public.tiff",
        "public.rtf",
        "public.png",
        "public.file-url",
        "public.html"
    );
    ignoredPasteboardTypes =     (
        "com.typeit4me.clipping",
[...]

PR #1149, opened 2025-07-26
Adds privacy mode setting with fixes for DisplayLink issues

moutella:
Re-Adds privacy mode(#1126) with fixes for DisplayLink issues #1136


Maccy/AppDelegate.swift
```diff
@@ -184,4 +185,10 @@ class AppDelegate: NSObject, NSApplicationDelegate {
       }
     }
   }
+
+  private func updatePrivacy() {
+    let isDisplayLinkRunning = NSRunningApplication.isDisplayLinkRunning()
+    Defaults[.privacyMode] =  !isDisplayLinkRunning
+    Defaults[.displayLinkDetected] = isDisplayLinkRunning
```
weisJ: We shouldn't overwrite user settings like that. Just don't respect the setting if display link is detected.
Also .displayLinkDetected is transient state and shouldn't be stored in the defaults.

Maccy/Views/ContentView.swift
```diff
@@ -66,6 +67,12 @@ struct ContentView: View {
         popover.animates = false
         // Prevent NSPopover from becoming first responder.
         popover.behavior = .semitransient
+        // Prevents NSPopOver from being captured by screen shots and screen sharings.
+        if Defaults[.privacyMode] {
+          popover.contentViewController?.view.window?.sharingType = .none
+        } else {
+          popover.contentViewController?.view.window?.sharingType = .readOnly
+        }
```
weisJ: If you rebase on master this is no longer necessary

Maccy/FloatingPanel.swift
```diff
@@ -73,6 +73,11 @@ class FloatingPanel<Content: View>: NSPanel, NSWindowDelegate {
     setFrameOrigin(popupPosition.origin(size: frame.size, statusBarButton: statusBarButton))
     orderFrontRegardless()
     makeKey()
+    if Defaults[.privacyMode] {
+      self.sharingType = .none
+    } else {
+      self.sharingType = .readOnly
+    }
```
weisJ: This should be extracted into its own function. We need to call it every time the settings change. See `AppDelegate` for how to listen for changes in the settings.

Maccy/Extensions/NSRunningApplication+DisplayLink.swift
```diff
@@ -0,0 +1,13 @@
+import AppKit.NSRunningApplication
+
+// Originally on Lunar.
```
weisJ:
```suggestion
// Originally from Lunar.
```

weisJ (changes requested):
We should check more often if display link is enabled. Let's check if before the Maccy window is opened and update the sharing mode then accordingly.

moutella:
@weisJ Updated with requested changes. Also changed the tips about privacy mode and displaylink in the settings panel

Maccy/FloatingPanel.swift
```diff
@@ -188,6 +194,14 @@ class FloatingPanel<Content: View>: NSPanel, NSWindowDelegate {
   func windowDidResignKey(_ notification: Notification) {
     AppState.shared.preview.disableAutoOpen()
   }
+  
+  func updateSharingType(isPrivacyModeEnabled: Bool) {
```
weisJ, line 200: Let's call this also when the Maccy window is opened. This way we always respect DisplayLink independent of whether it is started before or after the settings changed.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -14,6 +14,8 @@ body:
       options:
         - label: I have verified that there isn't already an issue reporting the same bug to prevent duplication.
           required: false
+        - label: I have disabled hiding from screen share so that Maccy can be seen in screenshots.
+          required: false
```
weisJ: Might be good to write the path inside the settings here, so people know where to look for this setting.

weisJ (changes requested):

moutella:
Updated with requested changes, also decided to improve the tooltip on preferences to better reflect Maccy's behavior when DisplayLink is running.

Maccy/Settings/en.lproj/AdvancedSettings.strings
```diff
@@ -8,3 +8,6 @@
 "ClearHistoryOnQuitTooltip" = "Automatically remove all unpinned items before quitting the application.";
 "ClearSystemClipboard" = "Clear the system clipboard too";
 "ClearSystemClipboardTooltip" = "When enabled, clearing the history would also clear the current system clipboard.";
+"PrivacyMode" = "Hide from screen capture";
+"PrivacyExplain" = "Hides application windows for screenshots and screen sharing.\nHowever, the system's built-in screen recorder is not affected.";
+"PrivacyDisplayLinkExplain" = "This setting will be ignored if DisplayLink user agent is running.";
```
weisJ:
```suggestion
"PrivacyDisplayLinkExplain" = "This setting will be ignored if the DisplayLink user agent is running.";
```

weisJ (commented):
Looking good now. It would be great if you could also provide the necessary localisation strings for the other languages. Usually AI translation tools do a pretty good job here and the first pass doesn't need to be perfect.

moutella:
@weisJ Updated with AI translations, thanks to @matheus-elevential

Issue #1063, opened 2025-03-13
Add OCR text content pasting function


leerenbo:
### Before Submitting Your Feature Request

- [x] Check that there isn't already a similar feature request to avoid creating a duplicate.
- [ ] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Problem

When you need the text in the image, or the text that cannot be selected.
It can be processed by taking screenshots or copying images and pasting them after OCR(Optical Character Recognition), which is a very useful function.

### Solution

You can assign a shortcut key, click for OCR and paste, just like paste without formatting.
It would be very helpful ,thanks.

leerenbo:
On this basis, add a paste shortcut key.

https://github.com/p0deje/Maccy/issues/992#issuecomment-2547430612

p0deje:
What do you think if we just assign "Paste without formatting" to paste text from image instead of image itself?

avichou:
I figure it shouldn't be too hard now that we have OCR in maccy already, and that'd be very welcome indeed. i'm tired of having to re-run OCR with third party apps for a screenshot that's already in my clipboard

avichou:
this hasn't been implemented yet has it ?

cjohnsonuk:
I use Shottr for screen capture and it includes OCR with text to clipboard https://shottr.cc/#section-tips:~:text=Text%20recognition%20(OCR)%20%26%20QR. Does this work for you?

robert-corlett:
> I use Shottr for screen capture and it includes OCR with text to clipboard https://shottr.cc/#section-tips:~:text=Text%20recognition%20(OCR)%20%26%20QR. Does this work for you?

Github issues aren't an appropriate place to surreptitiously market your apps

PR #1152, opened 2025-07-27
Add extract text from image feature to clipboard preview

DanNoby:
This feature aims to allows users to extract text from images in their clipboard history using Apple’s Vision framework.

When previewing an image in the clipboard history, an “Extract Text” button appears.
Clicking the button runs OCR on the image and copies the extracted text directly to the clipboard.
The feature is only visible for image items.
No popups or alerts are shown; feedback is via clipboard update.

ScreenRecording2025-07-27at11 55 15PM-ezgif com-video-to-gif-converter

Technical notes:
Uses Apple’s Vision framework for OCR (macOS 10.15+).
UI/UX is minimal and consistent with the Maccy's design.
Tested on:
macOS Sequoia 15.5, Xcode 16.43

### DIFF
--- a/Maccy.xcodeproj/project.pbxproj
+++ b/Maccy.xcodeproj/project.pbxproj
@@ -837,7 +837,7 @@
 			attributes = {
 				BuildIndependentTargetsInParallel = YES;
 				LastSwiftUpdateCheck = 0920;
-				LastUpgradeCheck = 1520;
+				LastUpgradeCheck = 1640;
 				ORGANIZATIONNAME = p0deje;
 				TargetAttributes = {
 					DA0EE7B5204657830025FC60 = {
@@ -893,6 +893,7 @@
 				sl,
 				be,
 				uz,
+				Base,
 			);
 			mainGroup = DAEE383A1E3DBEB100DD2966;
 			packageReferences = (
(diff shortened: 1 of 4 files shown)


p0deje:
This was requested part of https://github.com/p0deje/Maccy/issues/1063 and I think we should implement it as "Paste without formatting" shortcut. There is already logic to extract text from image and store it inside "title" field (see https://github.com/p0deje/Maccy/commit/2255498df150aeccd3341b89cad6596032db7c13). With this in-place, we only need to add a special handling of paste without formatting to copy the title of `HistoryItem` instead of its contents.

DanNoby:
Got it, thanks...

Just to confirm — shall I go ahead and work on the “Paste without formatting” shortcut and update this PR to follow the approach you described (i.e., copying from the title field of the HistoryItem)?

Happy to modify this PR accordingly if that works for you.

p0deje:
@DanNoby It would be great. This should be as simple as probably adding a condition to check for image + removeFormatting somewhere in https://github.com/p0deje/Maccy/blob/245986ba854058d6bcbc174aacecd5b86ee19b6a/Maccy/Clipboard.swift#L73-L109

DanNoby:
I’ve updated the PR to implement the “Paste without formatting” feature as discussed. 
When user presses OPTION + SHIFT + ENTER on a picture they've taken, it pastes the text now.

I ran into a few SwiftLint and build issues along the way — sorry for the extra commits while I worked through them 😅

avichou:
I wouldn't be against having both :

a shortcut (what @p0deje proposes) , as well as an option in the form of apple OCR in the bottom right after performing an image preview . generally, good software have both, as visual indicator helps paliate a potential memory issue

avichou:
> @DanNoby It would be great. This should be as simple as probably adding a condition to check for image + removeFormatting somewhere in
> 
> https://github.com/p0deje/Maccy/blob/245986ba854058d6bcbc174aacecd5b86ee19b6a/Maccy/Clipboard.swift#L73-L109

hi, has this been implemented yet ? i can't really tell because i have issues with paste without formatting anyway

jameswomack (changes requested):

Issue #1175, opened 2025-08-07
More Sorting options / paste consumption behaviour


merlondev:
### Before Submitting Your Feature Request

- [x] Check that there isn't already a similar feature request to avoid creating a duplicate.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Problem

I would REALLY like, if there were more sorting options / behaviours for what happens when pasting something..

The main reason I want to use Maccy is for transferring text from a discrete set of input fields to another, so I'll frequently have to copy item1, switch app, paste item1, switch back, copy item2....

I could imagine many users benefiting from this option, that are currently just copying everything and then pasting it via selection through cmd+number or arrow keys + enter, which is very suboptimal for this use case, once the amount of things to transfer get bigger than 6

Also the selection via cmd+number combined with the Time of first copy sorting option would only fix this until 9 items...

### Solution

one sorting mechanism I'm thus thinking about:
- remove (or at least put to bottom of the list) an item upon pasting it
    - in case of removal of the copied item, that's it!
    - in case of the no-removal option, I suppose you could call the sorting criterion for this "Time of last paste (most recent last)", i.e. a list that looked like this (from top to bottom, all just freshly copied): a, b, c, d, e would look like this after (successfully?) hitting paste twice: c, d, e, a, b

Honestly, one might want to revamp the "Sort by:" Storage settings into two parts entirely actually:
1. Action on copy item (via the native cmd+c or similar)
2. Action on paste item (via Maccy window or also via the native cmd+v or similar)
which would then have options like:
- for 1.:
  - always put item at top of list (equivalent to "Time of last copy" sorting)
  - put only NEW items at top of list (equivalent to "Time of first copy" sorting)
  - sort into list by times copied (equivalent to "Number of copies" sorting)
  - etc
- for 2.:
  - do nothing (equivalent to current default behaviour, I think)
  - remove item
  - put item at bottom of list
  - etc

I think this is probably the best way, since it doesn't impact previous users, due to the equivalence/compatibility with the previous options,
but still allowing more fine grained control for those who need it.

merlondev:
I've just noticed that this would also make this https://github.com/p0deje/Maccy/issues/239 much easier.
While you wouldn't have the "paste in one go" option here, you could just copy the items one by one and then paste them quickly one by one, alternating between "cmd+v" and "enter" (producing a newline, or whatever you want to write in between),
[...]

PR #1178, opened 2025-08-08
Add option to push pasted item to bottom of the list

merlondev:
Add the option to move the item you just pasted (via the Maccy ui, while the "paste automatically" option is enabled, either by pressing it or hitting enter with it is selected)

This is implemented by adding an extra sort criterion behind a boolean flag (in Defaults)

This pull request is still a WIP, just to show a minimal sample implementation of this issue: https://github.com/p0deje/Maccy/issues/1175

Things that are still missing:
- I'm not sure yet!

(please excuse the slight refactor in `bySortingAlgorithm`, it is not required for this pull request)

### DIFF
--- a/Maccy/Extensions/Defaults.Keys+Names.swift
+++ b/Maccy/Extensions/Defaults.Keys+Names.swift
@@ -54,6 +54,7 @@ extension Defaults.Keys {
   static let showTitle = Key<Bool>("showTitle", default: true)
   static let size = Key<Int>("historySize", default: 200)
   static let sortBy = Key<Sorter.By>("sortBy", default: .lastCopiedAt)
+  static let pushPastedToBottom = Key<Bool>("pushPastedToBottom", default: false)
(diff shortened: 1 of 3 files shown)


Maccy/Clipboard.swift
```diff
@@ -109,9 +109,13 @@ class Clipboard {
   }
 
   // Based on https://github.com/Clipy/Clipy/blob/develop/Clipy/Sources/Services/PasteService.swift.
-  func paste() {
+  func paste(_ item: HistoryItem? = nil) {
     Accessibility.check()
 
+    if let item = item {
+      item.lastPastedAt = Date.now
+    }
```
merlondev, line 117: unfortunately, this doesn't seem to change the `.lastPastedAt` property of the Actual HistoryItem.
Maybe because it is passed by value, instead of by reference?

Honestly, I don't really get how `.lastCopiedAt` is updated either (on copy).
@p0deje If you could point me towards that location, it would be very much appreciated!

Issue #239, opened 2021-02-11
Feature request: Paste multiple items


msubel:
First of all, thanks for creating Maccy and all the work that has been put in so far.

My request is the following. I'd like to be able to copy many individual items to the clipboard history, invoke Maccy, select multiple items and paste them. I imagine all the items would be pasted in one go, joined by `\n` or something similar.

This is handy in situations, in which the source is spread over multiple places, e.g. distinct address fields (Street, Zip, City, etc.) that need to pasted as one element later.

p0deje:
I will consider merging PR which adds support for this (e.g. select items by pressing Shift+arrow and then they all are pasted), but I don't have time to design the UX for this and implement it right now.

Deus-SiveNatura:
I came here to find out if this is possible as well. For my use case, I would just like to able to dump the whole clipboard history at once (something like Shift+A to select everything and then Cmd+V to paste?, or whatever other way would work best). I don't think any other clipboard manager has this function, or at least none that I can find, so I was wondering if there was some technical reason for that.

p0deje:
Why would you want to copy and save all clipboard history? This looks shady to me as it allows anyone with access to computer quickly save all the history.

Deus-SiveNatura:
It's just a quicker way to do what the original poster in this thread suggested, in the case that you need all of the items from your session, and not just the last 5-6. 

In my case it would be to save words from a book or a document that I'm reading in a separate document. At the moment, if I need to save 50 words—I need to perform 50 paste operations. I would prefer just copying as I read and then pasting everything at once when I'm done, but admittedly I hadn't thought of the security concerns. 

The original suggestion, and your response, would be a big enough improvement for me anyway (Shift+arrowing the last 50 from maccy's history, then pasting them with one operation, if I understood it correctly), so if you think that that is less shady, I would be happy to be able to do it that way.

sfc-gh-afedorov:
I've heard this called "sequential paste" and it's very useful. Something like ctrl+cmd+c then cmd-c to push a bunch of things onto the sequential paste queue, then ctrl-cmd-v to pop them off the paste queue (hitting return in between, if you'd like).

p0deje:
@sfc-gh-afedorov Interesting, can you share any links or examples where this "sequential paste" is implemented?

p0deje:
[...]

PR #1206, opened 2025-09-15
Implement paste stack

weisJ:
This is an implementation of the paste stack proposed in https://github.com/p0deje/Maccy/issues/1175#issuecomment-3179700597.

The basis for this is an added multiple selection functionality. To select multiple items one can either use one of the following keyboard shortcuts
- `SHIFT + DownArrow`/`CTRL + SHIFT + n`: Extend the selection to the next item
- `SHIFT + CMD/Option + DownArrow`/`CONTR + OPTION + SHIFT + n`: Extend the selection to the last item
- `SHIFT + UpArrow`/`CTRL + SHIFT + p`: Extend the selection to the previous item
- `SHIFT + CMD/Option + UpArrow`/`CONTR + OPTION + SHIFT + p`: Extend the selection to the first item 

or `CMD + Click` with the mouse to select additional items.

The order they are selected in is displayed next to the item content.
Once more than one item is selected moving the mouse over other items no longer auto selects the item as this would make mouse selection impossible and make accidentally clearing the whole selection to easy.

Pinning and deleting multiple items works as expected.

One can then use any of the existing shortcuts to create a **paste stack**.
This immediately copies the first selected item to the clipboard ready to paste. Once the item has been pasted it is removed from the top of the paste stack and the next item is automatically copied to the clipboard. The paste stack remembers the shortcut used to create the stack i.e. if the "copy without formatting" shortcut was used then all subsequent copies will also be copied without formatting.

Note that the "paste automatically" option does not work with paste stacks. Pasting always has to be initiated by the user.
I did not manage to listen for `CMD + V` using the carbon api without the shortcut being consumed (with the result that pasting becomes impossible). This feature therefore requires accessibility permissions as it globally listens for `CMD+V` key presses. 

The current paste stack is displayed at the top of the history list. The number displays the current number of items in the stack.

SCR-20250915-jwmx

Here is a video of the behaviour in action. The visuals are slightly outdated but the user interaction hasn't changed since:

https://github.com/user-attachments/assets/e2d0478f-f0b5-4150-9cec-1ba90f4f4f45

It can be selected like any other entry. To discard the current paste stack simply delete it using `OPTION + Backspace`.

If any other copy happens while the paste stack is active (either through Maccy or otherwise) the stack is discarded. Feedback on this would be welcome as I had no better idea what the behaviour should be in this case.

The paste stack does not reorder any items in the history list on its own. The items get moved to the start of the list just as if they were copied regularly. After pasting all items of the stack this effectively puts the items of the paste stack at the start of the history list in reverse oder (Assuming sorting by time of last copy).

I have tried to group the changes into logical steps. Every commit should compile on its own, but there may be things included in some changes, which are only used in a later commit.

Relates to #1175 #239

### DIFF
--- a/Maccy.xcodeproj/project.pbxproj
+++ b/Maccy.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		0ABDD5122BB47F1E0054963B /* NSWorkspace+ApplicationName.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0ABDD5112BB47F1E0054963B /* NSWorkspace+ApplicationName.swift */; };
+		2F0EC7792E7727E3003E2EA9 /* HeightReaderModifier.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2F0EC7782E7727E3003E2EA9 /* HeightReaderModifier.swift */; };
 		2F12271E2E5C932200A1592A /* Logging in Frameworks */ = {isa = PBXBuildFile; productRef = 2F12271D2E5C932200A1592A /* Logging */; };
 		2F1A79C02C6DFB7800C98EBD /* SearchVisibility.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2F1A79BF2C6DFB7800C98EBD /* SearchVisibility.swift */; };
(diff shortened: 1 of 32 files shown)


weisJ:
Is the bitwise CI working properly? The details page gives me an internal server error.

Issue #1125, opened 2025-07-01
Canned Responses


Eric2XU:
### Before Submitting Your Feature Request

- [x] Check that there isn't already a similar feature request to avoid creating a duplicate.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Problem

Summary
Add an optional “Canned Responses” section to Maccy—a dedicated place where users can save, organize, and quickly paste frequently used snippets or boilerplate text via a separate keyboard shortcut.

Motivation
I use Maccy over 1,000 times per day—it’s exactly what it needs to be for general clipboard history. In my line of work, however, I often retype the same commands, email templates, and code snippets repeatedly. A built-in way to store and access these “canned responses” would save time and reduce typing errors.

Proposed Solution
New Shortcut / Menu

Introduce a configurable global hotkey (e.g. ⌘⇧C) that opens the “Canned Responses” popover.

Alternatively, nest canned responses as a sub-category in the existing Maccy menu.

Organization

Allow users to create folders or tags for grouping related responses (e.g., “Email,” “Git Commands,” “API Keys”).

Display folder names in the popover sidebar for quick navigation.

Adding Responses

Context-menu action on any clipboard entry: “Save as Canned Response.”

Optionally allow manual entry in the preferences panel for snippets not currently in history.

Removing / Editing

Right-click on a canned response in the popover to rename, move folders, or delete.

Provide an “Edit” mode in the preferences UI for bulk management.

Persistence

Store canned responses in a JSON or SQLite file alongside existing Maccy configuration.

Ensure compatibility with iCloud Drive sync (if already supported by Maccy).

Alternatives Considered
Relying on external text-expander tools (adds overhead and breaks Maccy’s unified UX).

Storing favorite snippets in the main clipboard history (hard to distinguish from ephemeral items).

I believe this enhancement keeps Maccy lightweight while unlocking a powerful productivity boost for power users. Happy to discuss any details or refine the design!

twknab:
My only thoughts on this is I wonder if this now loses the original purpose of Maccy? Since the original intent was to create a lightweight and simple tool. This would turn Maccy into another sort of product. Are there any dedicated canned response tools that fit bill, than push that functionality into Maccy? Just a thought.

Eric2XU:
[...]

Issue #307, opened 2021-08-22
Pinned items take too much UI space


borekb:
I like to keep the number of items displayed at around 10 to keep the Maccy popup small and tidy. However, pinned items (if I'm not missing anything) cannot be limited or hidden from the main UI, which is a problem if I have many of them.

Also, while pinned items is a very useful feature of Maccy, I interact with it much less often than with the main history so it doesn't feel right that it should occupy the same space (or even more) in the UI.

Some ideas on how to solve this:

- Something like the mockup in https://github.com/p0deje/Maccy/issues/89#issuecomment-602265753 (a single "Pinned items" item in a list, opening a separate UI)
- Different shortcut and a dedicated UI to open pins, for example, like Clipy: <img width="641" alt="Screen Shot 2021-08-21 at 8 15 36" src="https://user-images.githubusercontent.com/101152/130312550-809855df-13fe-43f2-af58-aae0b1ec1e8b.png">

I'm sure there are other solutions, this is just to give an idea.

p0deje:
Sorry for the slow response, I was away from Maccy for a while.

Since you don't interact with pinned items a lot, would it be useful for you to simply show them at the bottom so that the popup can be big but the pinned items won't take your focus?

borekb:
I already have them at the bottom and I don't think it's enough.

Just to make sure, I'm cross-posting the mockup from [this comment](https://github.com/p0deje/Maccy/issues/89#issuecomment-602265753) which I think could be one of the solutions:

![image](https://user-images.githubusercontent.com/1752380/77259540-6d171900-6c82-11ea-980e-a1f76d7fcedc.png)

greenwar:
tangent to this, i didn't find out to pin items from within the app itself. had to come to github via home site. it'd be great to add a blurb in the pins view in settings that tells the user how to pin. i could pull and update, but i wasn't sure how open this project is. cheers, love the app.

WladyX:
I'm waiting for this feature to make the switch from clipy, i use lots of pinned items, so folders for pinned items would be perfect.
Thank you for this project.

maliayas:
+1 for folders for pinned items (possibly hierarchical?). Only the top folder would be visible in the immediate popup UI. Going inside it would reveal the full tree of pins. The tree hierarchy editable via settings.

douglasbeagley:
I would like to be able to create my own sub-menus with a set of saved pinned items.

[...]

PR #1221, opened 2025-09-25
feat: Add tabs and improve window sizing

hyperwd:
Summary of Changes

  This pull request introduces a new "Tabs" feature for better organization of clipboard history and pinned items, along with significant improvements to window size management.

  New Features

   * Tab-based Navigation:
       * A new optional "Tabs" feature can be enabled in the "Behavior" section of the settings.
       * When enabled, the main window is separated into two tabs: "History" and "Pinned".
       * The "History" tab displays only the copied items (excluding pinned items).
       * The "Pinned" tab exclusively shows pinned items.
       * Users can easily switch between tabs by pressing the Tab key.
       * This feature is disabled by default to maintain the classic user experience.

  Improvements

   * Window Size Persistence:
       * The application now remembers any manually adjusted window dimensions (both height and width) and restores them on the next launch.
       * The window size is no longer reset after searching or selecting an item.
       * The window will only expand automatically if the content requires more space, providing a more stable and predictable user experience.

  Technical Details

   * The "Tabs" feature is implemented using a new TabBarView and a Tab enum in AppState. The HistoryListView is updated to display items based on the selected tab.

https://github.com/user-attachments/assets/96275919-2432-4ddd-8c10-e0c3b0528c20 
  
https://github.com/user-attachments/assets/58b878d8-bb4d-4170-bf6b-379de7c90329
截屏2025-09-25 14 03 20


mithunsridharan:
Well, well, well! This is a cool feature. 
I have a small enhancement request, though: the current UI allows seamless navigation via keyboard.
If you could also assign keyboard shortcuts to directly go to the History or Pins pane, that would make this feature even more compelling, so the user wouldn't have to use the mouse at all!

alexisgaziello:
How can you expect merging Chinese into a repo that is standardized in English... 

But also I don't see the benefit. Its nice to see everything in one window.

mithunsridharan:
I tested this feature with my local development branch and though it's a cool feature that reduces clutter, it makes the usability overall clunky, especially when navigating the application via keyboard. Personally, I feel that the current experience with pins at the top and unpinned at the bottom gives a single pane to navigate via keyboard, which is easier to use, IMO

p0deje:
I suppose the idea is to reduce clutter when you have many pinned items that occupy a lot of space. Their list is currently unlimited, and we need to figure out a way to make it more organized. There is a long-standing issue with multiple proposals on this topic - #307 and in https://github.com/p0deje/Maccy/issues/1125#issuecomment-3297736397 @romanr suggested a similar tab-based approach. I'm still not sure which option would be better. I'm not a heavy pinned items user myself, so it's hard for me to make a decision. Any feedback and suggestions will be appreciated!

mithunsridharan:
@p0deje As a heavy pins user, I'm leaning towards ergonomics and efficiency of using only the keyboard, this feature could improve usability if:

- Pinned tabs are primary (default) and History tabs secondary in the popup
- Option to navigate the tabs using keyboard shortcuts, e.g., Option + H for History and Option + P for pinned

Intuitively, pins are curated items that I believe that users will access a few (if not many) times any given day, so positioning them as default sounds logical from a usability perspective.

I can't emphasize the amount of time I've saved using the Paste automatically feature in tandem with keyboard navigation, especially considering the number of times I've had to log on to websites / applications using credentials that I've pinned.

This App has definitely made my life easier and with such improvements as navigating entirely using the keyboard, this would definitely save a few hours each week!

romanr:
> the number of times I've had to log on to websites / applications using credentials that I've pinned.

You can save even more time by using a password manager for credentials. You would be able to log in with one click or one keyboard shortcut, and your credentials wouldn't be exposed to other apps through the clipboard. I don't see how your feedback can be considered when you're using Maccy for something other than its intended purpose. And you are very vocal about your feedback because regular use does not satisfy your need to use maccy as password manager.
This is an example of feedback subjectivity and how dangerous it is to make decisions based on arbitrary feedback, which, as in this case, may turn out to be rather harmful than useful.

weisJ:
I think it would be a good idea to discuss possible solutions for this problem before rushing to implement something. This is a major feature, which we have to get right otherwise it will be a pain if we choose a wrong direction here.

One option I see is to add the option to "tag" (maybe only pinned) history items (just like files in finder) and enhance the search functionality to be able to search for tags. In addition the pins section could be converted to a scroll view to declutter the UI.

We can use this issue https://github.com/p0deje/Maccy/issues/1286 for discussion.

PR #1247, opened 2025-11-04
Add Pinned Management Section and UI Polish

pokocat:
image
**Motivation**
Pinned items are a user's curated shortlist. Giving them a dedicated area and an easy way to reorder or unpin them makes day-to-day use faster and more intuitive. Persisting the pinned order removes the frustration of reordering after every app relaunch. The UI tweaks keep the numeric shortcut badges readable and add a convenient Pin action to normal history items without cluttering the interface.
**Changes**

- New PinnedSectionView at the top of the popup:

Shows a "Pinned" header with a "Manage"/"Done" toggle.
In manage mode, each pinned item exposes "Unpin" and "Move Up"/"Move Down" controls.
Adds a divider between pinned and unpinned lists when there’s no search query and both sections are present.

- Stable pinned order persistence:

Defaults.pinnedOrder stores the user-defined pin order.
Sorter.sort now respects Defaults.pinnedOrder when assembling the list, honoring the user's pins position preference.
History.togglePin updates Defaults.pinnedOrder on pin/unpin actions; History.movePinnedUp/Down adjusts the order and persists it.

- UI Refinements:

Increased the trailing padding of pinned item management controls to prevent them from overlapping with the right-side shortcut badges.
Added a hover-only quick "Pin" button on the right side of unpinned items.
When numeric shortcut badges are present, a thin separator is inserted to visually distinguish them from the new "Pin" action.

### DIFF
--- a/Maccy.xcodeproj/project.pbxproj
+++ b/Maccy.xcodeproj/project.pbxproj
@@ -126,6 +126,7 @@
 		DAEE38471E3DBEB100DD2966 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = DAEE38461E3DBEB100DD2966 /* AppDelegate.swift */; };
 		DAFE2DDA268A521B00990986 /* String+Shortened.swift in Sources */ = {isa = PBXBuildFile; fileRef = DAFE2DD9268A521A00990986 /* String+Shortened.swift */; };
 		DAFE2DE9268A9B1B00990986 /* HistoryItemTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DAFE2DE8268A9B1B00990986 /* HistoryItemTests.swift */; };
+		DAFED0C22D5E0A5B00ABCDEF /* PinnedSectionView.swift in Sources */ = {isa = PBXBuildFile; fileRef = DAFED0C12D5E0A5B00ABCDEF /* PinnedSectionView.swift */; };
 		DAFEF0B8249D7DEE006029E8 /* KeyboardShortcuts.Name+Shortcuts.swift in Sources */ = {isa = PBXBuildFile; fileRef = DAFEF0B7249D7DEE006029E8 /* KeyboardShortcuts.Name+Shortcuts.swift */; };
 /* End PBXBuildFile section */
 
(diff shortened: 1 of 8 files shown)


mithunsridharan:
@pokocat Nice feature and idea to manually sort the pins. I do have a couple of enhancement ideas that I'd like you to consider:

- Include a shortcut to move the pinned items up / down, e.g., using the arrow keys / drag and drop via mouse in managed mode. I find clicking the up / down arrows using the mouse to be somewhat cumbersome, especially if one has a lot of pinned items as I do
- I've attempted to allow sorting via Pin Keys (check PR #1322 ). Could you integrate the automatic and manual sorting logic, please?

Thanks for this cool feature!

PS: I've updated the PR with a (hopefully) better implementation of the sorting logic

p0deje:
@pokocat This is really cool, thank you for sharing this PR and working on it. I would like to see if we can push it even further and implement a drag-and-drop operation on the history items. This way, we won't even need to implement a separate "Manage" mode because users can click on the item, drag it to the position they want, and this would update the order.

mithunsridharan:
> @pokocat This is really cool, thank you for sharing this PR and working on it. I would like to see if we can push it even further and implement a drag-and-drop operation on the history items. This way, we won't even need to implement a separate "Manage" mode because users can click on the item, drag it to the position they want, and this would update the order.

To add my 2 cents: an option to move the items via keyboard, say via Control / Option key. That would be awesome!

pokocat:
> @pokocat This is really cool, thank you for sharing this PR and working on it. I would like to see if we can push it even further and implement a drag-and-drop operation on the history items. This way, we won't even need to implement a separate "Manage" mode because users can click on the item, drag it to the position they want, and this would update the order.

good advise! but I noticed that there're already hover event on each item, I have to find a way to do so. But I’ve got my hands full at the moment, I'll continue working on it later..

mithunsridharan:
@pokocat Please take a look at #1263 and #1265 (needs fixing, but you might want to check whether you could re-use some code / logic from this PR). The issue with #1265 is that there's a conflict between automatic and manual sorting that I haven't managed to fix yet, but hopefully, you have a head start here.

Maccy.xcodeproj/project.pbxproj
```diff
@@ -1663,7 +1667,7 @@
 				COMBINE_HIDPI_IMAGES = YES;
 				CURRENT_PROJECT_VERSION = 58;
 				DEAD_CODE_STRIPPING = YES;
-				DEVELOPMENT_TEAM = MN3X4648SC;
+				DEVELOPMENT_TEAM = VN2LG9M3Z8;
```
weisJ, line 1670: Avoid these changes

Maccy/Observables/AppState.swift
```diff
@@ -11,6 +11,8 @@ class AppState: Sendable {
   var popup: Popup
   var history: History
   var footer: Footer
+  // 管理固定区域的开关（弹出面板顶部“Pinned”区域的“管理”模式）
```
weisJ, line 14: Please keep code comments in english.

Maccy/Sorter.swift
```diff
@@ -24,9 +24,36 @@ class Sorter {
...
+          result.append(item)
+        }
+      }
+      // Then, any remaining pins (e.g., new ones) keeping their algorithm order
+      for item in pinned where !order.contains(item.pin ?? "") {
+        result.append(item)
+      }
+      orderedPinned = result
+    }
+
```
weisJ, line 51: Why not just keep the much shorted list of only pinned items sorted?

weisJ (changes requested):
As I mentioned in #1322 the way to go forward is to add a separate sorting option for pinned items only. Sorting manually should go in there as a separate option.
This PR needs to coordinate with #1322 in this regard.

Also I don‘t think putting the sorting UI into the history list this way is a good idea. Either
- Implement drag and drop reordering of items in the pinned section when manual sorting is enabled in the settings
or (probably easier)
- Make the list of pins in the settings manually reorderable. Preferably also through drag and drop handling. This should also only be possible if manual sorting is enabled.

PR #1260, opened 2025-11-21
Fix flashing beam cursor at top of popup

weisJ:
With the search bar hidden the text beam cursor would appear when hovering at the top of the window.


p0deje:
Can you show a screenshot? I can't reproduce.

p0deje:
Ah, I see now.

PR #1262, opened 2025-11-22
unlimited history

### LINKED ISSUE #310: Ability to store unlimited clipboard data
Thanks for this app. A great alternative to Ditto Clipboard manager on Windows. Ditto has an additional ability to store unlimited clips ( I know it is sounds too much), but maybe you can add that too), or instead of slider we can actually input a number?

Again, thanks for this wonderful app.

jaspermayone:
this should implement unlimited history. It did work locally for me, but i'd be curious for other people's thoughts / testing on implementation

closes #310


jaspermayone:
@p0deje lmk your thoughts on this. I know you were possibly waiting til some things changed, but I figured I would take a stab at it now.

weisJ:
I was only able to skim over the code, but this looks quite good to me. I'll try to do a full review soon. Thank you for tackling this problem.

Maccy/Observables/History.swift
```diff
@@ -70,6 +70,20 @@ class History { // swiftlint:disable:this type_body_length
...
+  @ObservationIgnored
+  private var currentPage = 0
+  @ObservationIgnored
+  var totalCount = 0
+  @ObservationIgnored
+  var isLoadingMore = false
+  @ObservationIgnored
+  var hasMoreItems: Bool {
+    all.count < totalCount
+  }
```
weisJ: Maybe the pagination logic could be encapsulated in a separate class

Maccy/Observables/History.swift
```diff
@@ -122,6 +153,38 @@ class History { // swiftlint:disable:this type_body_length
     }
   }
 
+  @MainActor
+  private func loadPage() async throws {
+    let offset = currentPage * pageSize
+    var descriptor = FetchDescriptor<HistoryItem>(
+      sortBy: [SortDescriptor(\.lastCopiedAt, order: .reverse)]
```
weisJ: This should take the current sorting mode into account. We need to make sure that all items on the next page are after the last item of the current page in sorting order.

Maccy/Settings/StorageSettingsPane.swift
```diff
@@ -93,12 +95,35 @@ struct StorageSettingsPane: View {
       }
 
       Settings.Section(label: { Text("Size", tableName: "StorageSettings") }) {
+        Toggle(
+          isOn: Binding(
+            get: { isUnlimitedHistory },
+            set: { newValue in
+              if newValue && size > 5000 {
```
weisJ: Is `5000` deliberately the same as `historySizeThreshold`? If yes, it should be a constant referenced at both places.

Maccy/Search.swift
```diff
@@ -35,13 +35,23 @@ class Search {
...
 
-    switch Defaults[.searchMode] {
+    var effectiveSearchMode = Defaults[.searchMode]
+
+    // Auto-disable fuzzy search for large histories
+    if Defaults[.isUnlimitedHistory] && History.shared.totalCount > historySizeThreshold {
+      if effectiveSearchMode == .fuzzy || effectiveSearchMode == .mixed {
+        effectiveSearchMode = .exact
+      }
+    }
```
weisJ, line 52: This should be shown as a warning in the settings for the search mode.

Maccy/Observables/History.swift
```diff
@@ -122,6 +153,38 @@ class History { // swiftlint:disable:this type_body_length
...
+    )
+    descriptor.fetchLimit = pageSize
+    descriptor.fetchOffset = offset
+
+    let results = try Storage.shared.context.fetch(descriptor)
+    let newItems = sorter.sort(results).map { HistoryItemDecorator($0) }
+    all.append(contentsOf: newItems)
```
weisJ: Previous pages should be unloaded, otherwise after a long usage we might be in the situation where the whole history is in-memory anyway.

weisJ (commented):
Mostly looking good to me, just a few comments. Currently the next page is fetched at the end of the current with a page size of 200. How about the following:

- Store (up to) 3 pages at a size of `100` each. The previous, the current and the next page.
- When reaching the end of the current page, load the page after the next and swap to the next page i.e.
	````
	previous = current
	current = next
	next = fetchNextPage()
	````
	(Similar scenario when scrolling up).
- Because the first page (and maybe last page) are frequently accessed (and easily jumped to with `CMD + UP/DOWN`, they could get special treatment and stored separately to avoid loading times.
- If a new item is added we need to make sure to remove it from any previous page it may be contained in and append it to the first (newest) page. Once the first page reaches 200 (or twice whatever the base page size is) it is split into two pages to prevent unlimited page size growing. Similar rebalancing needs to happen if the page the item is removed from grows too small (just discard the affected pages and reload them maybe).

I think this way we can almost certainly eliminate loading times when scrolling.
	In particular it does not keep the previous pages unnecessarily in memory when scrolling.

jaspermayone:
Will address!

Maccy/Observables/History.swift
jaspermayone: closed in 8d02b0b525b561f54f0a0ea0630e8f0566212f48

Maccy/Observables/History.swift
jaspermayone: closed in 8d02b0b525b561f54f0a0ea0630e8f0566212f48

Maccy/Settings/StorageSettingsPane.swift
jaspermayone: closed in 8d02b0b525b561f54f0a0ea0630e8f0566212f48

Maccy/Search.swift
jaspermayone: closed in 8d02b0b525b561f54f0a0ea0630e8f0566212f48

Maccy/Observables/History.swift
jaspermayone: closed in 8d02b0b525b561f54f0a0ea0630e8f0566212f48

jaspermayone:
Should be good for re-review @weisJ!

weisJ:
Some more things I noticed:

When changing the settings for unlimited history the storage needs to recomputed to accommodate the new value.
Currently `loadPreviousItems()` is not called at all, so scrolling up does not work. In general SwifUI is a bit confused
when scroll ing because the list is (almost) always at the constant size of `pageSize`.
If you scroll to the end the content gets replaced but the scroll position is still "at the end of the list", which causes the pagination mechanism to immediately load all of the history and fast forward to the last page.

This wasn't a problem with the previous version, as there the list size increased linearly with more pages being loaded into memory. We probably need a custom scrollable list implementation. I think something as described here would be sufficient: https://nilcoalescing.com/blog/CustomLazyListInSwiftUI/

It assumes that every item has the same height, which is almost true for Maccy. There are two different heights, one for text and one for images. I cannot say how much more complicated that makes the implementation. Maybe precomputing the number of image entries per page could help...

The benefit with the custom scroll view approach would be that the scrollbar always reflects the actual size of the history list. Not just what is currently in memory.

Maccy/Observables/PaginationManager.swift
```diff
@@ -0,0 +1,325 @@
...
+    currentPage.removeAll { $0.item == decorator.item }
+    nextPage.removeAll { $0.item == decorator.item }
+
+    // Insert at beginning of first page cache
+    firstPageCache.insert(decorator, at: 0)
+
+    // If we're on the first page, also update currentPage
+    if currentPageIndex == 0 {
+      currentPage.insert(decorator, at: 0)
+    }
```
weisJ: This could be avoided by storing pages not as lists by classes with a list inside. The current and first page could then just reference the same object.

Maybe this is also an opportunity to make `Page` adhere to the sequence protocol to avoid creating the concatenated array in `allLoadedItems`.

Maccy/Observables/PaginationManager.swift
```diff
@@ -0,0 +1,325 @@
...
+  /// Handle new item being added to history
+  @MainActor
+  func handleNewItem(_ decorator: HistoryItemDecorator) {
+    // Remove from all pages if duplicate exists
+    previousPage.removeAll { $0.item == decorator.item }
+    currentPage.removeAll { $0.item == decorator.item }
+    nextPage.removeAll { $0.item == decorator.item }
+
+    // Insert at beginning of first page cache
+    firstPageCache.insert(decorator, at: 0)
```
weisJ: This assumes that new items always are placed at the beginning. This is not true for e.g. "sort by time first copy" or "sort by number of copies".

Maccy/Observables/PaginationManager.swift
```diff
@@ -0,0 +1,325 @@
...
+
+  // MARK: - Private Helpers
+
+  @MainActor
+  private func fetchPage(at pageIndex: Int) async throws -> [HistoryItemDecorator] {
+    let offset = pageIndex * pageSize
+    var descriptor = FetchDescriptor<HistoryItem>(
+      sortBy: [sortDescriptorForCurrentMode()]
+    )
+    descriptor.fetchLimit = pageSize
```
weisJ: This currently assumes that all pages have the same size. This is not the case e.g. if something gets deleted or inserted. I have to think about that a good solution would be here. Maybe just invalidate the pages and refetch previous, current and next?

Maccy/Observables/PaginationManager.swift
```diff
@@ -0,0 +1,325 @@
...
+
+    currentPageIndex = lastPageIndex
+    currentPage = lastPageCache.isEmpty ? try await fetchPage(at: lastPageIndex) : lastPageCache
+    nextPage = []
+    previousPage = lastPageIndex > 0 ? try await fetchPage(at: lastPageIndex - 1) : []
+  }
+
+  /// Handle new item being added to history
+  @MainActor
+  func handleNewItem(_ decorator: HistoryItemDecorator) {
```
weisJ: You handle insertion, but not yet deletion (except for `clear` and `clearAll`).

weisJ (changes requested):

jaspermayone:
will implement latest rounds of feedback.

weisJ:
Sorry for taking so long to look into the new changes. My time was taken up by #1296. I will have a look later this week.

jaspermayone:
All good, 1296 looks super cool!

jaspermayone:
Following up @weisJ :)

weisJ:
I am sorry I still didn‘t get to fully go over the changes, so here are my preliminary impressions.

I tried out the code and manually set the page size to something small (10). The history list kept jumping while scrolling which is rather undesirable. Ultimately changing the underlying content of a list while scrolling will have to make use of a custom layout algorithm as described in https://nilcoalescing.com/blog/CustomLazyListInSwiftUI/. Otherwise we will not be able to get rid of the jumping.

jaspermayone:
> I am sorry I still didn‘t get to fully go over the changes, so here are my preliminary impressions.
> 
> I tried out the code and manually set the page size to something small (10). The history list kept jumping while scrolling which is rather undesirable. Ultimately changing the underlying content of a list while scrolling will have to make use of a custom layout algorithm as described in [nilcoalescing.com/blog/CustomLazyListInSwiftUI](https://nilcoalescing.com/blog/CustomLazyListInSwiftUI/). Otherwise we will not be able to get rid of the jumping.

All good @weisJ! I'll take a poke at this over the weekend.

jaspermayone:
@weisJ this should be a bit better!

jaspermayone:
Hey @weisJ, just bumping this :)

Maccy/Views/VirtualizedLayout.swift
```diff
@@ -0,0 +1,191 @@
...
+  let itemHeight: CGFloat
+  let totalItemCount: Int
+  let loadedRange: Range<Int>
+
+  func sizeThatFits(proposal: ProposedViewSize, subviews: Subviews, cache: inout ()) -> CGSize {
+    let width = proposal.width ?? 0
+    // Total height is fixed based on total item count - never changes during scrolling
+    let height = CGFloat(totalItemCount) * itemHeight
+    return CGSize(width: width, height: height)
+  }
```
weisJ: This needs to take into account the number of items with an image, as they have a different height from normal items.

jaspermayone: fixing!

jaspermayone:
@weisJ this is good for re-review whenever you have time :)

weisJ:
Thanks for the bumb. Sadly I have been swamped lately. Trying to find some time in the next few days though :)

jaspermayone:
Hey @p0deje or @weisJ! Checking in :)

jaspermayone:
tysm, appreciate all the review time to get this over the hurdle to merge!

floriandotorg:
@p0deje or @weisJ, please please please merge this. Maccy is useless without unlimted history, it's like a car that can only go 5 MPH. And there is literally no other good clipboard manager out there. You're so close to making this amazing, please don't stall at the last mile.

weisJ:
I finally got around testing out the latest changes. Here are immediate issues I found:
- with images in the history items get the incorrect height. Sometimes images are not tall enough and other non-image items are the height of images
- When scrolling fast loading is not reliably. This is very likely a limitation of the sentinel value design. Ideally what we need to do is to be able to map a scroll position to the page index and request that from the page manager.

To make testing easier the page manager and scroll view should be abstracted to allow arbitrary items and sources. I did some of those changes locally to build a testing view that serves just a very large list of numbers, where small pages (size 20) are computed on the fly (currently not at home, so I can’t push those changes, but they are fairly trivial). Whatever design we land on should be able to facilitate such a list. Maybe for a start try to get the layout/scrolling etc working for only fixed item heights. After that is resolved we can think about how to support images.

jaspermayone:
hey @weisJ, sorry for the delay, working on implementation now.

Maccy/Observables/PaginationManager.swift
```diff
@@ -0,0 +1,393 @@
...
+  private func refreshImageItemIndices() throws {
+    let descriptor = FetchDescriptor<HistoryItem>(sortBy: [sortDescriptorForCurrentMode()])
+    let results = try Storage.shared.context.fetch(descriptor)
+    let sorted = sorter.sort(results)
+    var indices: [Int] = []
+    for (index, item) in sorted.enumerated() where item.image != nil {
+      indices.append(index)
+    }
+    imageItemIndices = indices
+  }
```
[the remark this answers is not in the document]
jaspermayone: Reworked: image presence is now determined from `HistoryItemContent` types via a predicate (no image data is loaded or decoded), and the index pass over sorted items touches metadata only — no in-memory re-sort. Still O(n) per mutation, but that cost is inherent to needing global row indices for the layout.

Maccy/Observables/History.swift
```diff
@@ -99,16 +141,33 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
...
+      all = paginationManager.allLoadedItems
+    } else {
+      // Load all items for limited history
+      let descriptor = FetchDescriptor<HistoryItem>()
+      let results = try Storage.shared.context.fetch(descriptor)
+      all = sorter.sort(results).map { HistoryItemDecorator($0) }
+      limitHistorySize(to: Defaults[.size])
+    }
+
+    items = all
```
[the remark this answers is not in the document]
jaspermayone: Partially addressed: while a search is active the popup now falls back to the standard (in-memory) list, so results render correctly — but search still only matches items in the loaded window. Searching storage directly in unlimited mode is a real gap; I think it deserves its own follow-up rather than growing this PR further.

Maccy/Observables/History.swift
```diff
@@ -117,6 +176,66 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
...
+    do {
+      try await paginationManager.jumpToLast()
+      all = paginationManager.allLoadedItems
+      items = all
+      updateUnpinnedShortcuts()
+      AppState.shared.popup.needsResize = true
+    } catch {
+      logger.error("Failed to jump to last: \(error.localizedDescription)")
+    }
+  }
```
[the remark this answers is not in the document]
jaspermayone: The pagination logic is now generic over a `PaginatedItemSource` and covered by `PaginationManagerTests` (synthetic integer source, page size 20) plus `VirtualListMetricsTests` for the row geometry. I did not add `HistoryTests` cases because that suite is skipped in `Maccy.xctestplan`, so they would not run.

jaspermayone:
@weisJ this should be good for eyes again

p0deje:
@jaspermayone Do you use it regularly? I tried pulling the PR and enabling the unlimited history, but the pagination doesn't seem to load anything
Screenshot 2026-07-31 at 11 15 15

jaspermayone:
I tested this out, so odd! I'll see if I can reproduce your issue

jaspermayone:
@p0deje Found it, thank you for the report. The root cause is that SwiftData can ignore `fetchOffset`. Every page fetch then returns the first 100 rows. All pages duplicate page 0, SwiftUI sees duplicate row identities, and the list renders only a few items inside a full-height scroll area. That matches your screenshot.

The fix is in 7053360. The source now takes one ordered snapshot of lightweight item faults and slices pages from it, so `fetchOffset` is no longer used. Page refreshes also apply atomically now, so a failed fetch cannot leave the count and the loaded pages inconsistent.

I added a regression test for the duplicate-page case, plus end-to-end tests that load a 650-item store with images and pins through the real SwiftData source. Can you retest when you have a moment?

jaspermayone:
@weisJ can you give this another pass?

Maccy/Search.swift
```diff
@@ -35,13 +35,23 @@ class Search {
...
+
+    switch effectiveSearchMode {
     case .mixed:
```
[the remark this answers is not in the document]
jaspermayone: Added `testFuzzySearchFallsBackToExactForLargeUnlimitedHistory` to SearchTests. The history count is injectable in `Search` now, so the test simulates a history above the threshold without inserting thousands of items.

Maccy/Observables/History.swift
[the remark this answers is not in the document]
jaspermayone: Since the last reply, `HistoryPaginationSourceTests` also exercises the real SwiftData source end to end: page slicing over a 650-item store with images and pins, pin exclusion, image row indices, and a regression test for the duplicated-pages bug. Together with the synthetic-source `PaginationManagerTests` this covers the pagination behavior, so resolving.

Maccy/Observables/History.swift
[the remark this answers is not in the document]
jaspermayone: Closing this one out: the code this comment referenced (`allLoadedItems`) no longer exists, and during a search the popup now uses the standard in-memory list, so results render correctly. Searching storage beyond the loaded window in unlimited mode remains a known limitation and is planned as a follow-up PR rather than growing this one.

jaspermayone:
@weisJ this is good for your review again as well

PR #1263, opened 2025-11-22
Pins sorting logic for popup and panels

mithunsridharan:
This PR introduces a sorting logic for pins. The pins can now be sorted in ascending or descending order based on:

- Time of last copy
- Time of first copy
- Number of copies, and
- Pin shortcut key (NEW!)

The sorting order can be set in the Storage pane:

tempImagefVZvSk

The Pins are sorted in the popup and Pins pane:

Popup:

tempImagexH9kuA

Pins pane:

tempImage6vwYFn

PR #1265, opened 2025-11-24
Re-arrange pins using Option keys

mithunsridharan:
This PR enables re-arranging pins using the Option key. Press Option key and move the pins to the desired location using the Arrow keys (Up / Down). Once a pin is at the desired location, release the Option key and voila!


mithunsridharan:
Conflict with #1263 Need further investigation

Issue #819, opened 2024-07-24
[V2.x] Move popover instead of hide and show


weisJ:
### Before Submitting Your Feature Request

- [ ] Check that there isn't already a similar feature request to avoid creating a duplicate.
- [ ] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Problem

I think the hide and immediate reshow of the popover when the selection changes is a bit unwieldy. I think it would be beneficial to transition the position of the popover in the case where the selection moves to an adjacent item. Thought?

p0deje:
I would not mind implementing this, but I don't think it's possible in SwiftUI without implementing custom views. A much simpler solution would probably be a complete disabling of popover show/hide animation, but I haven't found a way to do this in SwiftUI too.

weisJ:
I noticed this too today. The API is pretty locked down. And various versions to implement it using `NSViewRepresentable` seem to not reliably produce the same behaviour (the positioning seems to be a bit off).

p0deje:
I did some smaller improvements in https://github.com/p0deje/Maccy/commit/f1f3ae717846abf5fcecb7948e86077828df4928, maybe this approach can be used for moving popover too.

weisJ:
I haven't found a way to utilise this for moving the popover, however I have experimented with having a "detached" popover without an arrow, that updates the content if the selection changes. 

I also added a shortcut for opening the popover (`⌘I` by default) for manual control instead of the automatic delay.

Here is how it looks like:

https://github.com/user-attachments/assets/6cdcfa75-de71-48a3-a7f5-317da4706471

What do you think about this? Would it be worth while to add (some of) the following options:
- Toggle visibility manually with keyboard shortcut. (The automatic delay still being an option)
  - A selection change could then either dismiss the popover or have the current open/close behaviour.
- Have a "global" popover as in the video without an anchor. Maybe as an alternative option, though I'm not sure whether it is a good idea to bloat the settings in this regard.

p0deje:
I think it looks great overall, can you share a rough implementation how it works? I am not sure it makes sense to change this feature in 2.0, maybe postpone until 2.1 or consider a completely different approach to previews. I was thinking about potentially adding a slideout right sidebar with preview similar to https://cindori.com/developer/expandable-floating-panel but that's something for Maccy 3.x.

weisJ:
The code can be found here: https://github.com/weisJ/Maccy/tree/custom/

[...]

PR #1296, opened 2025-12-27
Slideout preview

weisJ:
Note: This builds on top of #1206. The relevant commits are the newest two.

Implements a basic version of the new preview concept proposed in https://github.com/p0deje/Maccy/issues/819#issuecomment-2269217020

Things up for discussion, which mostly aren't implemented yet:

- ~~Should the header span the whole width or only be on the left column?~~
- Should Maccy remember the open/close state of the preview sidebar across being opened?
- The blog post in the linked comment implements auto closing/opening the sidebar based on the width. Is this desirable?

Things that don't properly work yet:
- There is a minimum width on the history list (or a maximum width on the preview) preventing the sidebar to be resized larger than a certain value (around 400px).
- ~~No toolbar sidebar toggle button yet.~~
- ~~Translations for the settings are missing~~
- ~~If Maccy is opened to far on the right of the screen (such that the sidebar doesn't have enough space) then the sidebar should open on the left. The opening animation and layout for this is already working, just the detection of which sidebar placement to choose doesn't work yet.~~
- ~~Somehow resizing the window is a bit tricky. Not sure what the reason is.~~

Demo video:

https://github.com/user-attachments/assets/dfda4537-6e8f-43b9-92e5-31108040fc1a


weisJ:
New changes:
- Button to toggle preview sidebar
- A toolbar with buttons for delete and pin in the sidebar. This replaces the text hints in the preview footer.

Will provide screenshots/video of this later.

weisJ:
New changes:

- The option where the search field is only visible while typing is now handled correctly.
- Auto resizing (during search) now respects if the preview is open.
- Added a preview for the paste stack.

Also cleaned up the commit history a bit.

weisJ:
Here is a view of how it looks now.

https://github.com/user-attachments/assets/dced2de7-834a-4877-bed3-0fd7897629d8

As can be seen there are minimal kinks to work out if the appearance settings change. Though that should be easily fixed.
Besides that PR should be ready for review. It looks larger than it actually is due to the fact that it builds on top of the paste stack PR.

I still have to do the translations for the new UI strings.

- The default keyboard shortcut for opening the preview is `CMD + I`. This is the same as in Finder.
- I have turned the shortcut hints in the old preview into buttons in the toolbar. Something to consider is moving the footer items into the toolbar as well. In general I think there is potential for adding things there e.g. a quick toggle for the sorting order.

Future improvements which I think would be nice. These will be separate PRs (if feasible and desired).

- A "stacked cards" view in the preview if multiple items are selected.
- Providing image metadata (size, encoding etc.) in the new preview.
- Depending on what modifier keys are currently pressed (and what the setting of paste with formatting etc is) display a rich text version of the content or the plain text version. This would maybe make it more obvious which version will be pasted. Of course this cannot cover all cases.

clicktodev:
does the image preview still take up the entire screen if a picture is big enough?

weisJ:
The new preview panel is fixed width (but can be manually resized)

clicktodev:
can you post a screenshot of what it looks like for image previews?

i noticed the image height can be configured statically in the settings. I'm thinking if the preview window height and width is already fixed then wouldn't it be better if the image expanded to the full width of the preview window while having the swift ui equivalent of `object-fit contain` in CSS. i think it's `contentMode: .fit`.

this would make sure the image would take up all the available preview space while maintaining aspect ratio and without overflowing

ilkiri23:
It looks great! However, I noticed that the preview toggle button shifts when the preview panel opens on the right. I think keeping it fixed would reduce visual noise.

Is there a plan to add an option for hiding buttons?

Maccy/Observables/SlideoutController.swift
```diff
@@ -137,7 +144,20 @@ class SlideoutController {
...
+    }
+
+    if trigger == .manual {
+      if state.isOpen {
+        autoOpenSuppressed = true
+      } else {
+        autoOpenSuppressed = false
+      }
+    }
+
```
weisJ: Thinking about this more I believe the proper way for this is to provide additionally to `togglePreview` a `setPreviewState(.open/.closed)`. Then the auto open task can simply call `setPreviewState(.open)` and we don't have to cancel the task just because it would close the preview after opening it. This would avoid the whole `SlideoutToggleTrigger` and `autoOpenSuppressed` logic.

Just to be sure we should also check if Maccy is currently visible in `setPreviewState`/`togglePreview`. Just to avoid any racy conditions between the window closing and the auto opening timer firing.

Maccy/Observables/SlideoutController.swift
```diff
@@ -165,6 +165,9 @@ class SlideoutController {
         var newSize = window.frame.size
         newSize.width = contentWidth
         newSize = computeSizeWithPreview(newSize, state: self.state)
+        let alignedContentWidth = contentWidth.rounded()
+        let alignedSlideoutWidth = slideoutWidth.rounded()
+        newSize.width = alignedContentWidth + (self.state.isOpen ? alignedSlideoutWidth : 0)
```
weisJ: This duplicates the logic in `computeSizeWithPreview`. The easier fix is to use

````swift
  private var _contentWidth: CGFloat = 0
  var contentWidth: CGFloat {
    get { return _contentWidth }
    set {
      _contentWidth = max(minimumContentWidth, newValue).rounded()
      onContentResize(_contentWidth)
    }
  }
  private var _slideoutWidth: CGFloat = 400
  var slideoutWidth: CGFloat {
    get { return _slideoutWidth }
    set {
      _slideoutWidth = max(minimumSlideoutWidth, newValue).rounded()
      onSlideoutResize(_slideoutWidth)
    }
  }
````
Then we don't have to remember to put `.rounded()` everywhere too.

Maccy/Observables/SlideoutController.swift
p0deje: The main reason to cancel auto close task and maintain `SlideoutToggleTrigger` is to ensure that when user manual presses ⌃Space, the preview doesn't re-open until a different item is highlighted or window is fully closed.

p0deje:
> It looks great! However, I noticed that the preview toggle button shifts when the preview panel opens on the right. I think keeping it fixed would reduce visual noise.

@ilkiri23 The button no longer jumps, though there is no way to hide it now.

p0deje:
@clicktodev It works pretty much as you described:

Screenshot 2026-02-08 at 07 17 23

Maccy/Observables/SlideoutController.swift
weisJ: I see. I still think there are some simplifications possible but that can come later. I want to implement an option to disable the auto open task, which has to touch that part anyway.

PR #1319, opened 2026-01-31
Add support to extract text from images + Language Localization updates

SimplyMinimal:
I originally intended to create two separate pull requests but there was a dependency on the language localization to be present for the best UI experience. 

## Added Support to Extract Text from Images
image

I find myself wanting to quickly extract text from images from time to time. I have added the option to do so by pressing `Option+T`. 

## Updated Languages
For this, I've updated all the other localization strings and spot checked a few that I knew. They were correct so will assume the others are correct too. Without it, languages with missing strings will end up only showing the key in place.

**Disclaimer**: This was AI-assisted and I am relatively new to Swift so I am open to critiques for better code quality or polish. In my testing, it works as expected. Here's the testing scenarios I have done so far:
- Copied an image with short text and long text to clipboard then `Option+T` to extract the text. Both scenarios were successful.
- Copied an image with no text present and attempted to extract text. Maccy correctly will report "No text found in image" in the user's primary language set by the OS.
- Set my Mac's language to another language and can confirm that the language localization correctly updates the text strings

### DIFF
--- a/Maccy/Clipboard.swift
+++ b/Maccy/Clipboard.swift
@@ -67,7 +67,11 @@ class Clipboard {
     pasteboard.clearContents()
     pasteboard.setString(string, forType: .string)
     sync()
-    checkForChangesInPasteboard()
+
+    Task {
+      Notifier.notify(body: string, sound: .knock)
+      checkForChangesInPasteboard()
+    }
   }
 
   @MainActor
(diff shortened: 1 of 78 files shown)


weisJ:
The consensus was to implement this as „paste without formatting“ for images: #1152 

Also why did you duplicate all the UI strings for the preview? The were already available.

SimplyMinimal:
I didn't catch that there was already an open pull request for a similar feature until after the fact so apologies for missing this. Breaking this out into 3 parts.

Looking at #1152, it does make more sense to place the OCR function inside of "Paste without Formatting" shortcut. It looks like the last update on that was back on "Nov 14, 2025". Is the pending task remaining to fix the formatting of the code?

---

As for the UI strings, it was a result of this when I was adding the text extract feature. It would look like this (example when German is selected).
image

<p> After looking again, I realized I was looking in the wrong place for the languages so it's now been fixed in the new commits so it's no longer duplicated. CA and EL have been updated as it was missing values entirely.</p>
image


---

There is one bug that's been introduced. Placing the cursor close to the edge of the screen and bringing up the clipboard no longer respects the bounds of the screen. I've converted this pull request back to a draft for now for the following reasons:

- Consider next steps for #1152 
- Revert this pull request and properly break it out into two separate requests (one dedicated just for the language)
- Fix the UI clipboard placement bug if we decide to keep this.

Right idea, just need better execution. :) Thank you for your time and patience.

PR #1322, opened 2026-02-05
Feature: Sort pins alphabetically at both, popup and panels

mithunsridharan:
Due to some issues with my local development settings, a bunch of merges were transported to my previous PR. Raising a clean PR with the sorting feature to be merged in one of the upcoming branches.


Maccy/Extensions/Defaults.Keys+Names.swift
```diff
@@ -54,6 +54,7 @@ extension Defaults.Keys {
   static let showTitle = Key<Bool>("showTitle", default: true)
   static let size = Key<Int>("historySize", default: 200)
   static let sortBy = Key<Sorter.By>("sortBy", default: .lastCopiedAt)
+  static let sortOrder = Key<Bool>("sortOrder", default: false)
```
weisJ, line 63: I don‘t like this being a boolean.

Maccy/Settings/PinsSettingsPane.swift
```diff
@@ -117,13 +118,20 @@ struct PinsSettingsPane: View {
 
   @Query(filter: #Predicate<HistoryItem> { $0.pin != nil }, sort: \.firstCopiedAt)
   private var items: [HistoryItem]
+  
+  @Default(.sortBy) private var sortBy
+    @Default(.sortOrder) private var sortOrder
+    private var sortedItems: [HistoryItem] {
+      _ = sortOrder
+      return Sorter().sort(items, by: sortBy)
+    }
```
weisJ, line 163: Why is this ignoring the `sortOder`?

Maccy/Views/HistoryListView.swift
```diff
@@ -14,7 +14,14 @@ struct HistoryListView: View {
...
-    appState.history.pinnedItems.filter(\.isVisible)
+    let pinned = appState.history.pinnedItems.filter(\.isVisible)
+      if pinned.isEmpty { return [] }
+
+      // Use Sorter to order pinned items according to current Sorter settings
+      let sortedItems = Sorter().sort(pinned.map(\.item))
+      return sortedItems.compactMap { model in
+        pinned.first(where: { $0.item == model })
+      }
   }
```
weisJ, line 26: The proper way to do this is to make Sorter() be able to sort anything that can expose an `HistoryItem` e.g. by a `.sortBy(_ item: [T], by mapper: (T) -> HistoryItem)`.

(Writing this from memory. Syntax might not be entirely correct)

Maccy/Views/PreviewItemView.swift
```diff
@@ -58,6 +58,17 @@ struct PreviewItemView: View {
...
+          Text("NumberOfCopies", tableName: "PreviewItemView")
+          Text(String(item.item.numberOfCopies))
+        }
+        .padding(.bottom)
+        
+        HStack(spacing: 3) {
+          Text("PinShortcutKey", tableName: "PreviewItemView")
+          Text(item.item.pin ?? "")
+        }
+      .padding(.bottom)
```
weisJ, line 71: Lets put this in a separate PR and discuss it there.

Maccy/Sorter.swift
```diff
@@ -30,14 +33,19 @@ class Sorter {
   }
 
   private func bySortingAlgorithm(_ lhs: HistoryItem, _ rhs: HistoryItem, _ by: By) -> Bool {
+    let order = Defaults[.sortOrder]
```
weisJ: This should be an argument to the function.

Maccy/Sorter.swift
```diff
@@ -30,14 +33,19 @@ class Sorter {
...
+    let result: Bool
     switch by {
     case .firstCopiedAt:
-      return lhs.firstCopiedAt > rhs.firstCopiedAt
+      return lhs.firstCopiedAt < rhs.firstCopiedAt
     case .numberOfCopies:
-      return lhs.numberOfCopies > rhs.numberOfCopies
+      return lhs.numberOfCopies < rhs.numberOfCopies
+    case .pinShortcutKey:
+      result = (lhs.pin ?? "").localizedCaseInsensitiveCompare(rhs.pin ?? "") == .orderedAscending
```
weisJ, line 44: I think I have mentioned this before (and this line shows why): It does not make sense to habe `.pinShortcutKey` available as a general sorting strategy for all items. It only makes sense for pinned items. Therefore, you should add a separate sorting strategy enumerate and expose it through the pins settings. There you can make `.pinShortcutKey` available. Of course this can be written in a way to not duplicate the whole sorting logic.

weisJ (changes requested):

Maccy/Extensions/Defaults.Keys+Names.swift
mithunsridharan: Would you rather use something like 'asc' and 'desc'? With just 2 states - ascending and descending - binary was a simple choice.

Maccy/Settings/PinsSettingsPane.swift
mithunsridharan: I reckon this was already considered by line 125. Sorry, I'm not a Swift programmer and am trying to get this feature working with a blend of AI support, intuition and lots of search.

Maccy/Views/HistoryListView.swift
mithunsridharan: My attempt to include this feature introduced more changes to multiple file, which I found too invasive. Need Swift program support here!

Maccy/Sorter.swift
mithunsridharan: Fixed this in the latest push.

Maccy/Sorter.swift
mithunsridharan: I intuitively like the idea of restricting the sorting logic only to pins, but this calls for adding another visual element, which I've avoided. The behavior, however, affects only the pins on the popup menu and the pins setting pane.

mithunsridharan: @weisJ Looks like this logic, including the alphabetic sorting, impacts only the "Sort by" functionality. I don't think any further changes are needed, but please validate whether my observation is correct.

PR #1325, opened 2026-02-08
Use correct icon for preview toggle button

weisJ:
The left and right icon were the wrong way round


p0deje:
I actually thought it made more sense to flip them.

1. The left part of the icon shows a list with multiple entries. This resembles the main Maccy window.
2. The right part is large and shows the main contents of a single entry.

Screenshot 2026-02-10 at 07 07 47

weisJ:
I see your point. I guess it goes against my muscle memory because usually it is the other way round.

PR #1328, opened 2026-02-08
fix: show red border for invalid history size input

### LINKED ISSUE #1292: Setting history size fails silently if above 999

Version 2.3.0 (54)

### macOS Version

15.6.1 (24G90)

### Maccy Settings

```Shell
{
    "KeyboardShortcuts_delete" = "{\\"carbonKeyCode\\":51,\\"carbonModifiers\\":2048}";
    "KeyboardShortcuts_pin" = "{\\"carbonKeyCode\\":35,\\"carbonModifiers\\":2048}";
    "KeyboardShortcuts_popup" = "{\\"carbonModifiers\\":768,\\"carbonKeyCode\\":8}";
    NSNavPanelExpandedSizeForOpenMode = "{800, 448}";
    "NSWindow Frame NSNavPanelAutosaveName" = "267 156 800 448 0 0 1312 818 ";
    "NSWindow Frame com.sindresorhus.Settings.FrameAutosaveName" = "3409 713 510 233 0 0 5120 1415 ";
    SUEnableAutomaticChecks = 0;
    SUHasLaunchedBefore = 1;
        "public.rtf",
        "public.png",
        "public.utf8-plain-text",
        "public.tiff",
        "public.file-url",
        "public.html"
    );
    historySize = 999;
    ignoredApps =     (
        "com.1password.1password"
    );
    migrations =     {
        "2024-07-01-version-2" = 1;
    };
    pasteByDefault = 1;
    removeFormattingByDefault = 1;
    searchMode = exact;
    showFooter = 1;
    showSearch = 1;
    showTitle = 1;
    sortBy = lastCopiedAt;
    suppressClearAlert = 0;
    windowPosition = "[0.7675304878048781,0.8660598179453837]";
    windowSize = "[450,769]";
}
```

### Description

If I try to set my history size to above 999, it appears to accept this value, but in reality it ignores this and uses whatever I had entered previously.

### Steps to Reproduce

_No response_

claudeaceae:
## Summary

- Replaces the `NumberFormatter`-based TextField with a text-based TextField that validates input in real-time
- Shows a red border when the entered value is outside the valid range (1–999) or non-numeric
- On submit (Return key), invalid values revert to the last valid value
- Stepper updates sync back to the text field

This addresses the silent failure where typing a value above 999 appeared to be accepted but was actually ignored.

Fixes #1292

## Test plan

- [ ] Type a value above 999 → red border appears
- [ ] Type a non-numeric value → red border appears  
- [ ] Type a valid value (1–999) → no red border, value is saved
- [ ] Use stepper → text field updates correctly
- [ ] Press Return with invalid value → reverts to last valid value
- [ ] Close and reopen settings → value persists correctly

### DIFF
--- a/Maccy/Settings/StorageSettingsPane.swift
+++ b/Maccy/Settings/StorageSettingsPane.swift
@@ -62,9 +62,14 @@ struct StorageSettingsPane: View {
   @State private var viewModel = ViewModel()
   @State private var storageSize = Storage.shared.size
   @State private var sizeText = ""
-  @State private var isSizeValid = true
 
   private static let sizeRange = 1...999
+
+  private func isValidSize() -> Bool {
+    if sizeText.isEmpty { return true }
+    guard let value = Int(sizeText) else { return false }
+    return Self.sizeRange.contains(value)
+  }
 
   var body: some View {
     Settings.Container(contentWidth: 450) {
@@ -94,29 +99,24 @@ struct StorageSettingsPane: View {
           TextField("", text: $sizeText)
             .frame(width: 80)
             .help(Text("SizeTooltip", tableName: "StorageSettings"))
-            .border(isSizeValid ? Color.clear : Color.red)
+            .border(isValidSize() ? Color.clear : Color.red)
             .onAppear { sizeText = "\(size)" }
             .onChange(of: sizeText) {
-              if let value = Int(sizeText), Self.sizeRange.contains(value) {
-                isSizeValid = true
+              if isValidSize(), let value = Int(sizeText) {
                 size = value
-              } else {
-                isSizeValid = sizeText.isEmpty ? true : false
               }
             }
             .onSubmit {
-              if let value = Int(sizeText), Self.sizeRange.contains(value) {
+              if isValidSize(), let value = Int(sizeText) {
                 size = value
               } else {
                 sizeText = "\(size)"
-                isSizeValid = true
               }
             }
           Stepper("", value: $size, in: Self.sizeRange)
             .labelsHidden()
             .onChange(of: size) {
               sizeText = "\(size)"
-              isSizeValid = true
             }
           Text(storageSize)
             .controlSize(.small)
--- a/Maccy/Settings/StorageSettingsPane.swift
+++ b/Maccy/Settings/StorageSettingsPane.swift
@@ -61,13 +61,10 @@ struct StorageSettingsPane: View {
 
   @State private var viewModel = ViewModel()
   @State private var storageSize = Storage.shared.size
+  @State private var sizeText = ""
+  @State private var isSizeValid = true
 
-  private let sizeFormatter: NumberFormatter = {
-    let formatter = NumberFormatter()
-    formatter.minimum = 1
-    formatter.maximum = 999
-    return formatter
-  }()
+  private static let sizeRange = 1...999
 
   var body: some View {
     Settings.Container(contentWidth: 450) {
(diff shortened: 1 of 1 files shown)


Maccy/Settings/StorageSettingsPane.swift
```diff
@@ -61,13 +61,10 @@ struct StorageSettingsPane: View {
 
   @State private var viewModel = ViewModel()
   @State private var storageSize = Storage.shared.size
+  @State private var sizeText = ""
+  @State private var isSizeValid = true
```
prabhavagrawal7: much better to use isSizeValid as a method to perform checks. 
```suggestion
  private func isSizeValid() -> Bool {
    if sizeText.isEmpty { return true }
    guard let value = Int(sizeText) else { return false }
    return Self.sizeRange.contains(value)
  }
```

Maccy/Settings/StorageSettingsPane.swift
```diff
@@ -94,11 +91,33 @@ struct StorageSettingsPane: View {
...
       Settings.Section(label: { Text("Size", tableName: "StorageSettings") }) {
         HStack {
-          TextField("", value: $size, formatter: sizeFormatter)
+          TextField("", text: $sizeText)
             .frame(width: 80)
             .help(Text("SizeTooltip", tableName: "StorageSettings"))
-          Stepper("", value: $size, in: 1...999)
+            .border(isSizeValid ? Color.clear : Color.red)
+            .onAppear { sizeText = "\(size)" }
+            .onChange(of: sizeText) {
```
prabhavagrawal7, line 104: remove the redundent logic in `onChange` and `onSubmit` or atleast reuse them by creating another method.

prabhavagrawal7 (commented):
Added some suggestions for cleaner code.

claudeaceae:
Thanks for the review suggestions @prabhavagrawal7!

On converting `isSizeValid` to a computed method — I considered this, but `isSizeValid` is used as reactive `@State` that drives the border color via SwiftUI's view diffing. A computed method would work for initial evaluation, but the current approach lets us handle the edge cases cleanly (e.g., empty text during typing should show no border, and `onSubmit` resets validity when reverting to the last good value). That said, I'm open to the maintainer's preference here.

On the duplicated validation — good catch. I can extract a `validateSize()` helper to share the `Int(sizeText)` + range check between `onChange` and `onSubmit`. Will wait to see if the maintainer wants any changes before pushing an update though, since the two handlers do intentionally differ in behavior (onChange updates live, onSubmit reverts invalid input).

prabhavagrawal7:
Let me share the snippet for much cleaner approach, where the states will be well defined: 
```suggestion
  @State private var viewModel = ViewModel()
  @State private var storageSize = Storage.shared.size
  @State private var sizeText = ""

  private static let sizeRange = 1...999
  
  private func isValidSize() -> Bool {
    if sizeText.isEmpty { return true }
    guard let value = Int(sizeText) else { return false }
    return Self.sizeRange.contains(value)
  }

  var body: some View {
    Settings.Container(contentWidth: 450) {
      Settings.Section(
        bottomDivider: true,
        label: { Text("Save", tableName: "StorageSettings") }
      ) {
        Toggle(
          isOn: $viewModel.saveFiles,
          label: { Text("Files", tableName: "StorageSettings") }
        )
        Toggle(
          isOn: $viewModel.saveImages,
          label: { Text("Images", tableName: "StorageSettings") }
        )
        Toggle(
          isOn: $viewModel.saveText,
          label: { Text("Text", tableName: "StorageSettings") }
        )
        Text("SaveDescription", tableName: "StorageSettings")
          .controlSize(.small)
          .foregroundStyle(.gray)
      }

      Settings.Section(label: { Text("Size", tableName: "StorageSettings") }) {
        HStack {
          TextField("", text: $sizeText)
            .frame(width: 80)
            .help(Text("SizeTooltip", tableName: "StorageSettings"))
            .border(isValidSize() ? Color.clear : Color.red)
            .onAppear { sizeText = "\(size)" }
            .onChange(of: sizeText) {
              if isValidSize() {
                size = Int(sizeText) ?? size
              }
            }
            .onSubmit {
              if isValidSize(), let newSize = Int(sizeText) {
                size = newSize
              }
            }
          Stepper("", value: $size, in: Self.sizeRange)
            .labelsHidden()
            .onChange(of: size) {
              sizeText = "\(size)"
            }
          Text(storageSize)
            .controlSize(.small)
            .foregroundStyle(.gray)
            .help(Text("CurrentSizeTooltip", tableName: "StorageSettings"))
            .onAppear {
              storageSize = Storage.shared.size
            }
        }
      }

      Settings.Section(label: { Text("SortBy", tableName: "StorageSettings") }) {
        Picker("", selection: $sortBy) {
          ForEach(Sorter.By.allCases) { mode in
            Text(mode.description)
          }
        }
        .labelsHidden()
        .frame(width: 160, alignment: .leading)
        .help(Text("SortByTooltip", tableName: "StorageSettings"))
      }
    }
  }
}
```

claudeaceae:
@prabhavagrawal7 thanks for the detailed snippet! The computed `isValidSize()` approach is clean — I appreciate you taking the time to work it out.

You're right that this eliminates the explicit state tracking. One thing I want to preserve is the revert behavior on submit — when the user presses Enter with invalid input, the field should snap back to the last good value (so the red border resolves). The current code does `sizeText = "\(size)"` in the else branch of `onSubmit` for that. Your snippet's `onSubmit` silently ignores invalid input, which would leave the red border showing after Enter.

That said, I'll hold off on changes until the maintainer weighs in — they may have their own preferences on the approach. No point iterating between us if they want something different entirely.

weisJ:
I like the approach of @prabhavagrawal7. You can restore the text content on `onSubmit` if the size is invalid.

claudeaceae:
Thanks @weisJ! Pushed an update adopting the computed `isValidSize()` approach. The `onSubmit` handler now reverts to the last good value when invalid, so the red border clears on Enter.

PR #1338, opened 2026-02-16
Add Hex Color Swatch Show Setting in Preferences

umang350:
Summary

  - Adds a new "Show hex color swatch" toggle in the Appearance settings pane
  - When enabled (default), a 12x12 color swatch preview is shown next to clipboard items whose text is a valid hex color code (e.g. #ff8942, fff)
  - When disabled, hex color strings appear as plain text without a swatch

  Implementation

  - New showHexColorSwatch boolean preference key (default: true) in Defaults.Keys
  - ListItemView gates the accessoryImage rendering on the new setting, following the same @Default pattern used by showApplicationIcons
  - Defaults.Toggle added to AppearanceSettingsPane between "Show application icons" and "Show footer"
  - Localization strings added to all 41 locale files (English translated, others left empty for future translation)

  Files changed

  - Maccy/Extensions/Defaults.Keys+Names.swift — new key
  - Maccy/Views/ListItemView.swift — gate accessory image on setting
  - Maccy/Settings/AppearanceSettingsPane.swift — new toggle
  - Maccy/Settings/*/AppearanceSettings.strings — localization entries (41 files)

  Test plan

  - Open Preferences > Appearance, verify "Show hex color swatch" toggle is visible
  - Copy a hex color string (e.g. #ff8942), verify swatch appears when toggle is on
  - Disable the toggle, verify swatch no longer appears for hex color items
  - Re-enable the toggle, verify swatch reappears

### DIFF
--- a/Maccy/Views/ListItemView.swift
+++ b/Maccy/Views/ListItemView.swift
@@ -43,7 +43,6 @@ struct ListItemView<Title: View, ID: Hashable>: View {
   @ViewBuilder var title: () -> Title
 
   @Default(.showApplicationIcons) private var showIcons
-  @Default(.showHexColorSwatch) private var showHexColorSwatch
   @Environment(AppState.self) private var appState
   @Environment(ModifierFlags.self) private var modifierFlags
 
@@ -62,7 +61,7 @@ struct ListItemView<Title: View, ID: Hashable>: View {
       Spacer()
         .frame(width: showIcons ? 5 : 10)
 
-      if showHexColorSwatch, let accessoryImage {
+      if let accessoryImage {
         Image(nsImage: accessoryImage)
           .accessibilityIdentifier("copy-history-item")
           .padding(.trailing, 5)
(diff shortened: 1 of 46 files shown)


Maccy/Views/ListItemView.swift
```diff
@@ -61,7 +62,7 @@ struct ListItemView<Title: View, ID: Hashable>: View {
       Spacer()
         .frame(width: showIcons ? 5 : 10)
 
-      if let accessoryImage {
+      if showHexColorSwatch, let accessoryImage {
```
weisJ: This is not the correct place for this check. You need to pass `nil` to `accessoryImage` in `HistoryItemView`. In the future there might be other accessory images for different reasons.

weisJ (changes requested):
Is the color swatch showing really that bothering you? (I don't have nothing against this change)

It would be great if you could provide real translations instead of just stubbing them out.

Maccy/Views/ListItemView.swift
umang350: Moved the logic to correct location.

umang350:
> Is the color swatch showing really that bothering you? (I don't have nothing against this change)
> 
> It would be great if you could provide real translations instead of just stubbing them out.

Color swatch seems to be annoying when I am not dealing with HEX codes but just bare numbers. Just OCD and not able to figure out what the square box was until i read the source code to figure it out. I don't code everyday, just a PM.

Issue #1080, opened 2025-04-18
Maccy popup very slow when large clippings are in it


freebrowser1:
### Before Submitting Your Bug Report

- [x] I have verified that there isn't already an issue reporting the same bug to prevent duplication.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Maccy Version (see 'About' window)

2.3.0

### macOS Version

15.3

### Maccy Settings

```Shell
{
    "KeyboardShortcuts_delete" = "{\\"carbonModifiers\\":2048,\\"carbonKeyCode\\":51}";
    "KeyboardShortcuts_pin" = "{\\"carbonModifiers\\":2304,\\"carbonKeyCode\\":20}";
    "KeyboardShortcuts_popup" = "{\\"carbonModifiers\\":4352,\\"carbonKeyCode\\":18}";
    "LaunchAtLogin__hasMigrated" = 1;
    "NSStatusItem Preferred Position Item-0" = 414;
    "NSStatusItem Visible Item-1" = 0;
    "NSWindow Frame SUUpdateAlert" = "650 511 620 398 0 0 1920 1080 ";
    "NSWindow Frame com.sindresorhus.Settings.FrameAutosaveName" = "1985 206 532 460 1280 -280 1920 1080 ";
    SUAutomaticallyUpdate = 0;
    SUEnableAutomaticChecks = 0;
    SUHasLaunchedBefore = 1;
    SULastCheckTime = "2024-12-11 13:27:17 +0000";
    SUSendProfileInfo = 0;
    avoidTakingFocus = 1;
    enabledPasteboardTypes =     (
        "public.rtf",
        "public.tiff",
        "public.png",
        "public.html",
        "public.utf8-plain-text"
    );
    ignoredPasteboardTypes =     (
    );
    migrations =     {
        "2024-07-01-version-2" = 1;
    };
    searchVisibility = always;
    showFooter = 1;
    showSearch = 1;
    showTitle = 1;
    sortBy = lastCopiedAt;
    windowSize = "[450,800]";
}
```

### Description

When large clippings (text of several tens of kB) are in the clipboards, then popping up Maccy is slow (several seconds).
Can this be resolved by an optional setting to limit preview (not the actual clipping) to so many lines / bytes ?

riyadparvez:
I am facing the same issue. I suspect there's a memory leak somehwere. On my machine it was using >5gb of memory.

[...]

PR #1342, opened 2026-02-25
perf: use in-memory cache instead of full DB fetch in findSimilarItem

ZHLHZHU:
## Problem

  Every clipboard change triggered a full SwiftData fetch of all history items with no predicate, causing synchronous disk I/O on the main thread. When large text was copied, this blocked the main thread for several seconds before the hotkey could respond.

## Root Cause

  `findSimilarItem` was fetching all records from SQLite on every call:

  ```swift
  let descriptor = FetchDescriptor<HistoryItem>()
  Storage.shared.context.fetch(descriptor)  // full table scan, no predicate
```

  History already maintains self.all as a complete in-memory list. There was no need to hit the database at all.

## Related Issues
https://github.com/p0deje/Maccy/issues/1080

### DIFF
--- a/Maccy/Observables/History.swift
+++ b/Maccy/Observables/History.swift
@@ -445,17 +445,11 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
 
   @MainActor
   private func findSimilarItem(_ item: HistoryItem) -> HistoryItem? {
-    let descriptor = FetchDescriptor<HistoryItem>()
-    if let all = try? Storage.shared.context.fetch(descriptor) {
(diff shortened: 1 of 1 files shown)


kmarekspartz:
I think this might still have memory issues. @ZHLHZHU have you compared memory utilization between the trunk and your fix?

ZHLHZHU:
@kmarekspartz This change is intended to improve clipboard-change latency rather than reduce memory usage

ZHLHZHU:
looks like this change was already included in #1441, so it’s in master now. I’ll close this PR.

PR #1343, opened 2026-02-26
Configurable popup size (width/height) in Appearance settings

MGeovany:
Hi! I’m a big fan of Maccy, I use it every day. Thanks for building and maintaining it.

This PR adds a preference to customize the popup window size (width + max height). My main pain point was that the popup could feel “full screen” on my setup, covering most of the screen and making it hard to keep seeing what 
I’m working on while browsing clipboard history.

What’s included:
- New Appearance preference: Popup size (Width / Height) with a Reset button
- Height continues to behave as a maximum (popup can shrink if there are only a few items)
- Width is respected on open (keeps content width in sync so it doesn’t snap back)

Why merge to master:
- Improves usability for users on smaller screens / large history lists / multi-monitor workflows
- Makes the UI feel less intrusive without changing defaults
- Low risk: uses existing windowSize preference plumbing and stays within reasonable bounds

Notes:
- No behavior change unless the user modifies the new setting
- Default remains the same as current behavior

Screenshots:

image

### DIFF
--- a/Maccy/FloatingPanel.swift
+++ b/Maccy/FloatingPanel.swift
@@ -23,10 +23,10 @@ class FloatingPanel<Content: View>: NSPanel, NSWindowDelegate {
     self.onClose = onClose
 
     super.init(
-        contentRect: contentRect,
-        styleMask: [.nonactivatingPanel, .resizable, .closable, .fullSizeContentView],
-        backing: .buffered,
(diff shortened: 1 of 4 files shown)


weisJ:
It is already possible to resize the popup just as any other window by dragging the edge of the window.

Issue #581, opened 2023-07-18
Feature request: items folders for quick access but keeping space taken on screen small 


nicolnt:
This feature request is directly inspired from https://github.com/Clipy/Clipy
Issue is well described here: 
https://github.com/Clipy/Clipy/issues/508#issuecomment-1234234893

Search is not always convenient, so a list of items and search bar complement each other well. But to prevent long lists, we can use a folder system.

It looks like screenshot below:
<img width="400" alt="image" src="https://github.com/p0deje/Maccy/assets/12377588/84a586cb-926c-4f77-905f-f33045e10788">

Clippy configuration lets the user choose the number of items shown, which is pretty ideal:
<img width="475" alt="image" src="https://github.com/p0deje/Maccy/assets/12377588/1b56adb1-30e2-48b4-980f-6631f732e8a9">

p0deje:
Duplicate of #229

Issue #254, opened 2021-03-22
Feature Request - Ability to tag items with meta keywords and expose them as searchable


madhurjain2095:
Hi,

I am recent MacOS user and because of my work, I need to use a clipboard manager to copy and manage multiple meeting IDs (zoom/slack/chime) and I was wondering if there was in any way to tag them as such. Like, when you copy the meeting ID from the clipboard, there's no way to know which meeting ID corresponds to which meeting and as such, I end up going back to my calendar and copy the meeting ID again (which seems like a waste to me). I would like to tag such meeting IDs or in general, any item, with some extra meta keywords and make them searchable so that whenever I type in the meeting name, the ID should pop up and I can copy and join the meeting.

I understand that the use case I am depicting here is quite specific, but the addition of meta keywords can be very helpful in other scenarios too.

Please let me know what you think about this.

PS: I really love using Maccy since its very minimalist and has a lot of keyboard shortcuts.

p0deje:
This is very similar to #89 where aliases for pins are requested. I'm not sure when and if I'm going to implement this, but PRs are very much welcome!

Issue #176, opened 2020-09-21
Organize pinned items into folders


vurtomatic:
Hello @p0deje ! I'm exploring Maccy coming from Clipy. I love the UI like many Maccy users.

The only thing I find lacking so far is the ability to organize pinned items into folders.

p0deje:
I have no plans to add this feature. I don't think that the clipboard manager should be a note-taking app or snippets organizer. I will consider any PRs implementing it given this feature is intuitive and doesn't spoil current UX.

vurtomatic:
That's too bad, I think of it as a way to organize visually and reduce the vertical length of the list; especially since pinned items are meant to be "permanent"—I have pinned items for different purposes and find it convenient to group them categorically, also makes it easier to find them as I seldom maintain such a long list that I'd use search.

p0deje:
This is reasonable and I believe Clipy could be a better solution for your needs then!

vurtomatic:
I appreciate you hearing me out and discussing it! I'll use both for a while and hope you'll reconsider in future :)

rmasci:
This is the very feature I am looking for in a clipboard manager. I am not looking for a note-taking app, just something to store commonly used items.  Think about it, if I had this in a note somewhere, you have to go to the note, copy / paste into the application. Its a lot easier if its in a tool like maccy, select item, push button and it pastes in.  Until then...
==> Uninstalling Cask maccy
==> Backing App 'Maccy.app' up to '/usr/local/Caskroom/maccy/0.17.0/Maccy.app'.
==> Removing App '/Applications/Maccy.app'.
==> Purging files for version 0.17.0 of Cask maccy

Issue #156, opened 2020-08-26
Ability to always keep certain items and to label/name them


sprdnja:
Hi,

From time to time i copy and paste my server passwords and other things that are used at least once per day. It would be very convenient if we get the ability to always keep certain items in the copied items list and also to name/label them for easier searching. Because my server passwords are random characters it is very hard to remember characters from it for searching the list so naming it or labelling it would save us ton of job.

Thanks.

p0deje:
You can already keep items by pinning them using Option+P. As for labeling, it has been requested in #89 so I'm going to close this issue in favor of it.

PR #1350, opened 2026-03-04
feat: Tags on Pins

batuyilmazer:
## Description

Currently, all pinned items in Maccy are displayed in a flat list. As users pin more items (such as code snippets, email templates, API keys, etc.), it becomes harder to organize and find them. 

This PR introduces a lightweight **Tagging / Categorization** feature for pinned items. Users can now assign tags (e.g., Work, Code, Email) to their pins. This allows filtering the list seamlessly by simply typing `#code` in the Maccy search bar, keeping the UI minimalist while adding powerful organizational capabilities. 

Unlike heavy "folder" implementations, this tagging mechanism strictly adheres to Maccy's philosophy of keeping the UX simple, fast, and keyboard-centric.

## How is Tagging Different from the Existing Alias Feature?

While Maccy already supports **Aliases**, Tagging serves a fundamentally different purpose:

- **1:1 Naming vs. N:N Grouping:** An Alias gives a specific name to a single item *(e.g., naming `¯\_(ツ)_/¯` as `shrug`)*. **Tags** are meant for categorization, allowing multiple items to share the same tag *(e.g., tagging 10 different strings as `#code`)* and single items to have multiple tags *(e.g., `#work`, `#email`)*.
- **Precision Searching:** Searching for an alias keyword might bring up unrelated clipboard history that happens to contain that word. Searching with the `#` prefix *(e.g., `#templates`)* acts as a strict category filter, acting like virtual folders without the UI clutter.
- **Visual Clarity:** Tags can be visually represented as small, unobtrusive badges/pills next to pinned items, allowing users to scan and identify categories at a glance, whereas aliases replace or blend with the text content.

## Related Issues

This implementation satisfies community demand for organizing pinned items, addressing use cases from the following past issues:

* **Resolves / Addresses #254** - [Feature Request - Ability to tag items with meta keywords and expose them as searchable](https://github.com/p0deje/Maccy/issues/254) *(Note: Closed as duplicate of Aliases, but Tagging fundamentally solves the N:N grouping need requested here).*
* **Addresses #176** - [Organize pinned items into folders](https://github.com/p0deje/Maccy/issues/176) *(Note: As mentioned in the comments, this tagging approach fulfills the organizational need without spoiling the current UX with complex folder hierarchies).*
* **Related to #581** - [Feature request: items folders for quick access](https://github.com/p0deje/Maccy/issues/581)
* **Related to #156** - [Ability to always keep certain items and to label/name them](https://github.com/p0deje/Maccy/issues/156)
* **Related to #40** - [[Feature Request] snippets/folders like Clipy](https://github.com/p0deje/Maccy/issues/40)

## Implementation Highlights

- **Unobtrusive UX:** Tagging doesn't introduce complex new UI views like folders. It organically integrates into the existing pin mechanisms.
- **Search Integration:** Native search integration (e.g., typing `#tag`) feels fast and natural for existing macOS users.

## Example Screenshot of new UI
WhatsApp Image 2026-03-04 at 16 14 36


weisJ:
Thank you for the PR. This has actually been on my todo list (but further down), so it's great that you took a stab at it. i haven't checked the code too carefully yet, so for now just some notes on what I think the feature should look like.

- Are the tags displayed in the history list? Ideally one could display them like the tags in finder as a coloured dot. Of course this requires adding a colour selection for tags in the settings.
- It should be possible to tag an item from the preview pane toolbar.
- Support of searching by tag is the most important part. The most native approach would be to use a `https://developer.apple.com/documentation/appkit/token-field` under the hood. This way we can support searching by `tag:foo` with support for auto-completion of the available tags. In the future this could be extended to support other queries like time ranges or type (text, image, etc.).

Getting the searching feature right is rather important, as its is very difficult to change after people get used to a specific version of it. Would it be possible to you to look at implementing tokenised searching using `NSTokenField` with auto complete?

batuyilmazer:
> Thank you for the PR. This has actually been on my todo list (but further down), so it's great that you took a stab at it. i haven't checked the code too carefully yet, so for now just some notes on what I think the feature should look like.
> 
> * Are the tags displayed in the history list? Ideally one could display them like the tags in finder as a coloured dot. Of course this requires adding a colour selection for tags in the settings.
> * It should be possible to tag an item from the preview pane toolbar.
> * Support of searching by tag is the most important part. The most native approach would be to use a `https://developer.apple.com/documentation/appkit/token-field` under the hood. This way we can support searching by `tag:foo` with support for auto-completion of the available tags. In the future this could be extended to support other queries like time ranges or type (text, image, etc.).
> 
> Getting the searching feature right is rather important, as its is very difficult to change after people get used to a specific version of it. Would it be possible to you to look at implementing tokenised searching using `NSTokenField` with auto complete?

None of these 3 features implemented in my code. I found them highly logical and ux-friendly, so i am closing this PR to come up with new one, with these 3 feats implemented. Thanks!

PR #1358, opened 2026-03-17
Fix clipboard/source attribution, history edge cases, and storage recovery

visoar:
## Summary
- fix clipboard source-app detection by reading pasteboard `.source` first, with fallback to frontmost app
- fix footer confirmation flow to honor per-item `suppressConfirmation` state
- fix history duplicate handling edge cases:
  - return `nil` when storage fetch fails in `findSimilarItem`
  - keep `lastCopiedAt` strictly newer than `firstCopiedAt`
  - refresh list/shortcuts/resize for both pinned and unpinned insertion paths
  - correctly insert pinned duplicate updates when no removal index exists
- harden `Storage` initialization with recovery flow:
  - create missing directory
  - back up broken sqlite store
  - retry persistent store creation
  - fall back to in-memory store as last resort
- unskip `HistoryTests` in the xctest plan and stabilize related unit tests
- remove empty translation entries (`"key" = "";`) from `.strings` files to avoid shipping placeholders

## Validation
- `xcodebuild test -project Maccy.xcodeproj -scheme Maccy -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO -only-testing:MaccyTests`
  - Result: `** TEST SUCCEEDED **`


weisJ:
There seem to be loads of unrelated changes to the localisation. Please fix those.

PR #1362, opened 2026-03-22
Fix critical bugs: off-by-one crash, OCR data race, DB recovery, accessibility prompt

twitchyvr:
## Summary

Five targeted fixes for bugs found during code review — no new features, no model changes, no behavioral shifts.

### 1. Off-by-one crash in AppIntents (Get, Select, Delete)
**Files:** `Intents/Get.swift`, `Intents/Select.swift`, `Intents/Delete.swift`

The bounds check `items.count >= index` allows `items[items.count]`, which is out of bounds. For example, with 5 items, requesting item #5 yields `index = 4`, and `5 >= 4` passes the guard, but `items[4]` is valid — however, requesting item #6 yields `index = 5`, and `5 >= 5` passes the guard, then `items[5]` crashes. More critically, when `items.count == index` (boundary case), the access crashes.

**Fix:** Changed to `index >= 0 && index < items.count` — standard bounds checking that also prevents negative index access.

### 2. Data race in Vision OCR callback
**File:** `Models/HistoryItem.swift`

`recognizeTextHandler` runs on an arbitrary Vision framework queue and writes directly to `self.title`. Since `HistoryItem` is a `@Model` class, its properties should only be mutated on the main actor (the model context's actor). This is a data race that can corrupt SwiftData's change tracking.

**Fix:** Dispatch the title update to `DispatchQueue.main.async`.

### 3. Graceful database recovery instead of `fatalError`
**File:** `Storage.swift`

If `ModelContainer` creation fails (corrupt SQLite from forced quit, disk error, etc.), the app calls `fatalError()` and is permanently bricked until the user manually deletes `~/Library/Application Support/Maccy/Storage.sqlite`.

**Fix:** Catch the error, delete the corrupt `.sqlite`, `.shm`, and `.wal` files, and retry container creation. The user loses history but the app remains functional. Only `fatalError`s if re-creation also fails.

### 4. Cache compiled regexes in clipboard ignore check
**File:** `Clipboard.swift`

`shouldIgnore(_ item:)` creates new `NSRegularExpression` objects for every pattern on every clipboard poll (default: every 0.5s). Regex compilation is expensive.

**Fix:** Cache compiled regexes and only recompile when the user's pattern list changes. Also restructured to extract the string once before the loop.

### 5. Prompt user for accessibility permissions
**Files:** `Accessibility.swift`, `Clipboard.swift`

`Accessibility.check()` was a no-op when permissions were missing — it returned silently, and `paste()` would proceed to simulate keyboard events that did nothing. Users had no indication why paste wasn't working.

**Fix:** `check()` now returns a `Bool` and triggers the macOS accessibility permissions prompt via `AXIsProcessTrustedWithOptions` with the prompt option. `paste()` gates on the return value.

## Test plan

- [ ] Verify AppIntents work via Shortcuts app with boundary item numbers
- [ ] Copy an image to trigger OCR — confirm no crash or unexpected behavior
- [ ] Simulate corrupt database (rename Storage.sqlite to corrupt it) — verify app recovers
- [ ] Add ignore regex patterns — verify clipboard polling remains performant
- [ ] Revoke accessibility permissions — verify paste prompts for permissions instead of silently failing
- [ ] General regression: copy, paste, search, pin/unpin, delete all work as before

### DIFF
--- a/Maccy/Accessibility.swift
+++ b/Maccy/Accessibility.swift
@@ -3,9 +3,17 @@ import AppKit
 struct Accessibility {
   private static var allowed: Bool { AXIsProcessTrustedWithOptions(nil) }
 
-  static func check() {
+  /// Checks if the app has accessibility permissions.
+  /// If not, prompts the user to grant them via System Settings.
+  /// Returns true if permissions are granted, false otherwise.
+  @discardableResult
+  static func check() -> Bool {
     guard !allowed else {
-      return
+      return true
     }
+
+    let options = [kAXTrustedCheckOptionPrompt.takeRetainedValue(): true] as CFDictionary
+    AXIsProcessTrustedWithOptions(options)
+    return false
   }
 }
--- a/Maccy/Clipboard.swift
+++ b/Maccy/Clipboard.swift
@@ -114,7 +114,7 @@ class Clipboard {
 
   // Based on https://github.com/Clipy/Clipy/blob/develop/Clipy/Sources/Services/PasteService.swift.
   func paste() {
-    Accessibility.check()
+    guard Accessibility.check() else { return }
 
     // Add flag that left/right modifier key has been pressed.
     // See https://github.com/TermiT/Flycut/pull/18 for details.
--- a/Maccy/Clipboard.swift
+++ b/Maccy/Clipboard.swift
@@ -35,6 +35,10 @@ class Clipboard {
 
   private var sourceApp: NSRunningApplication? { NSWorkspace.shared.frontmostApplication }
 
+  // Cache compiled regexes to avoid recompiling on every clipboard poll
+  private var cachedIgnoreRegexps: [String] = []
+  private var cachedIgnoreRegexes: [NSRegularExpression] = []
+
   init() {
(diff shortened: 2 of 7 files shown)


twitchyvr:
## CI Note

The Bitrise CI run shows `testCopyRTF` failing (all 3 retries) — this is **not related to this PR**.

The failure is an RTF title parsing issue on the CI VM:
```
XCTAssertEqual failed: ("["foo", "DC74B875-..."]") is not equal to ("["foo", "bar"]")
```

The RTF `NSAttributedString` parser returns a UUID instead of the parsed text "bar" — likely a font/text services issue on the CI's macOS Sequoia VM. The test suite reports `38 tests, 6 failures (0 unexpected)`, confirming these are known/expected failures.

This PR does not modify:
- Any test files (`git diff upstream/master..fix/critical-bugs-and-performance -- MaccyUITests/` is empty)
- `generateTitle()` or `previewableText` (the RTF parsing path)
- Any RTF handling code

The only `HistoryItem.swift` change is dispatching the Vision OCR callback to the main queue — completely unrelated to RTF.

PR #1376, opened 2026-04-12
Fix: Additional unicode characters getting added to pinned items

batsdk:
Fix unicode/garbled text on paste from browsers and Notes
                                                                                                                                                                                                                           
  Two fixes:
                                                                                                                                                                                                                           
  1. Clipboard.swift — strips metadata-only pasteboard types (Chrome's customWebKitPasteboardData, chromiumSourceToken, Notes' notesRichText, etc.) before storing. These tags travel alongside the real content but       
  confuse paste targets, producing garbled or extra characters.
  2. History.swift — deep-copies HistoryItemContent objects when merging a duplicate item, instead of sharing the reference. The old item is deleted from Core Data immediately after, which was silently invalidating the 
  shared content objects and leaving the new item with blank/corrupted data.

### DIFF
--- a/Maccy/Clipboard.swift
+++ b/Maccy/Clipboard.swift
@@ -29,6 +29,11 @@ class Clipboard {
     .concealed,
     .transient
   ]
+  private let metadataTypes: Set<NSPasteboard.PasteboardType> = [
+    .fromMaccy, .source, .linkPresentationMetadata,
+    .customWebKitPasteboardData, .customChromiumWebData,
+    .chromiumSourceUrl, .chromiumSourceToken, .notesRichText
+  ]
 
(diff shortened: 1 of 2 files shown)


Maccy/Clipboard.swift
```diff
@@ -197,6 +202,7 @@ class Clipboard {
 
       types = types
         .subtracting(disabledTypes)
+        .subtracting(metadataTypes)
```
weisJ, line 205: Maccy deliberatly doesn't strip these types unless pasting without formatting is enabled. Some of those are needed for specific features for the relevant applications to function properly.

weisJ (commented):
Can you elaborate on what issue this is solving?

PR #1379, opened 2026-04-16
Auto-paste first item on quick shortcut tap  

czarny:
When "Paste by Default" is enabled and the search field is hidden, quickly tapping the popup shortcut (press and release modifiers without cycling) automatically selects and pastes the  first history item instead of leaving the popup open in toggle mode.
                                                                                                                                                                                                                                                                                         
How to test:                                                                                                                                                                                   
- Enable "Paste by Default" and disable "Show Search", tap the shortcut quickly — first item should be pasted automatically


LosManos:
Just curious: What is the difference between this and ordinary cmd-v?

czarny:
I would say it is matter of habit to the shortcut. If you have some records in the clipboard and later it turns out that the one you need is actually first one you dont need to select it additionally.

czarny:
@LosManos ping.

LosManos:
Yeah?

weisJ:
How would one actually open the popup in this case to search for other items?

Issue #1113, opened 2025-06-12
Flipped labels in macOS Tahoe 26.0


missuo:
### Before Submitting Your Bug Report

- [x] I have verified that there isn't already an issue reporting the same bug to prevent duplication.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Maccy Version (see 'About' window)

2.3.0 (54)

### macOS Version

26.0 Beta (25A5279m)

### Maccy Settings

```Shell
defaults read org.p0deje.Maccy
{
    "KeyboardShortcuts_delete" = "{\\"carbonModifiers\\":2048,\\"carbonKeyCode\\":51}";
    "KeyboardShortcuts_pin" = "{\\"carbonModifiers\\":2048,\\"carbonKeyCode\\":35}";
    "KeyboardShortcuts_popup" = "{\\"carbonModifiers\\":768,\\"carbonKeyCode\\":8}";
    "NSStatusItem Preferred Position Item-0" = 374;
    "NSStatusItem VisibleCC Item-1" = 0;
    "NSWindow Frame com.sindresorhus.Settings.FrameAutosaveName" = "897 628 510 473 0 0 2304 1271 ";
    SUEnableAutomaticChecks = 1;
    SUHasLaunchedBefore = 1;
    SULastCheckTime = "2025-05-15 13:10:23 +0000";
    SUUpdateGroupIdentifier = 1159928500;
    enabledPasteboardTypes =     (
        "public.utf8-plain-text",
        "public.tiff",
        "public.html",
        "public.file-url",
        "public.png",
        "public.rtf"
    );
    historySize = 400;
    menuIcon = scissors;
    migrations =     {
        "2024-07-01-version-2" = 1;
    };
    showFooter = 1;
    showSearch = 1;
    showTitle = 1;
    windowSize = "[450,800]";
}
```

### Description

The text will overlap and will be directly flipped upside down.

<img width="515" alt="Image" src="https://github.com/user-attachments/assets/6f2e6dc1-0821-45c4-978e-16a29c4cc6c5" />

### Steps to Reproduce

1. Update to macOS 26.0
2. Click the maccy icon on the menu bar.

Trychlik:
same

p0deje:
This is so weird, some text is flipped and some is not?

missuo:
[...]

PR #1382, opened 2026-04-20
Fix 100% CPU spin on macOS 26 caused by .truncationMode(.middle)

zhouyeyu:
## Problem
On macOS 26, Maccy's CPU usage climbs to 100% immediately after launch,
causing the app to become unresponsive. Reproduces on every open of the popup.

## Root Cause
`NSCoreTypesetter.__NSCoreTypesetterTruncateLine` (in UIFoundation) enters
an infinite loop when measuring `Text` views with `.truncationMode(.middle)`
under an unconstrained height size proposal.

The trigger path:
1. `SlideoutView` preview panel starts opening (isAnimating = true)
2. Content VStack gets `.fixedSize(horizontal: true, vertical: false)`
3. SwiftUI measures the content at ideal (unconstrained) height
4. Traverses down: ZStack → VStack → SlideoutView → ScrollView → LazyVStack
5. Each `HistoryItemView` title is measured via `NSAttributedString.boundingRect`
6. CoreText's truncation loop hangs indefinitely for certain text content

Confirmed via `sample` profiling: 100% of CPU time in
`__NSCoreTypesetterTruncateLine_block_invoke`.

## Fix
- Use `.truncationMode(.tail)` on macOS 26 in `ListItemTitleView`.
  The existing `.drawingGroup()` workaround for flipped text (#1113)
  is preserved and works correctly with tail truncation.
- (Defensive) `HeightReaderModifier`: write to `@Observable` keypaths
  directly instead of via a custom `Binding` to avoid unintended render
  dependency registration.
- (Defensive) `AppDelegate` / `HistoryItemDecorator`: guard
  `withObservationTracking` re-registration callbacks with a pending flag
  to prevent observation count from growing when multiple mutations fire
  in the same turn.

### DIFF
--- a/Maccy/AppDelegate.swift
+++ b/Maccy/AppDelegate.swift
@@ -22,6 +22,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
   }
 
   private var statusItemVisibilityObserver: NSKeyValueObservation?
+  private var menuIconTextUpdatePending = false
 
   func applicationWillFinishLaunching(_ notification: Notification) { // swiftlint:disable:this function_body_length
     #if DEBUG
@@ -161,7 +162,15 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     _ = withObservationTracking {
       AppState.shared.menuIconText
     } onChange: {
+      // Guard against multiple onChange callbacks firing before the async block
+      // runs (e.g. when History.add() mutates `items` and then updates shortcuts
+      // in the same turn). Without this guard each callback queues its own async
+      // block, and each block re-registers an observation, causing the number of
+      // active observations to grow unboundedly and CPU usage to spike over time.
+      guard !self.menuIconTextUpdatePending else { return }
+      self.menuIconTextUpdatePending = true
       DispatchQueue.main.async {
+        self.menuIconTextUpdatePending = false
         if Defaults[.showRecentCopyInMenuBar] {
           self.statusItem.button?.title = AppState.shared.menuIconText
         }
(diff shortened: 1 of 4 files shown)


jaspermayone:
I wasn't able to produce this on macos26....

Primer-Fan:
I can reproduce this deterministically, and I think I found the condition that
was missing — full details and a standalone repro in #1520.

The trigger isn't `.truncationMode(.middle)` alone. It needs a title containing
**two or more U+FFFC** (OBJECT REPLACEMENT CHARACTER) plus at least one
non-Latin character. Those arrive from rich-text copies with inline images: each
inline image becomes one U+FFFC in the `public.utf8-plain-text` flavor, so
copying a chat message with 2+ inline emoticons is enough. `generateTitle()`
takes `previewableText` verbatim, so they get persisted into
`ZHISTORYITEM.ZTITLE` and the app then hangs on **every** launch.

That's probably why @jaspermayone couldn't reproduce it — an ordinary history
won't do it. With Maccy idle at 0% CPU:

```sh
python3 -c "import subprocess; subprocess.run(['pbcopy'], input='￼复制粘贴测试￼'.encode())"
```

pegs it at 100% within a second.

One more shape dependency worth knowing when testing: in my standalone repro,
measuring a bare row does **not** hang — it needs the builder to emit a
`_ConditionalContent`, which is exactly what `ListItemTitleView`'s
`if let attributedTitle { … } else { … }` produces.

I verified your `.tail` change clears the hang. I also verified an independent
alternative — stripping U+FFFC in `generateTitle()` — which fixes it while
keeping middle truncation (useful for long URLs/paths where the distinguishing
part is at the end). The two are orthogonal and could both land; the full
condition matrix is in #1520.

zhouyeyu:
sounds great, seems the bug has been fixed. So this pr should be merged independently?

PR #1386, opened 2026-04-27
Add normalize whitespace option

spyrae:
## Summary

- Adds a new **"Normalize whitespace"** toggle in General → Behavior settings
- When enabled, normalizes whitespace **immediately on copy** — both Maccy history and regular Cmd+V produce clean text
- Useful when copying from terminal where word-wrap inserts extra whitespace

## How it works

- Applied in `checkForChangesInPasteboard()` when new clipboard content is detected
- Rewrites system clipboard with normalized text (marked `.fromMaccy` to prevent re-triggering)
- Also normalizes when pasting from Maccy history
- Original text is never stored — clipboard is cleaned on the fly
- Only affects `.string` content type — files, images, RTF untouched

## Normalization rules

- Spaces/tabs around newlines → single space (joins wrapped lines)
- Multiple consecutive spaces → single space
- Leading/trailing whitespace trimmed

## Changes

- `Clipboard.swift` — `normalizeWhitespace()` function, called in both `checkForChangesInPasteboard()` and `copy(_:removeFormatting:)`
- `Defaults.Keys+Names.swift` — new `normalizeWhitespace` key (default: `false`)
- `GeneralSettingsPane.swift` — toggle UI in Behavior section
- Localization: EN + RU

## Test plan

- [ ] Enable "Normalize whitespace" in General → Behavior
- [ ] Copy text with extra spaces from terminal → Cmd+V → spaces normalized
- [ ] Copy multi-line wrapped text → Cmd+V → lines joined, single spaces
- [ ] Paste same item from Maccy history → also normalized
- [ ] Disable toggle → copy text → original whitespace preserved
- [ ] File/image clipboard items unaffected


mithunsridharan:
Beware! If the source is Markdown, Python or other languages that rely on whitespaces for proper functionality, enabling this option would defeat the purpose! Please consider adding a validation (preferably including an option in the UI to exempt code modifications) to avoid "Gotcha" scenarios. 

Overall, this PR works! Today, I copied some Markdown code and this feature stripped the whitespaces, so had to turn this feature off. A setting that skips stripping whitespaces from code would be splendid and win-win!

spyrae:
Thanks for the feedback! Great catch - I've pushed an update that automatically detects code and structured content and skips whitespace normalization for it.

How it works:
The detection uses two layers:
1. Pasteboard type detection - when copying from code editors like VS Code, they include a special pasteboard type (vscode-editor-data). When detected, normalization is skipped entirely. This is a definitive signal with zero false positives.
2. Content-based heuristic - for any source (browsers, text editors, terminals, etc.), the content is analyzed for structural patterns where whitespace is semantically significant:
- Markdown code blocks - fenced with ``` or ~~~
- Indented code - Python, Swift, YAML, etc. (≥3 indented lines, or ≥20% of non-empty lines)
- Pipe tables - Markdown/ASCII tables with | delimiters
- Tab-separated data - lines with internal tab characters

 This means you no longer need to toggle the setting on/off when switching between prose and code - structured content is automatically preserved while plain text with extra whitespace is still normalized as expected.

Both paths are covered: normalization is skipped when content is first captured from the clipboard AND when it's pasted from Maccy history. Added tests for all detection scenarios.

PR #1387, opened 2026-04-29
fix: prevent HistoryItemContent orphaning during duplicate resolution

### LINKED ISSUE #1368: Duplicate history insertion can orphan HistoryItemContent rows
## Description

Adding a duplicate history item can leave orphaned `HistoryItemContent` rows in the SwiftData store.

## Environment

- Maccy version: 2.6.1 (60)
- macOS version: 26.4 (25E246)
- Reproduced from source on commit `4cd1f979bdf54930519975f951245acc3533f857` (the parent of `a9e452a2ab73a6b3e5446a67ca73200b7bcc2c09`)
- Reproduces against the in-memory test store, so local app data/settings are not required

While narrowing a fix for the unsaved-item path, the same orphaning also showed up when the duplicate item had already been persisted.

## Steps to Reproduce

1. Start with an empty history/store.
2. Add a history item for `"foo"`.
3. Add the same item again through the duplicate-handling path.
4. Inspect `HistoryItemContent` rows.

A focused regression test on commit `4cd1f979bdf54930519975f951245acc3533f857` reproduces the unsaved-item variant consistently.

## Expected Behavior

After duplicate collapse, there should be exactly one surviving `HistoryItem` and exactly one `HistoryItemContent` row attached to it.

## Actual Behavior

After duplicate collapse, there is one surviving `HistoryItem`, but there can be an extra `HistoryItemContent` row with `item == nil`.

## Likely Root Cause

`History.add(_:)` persists the incoming item before duplicate resolution, and the duplicate-removal path can disconnect the incoming contents from any `HistoryItem`.

Project-A-E-G-I-S:
## Summary

Fixes a bug where adding a duplicate clipboard item could leave orphaned `HistoryItemContent` rows in the SwiftData store with `item == nil`, and cascade deletion of the existing item could delete content objects shared with the replacement item.

## Root Cause

In `History.add(_:)` (Maccy/Observables/History.swift:147-163), when a duplicate was detected:

1. The incoming item was already persisted to SwiftData (line 140)
2. Its contents were replaced with the existing item's contents (line 149)
3. The original contents were left orphaned in the store
4. The existing item was cascade-deleted, which could wipe content objects now referenced by the replacement

## Fix

- Delete the incoming item's contents from the store before replacing them (prevents orphaning)
- Detach the existing item's contents (`existingHistoryItem.contents = []`) before cascade deletion (prevents shared content from being wiped)
- Both operations are gated behind `if isModified(item) == nil`, so they only run when contents are actually transferred

## Testing

Added `testDuplicateDoesNotOrphanContents` which verifies:
- Exactly 1 `HistoryItemContent` row exists after duplicate merge
- The surviving content has a valid `item` reference

Closes #1368

### DIFF
--- a/Maccy/Observables/History.swift
+++ b/Maccy/Observables/History.swift
@@ -146,7 +146,11 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
     var removedItemIndex: Int?
     if let existingHistoryItem = findSimilarItem(item) {
       if isModified(item) == nil {
+        for content in item.contents {
+          Storage.shared.context.delete(content)
+        }
         item.contents = existingHistoryItem.contents
+        existingHistoryItem.contents = []
       }
       item.firstCopiedAt = existingHistoryItem.firstCopiedAt
       item.numberOfCopies += existingHistoryItem.numberOfCopies
--- a/MaccyTests/HistoryTests.swift
+++ b/MaccyTests/HistoryTests.swift
@@ -1,5 +1,6 @@
 import XCTest
 import Defaults
+import SwiftData
 @testable import Maccy
 
 @MainActor
@@ -121,6 +122,23 @@ class HistoryTests: XCTestCase {
 
     XCTAssertEqual(history.items, [second])
     XCTAssertEqual(Set(history.items[0].item.contents), Set(firstContents))
+  }
+
+  func testDuplicateDoesNotOrphanContents() {
+    let item = historyItem("foo")
+    history.add(item)
+
+    let descriptor1 = FetchDescriptor<HistoryItemContent>()
+    let contents1 = try? Storage.shared.context.fetch(descriptor1)
+    XCTAssertEqual(contents1?.count, 1)
+
+    let duplicate = historyItem("foo")
+    history.add(duplicate)
+
+    let descriptor2 = FetchDescriptor<HistoryItemContent>()
+    let contents2 = try? Storage.shared.context.fetch(descriptor2)
+    XCTAssertEqual(contents2?.count, 1)
+    XCTAssertNotNil(contents2?.first?.item)
   }
 
   func testAddingItemFromMaccy() {


yonatangross:
This still reproduces on 2.7.0 (60), macOS 26.5.2. Posting measurements in case
they help, since the issue is closed but neither #1369 nor #1387 landed and
`History.swift:149` is unchanged on `main` (b0af655).

**Anyone can reproduce it in about a minute.** Copy the same thing repeatedly
and watch the orphan count climb:

```bash
STORE=~/Library/Containers/org.p0deje.Maccy/Data/Library/Application\ Support/Maccy/Storage.sqlite
orphans() { cp "$STORE" /tmp/m.sqlite; cp "$STORE-wal" /tmp/m.sqlite-wal 2>/dev/null
            sqlite3 /tmp/m.sqlite "SELECT COUNT(*) FROM ZHISTORYITEMCONTENT WHERE ZITEM IS NULL;"; }

orphans                                   # baseline
for i in 1 2 3 4 5; do printf 'same-payload' | pbcopy; sleep 2; orphans; done
```

What I measured here:

| action | orphan rows |
|---|---|
| 5x same text | +1 per re-copy |
| 3x same image | +1 per re-copy |
| 4x alternating A/B | +2 per cycle |
| 10x *unique* content | +0 (this path is fine) |
| 30s idle | +0 |

Only the duplicate path leaks, which matches the diagnosis in the issue body.

**The cost is the full payload, not just a row.** Re-copying one 2.7 MB image
moved orphan bytes from 0.66 MB to 3.40 MB, so 2.74 MB stranded by a single
re-copy. macOS puts screenshots on the pasteboard as uncompressed TIFF, so
image re-copies are expensive.

Accumulated on this machine over 11 months: **46,574 orphaned rows holding
2.4 GB**, in a store whose live history was 262 rows / 125 MB. After removing
only the orphans the file went from 2,572 MB to 136 MB, with all 200 history
items intact and `PRAGMA integrity_check` clean.

**Why it is invisible to users.** The rows have `ZITEM IS NULL`, so no
`HistoryItem` reaches them. They cannot be shown, and no UI action removes
them. "Clear" does clear them, incidentally, because
`delete(model: HistoryItemContent.self, where: #Predicate { $0.item?.pin == nil })`
matches a NULL item too. "Clear all" does not, since it only deletes
`HistoryItem`.

**On the mechanism.** I first assumed the bulk `delete(model:)` calls were
skipping the `.cascade` rule. That is not the case. I tested all three delete
paths against a nullify positive control and every one of them cleans up
correctly. The leak is not a delete at all, it is the reassignment:

```swift
item.contents = existingHistoryItem.contents
```

`item` was already inserted with freshly-read pasteboard contents. Assigning to
a to-many relationship detaches the previous members rather than deleting them,
so the delete rule never fires because nothing was deleted. #1387 handles
exactly this and carries a regression test.

Standalone repro of just the reassignment, no Maccy needed:
https://github.com/yonatangross/clipvault/tree/main/repro

Happy to help verify a fix against a real store if that is useful.

p0deje:
Implemented in https://github.com/p0deje/Maccy/pull/1509

PR #1394, opened 2026-05-04
Cap search query length to prevent crash on huge paste (fixes #1147)

### LINKED ISSUE #1147: Copy pasting inside maccy searching crashes it

- [ ] I have verified that there isn't already an issue reporting the same bug to prevent duplication.
- [ ] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

2.4.1

### macOS Version

15.5

### Maccy Settings

```Shell
{
    "KeyboardShortcuts_delete" = "{\\"carbonKeyCode\\":51,\\"carbonModifiers\\":2048}";
    "KeyboardShortcuts_pin" = "{\\"carbonKeyCode\\":35,\\"carbonModifiers\\":2048}";
    "KeyboardShortcuts_popup" = "{\\"carbonKeyCode\\":9,\\"carbonModifiers\\":768}";
    "NSWindow Frame SUUpdateAlert" = "554 532 620 398 0 0 1728 1079 ";
    "NSWindow Frame com.sindresorhus.Settings.FrameAutosaveName" = "609 488 510 473 0 0 1728 1079 ";
    SUEnableAutomaticChecks = 1;
    SUHasLaunchedBefore = 1;
    SULastCheckTime = "2025-07-26 15:46:56 +0000";
    SUUpdateGroupIdentifier = 3655859351;
        "public.utf8-plain-text",
        "public.html",
        "public.png",
        "public.tiff",
        "public.rtf",
        "public.file-url"
    );
    historySize = 990;
    migrations =     {
        "2024-07-01-version-2" = 1;
    };
    pasteByDefault = 0;
    searchMode = fuzzy;
    showFooter = 1;
    showSearch = 1;
    showTitle = 1;
    windowSize = "[450,800]";
}
```

### Description

Try to invoke maccy, paste anything in the search bar. App crashes instantly 

### Steps to Reproduce

1. Copy any text in the browser.
2. Invoke maccy by shortcut key. 
3. Press Cmd + V to paste anything.
4. Maccy crashes instantly

williamruiz1:
## Summary
When a user accidentally pastes large clipboard contents into Maccy's search field (e.g. cmd+v while the popup is open), regex compilation and Fuse pattern construction can hang or crash the app. Caps the search query at 1,000 chars at the entry of `Search.search` so all search modes are protected.

## Repro
See [#1147 (comment)](https://github.com/p0deje/Maccy/issues/1147#issuecomment-4368372532) — `pbcopy < /usr/share/dict/words`, open Maccy, cmd+v in the search bar.

## Root cause
- `simpleSearch` with `.regularExpression` compiles the entire query as an `NSRegularExpression` pattern — very slow / can OOM on multi-MB input.
- `fuzzySearch` already caps the *target* string at `fuzzySearchLimit = 5_000`, but the *pattern* passed to `fuse.createPattern` is unbounded.

## Fix
One-line guard in `Search.search` that truncates the query to 1,000 chars before dispatching to any backend. No real user is searching clipboard history with a 1,000+ char pattern, so this is safe.

## Test plan
- [x] Built locally on Xcode 26.4.1, ran with a 1MB blob pasted into the search field — no crash, results render in <100ms.
- [x] Existing search behavior unchanged for queries under 1,000 chars.

Fixes #1147

### DIFF
--- a/Maccy/Search.swift
+++ b/Maccy/Search.swift
@@ -35,11 +35,16 @@ class Search {
 
   private let fuse = Fuse(threshold: 0.7) // threshold found by trial-and-error
   private let fuzzySearchLimit = 5_000
+  private let queryLimit = 1_000
 
   func search(string: String, within: [Searchable]) -> [SearchResult] {
     guard !string.isEmpty else {
       return within.map { SearchResult(object: $0) }
     }
+
+    // Cap query length so an accidental cmd+v of huge clipboard contents
+    // into the search field can't blow up regex compilation or fuzzy matching.
+    let string = string.count > queryLimit ? String(string.prefix(queryLimit)) : string
 
     switch Defaults[.searchMode] {
     case .mixed:


Maccy/Search.swift
```diff
@@ -35,12 +35,17 @@ class Search {
...
 
   func search(string: String, within: [Searchable]) -> [SearchResult] {
     guard !string.isEmpty else {
       return within.map { SearchResult(object: $0) }
     }
 
+    // Cap query length so an accidental cmd+v of huge clipboard contents
+    // into the search field can't blow up regex compilation or fuzzy matching.
+    let string = string.count > queryLimit ? String(string.prefix(queryLimit)) : string
+
```
prabhavagrawal7, line 48: Instead of capping the query we should avoid searching blobs alltogether. 
For example if someone copy a text then image (very large) then text, theoretically because of image fuzzy search won't be able to find text below the image.

williamruiz1: Now you're thinking.

PR #1416, opened 2026-06-06
Memory (RAM consumption) fix

jorycz:


goetzc:
@p0deje ^

kmarekspartz:
I think this might still have memory issues. @jorycz  have you compared memory utilization between the trunk and your fix?

PR #1422, opened 2026-06-18
Add clipboard automations

dansimau:
Add an automations pane where users can configure scripts to run on matching clipboard regexps.

* Scripts run via `NSUserUnixTask` from `~/Library/Application Scripts/<bundle-id>/`, which executes them outside the app sandbox with the user's normal privileges (network, filesystem, Homebrew tools)
* Clipboard contents is passed to scripts via stdin, and if the script writes output, that output will be added to the top of the clipboard history and replace current pasteboard contents
* Although scripts are the only automation type right now, this can be expanded in future to allow other (built-in) transformations (e.g. could potentially cover the use-cases in #1413, #1386)

**Screenshot/example**

In the following example, I've written a script to convert plain text GitHub PR URLs to a HTML link with the PR name as the title:

Screenshot 2026-06-18 at 23 31 38

### DIFF
--- a/Maccy/Observables/AutomationProcessor.swift
+++ b/Maccy/Observables/AutomationProcessor.swift
@@ -1,0 +1,96 @@
+import AppKit
+import Defaults
+import Foundation
+import Logging
+
+// Runs user-configured automations against freshly copied text.
+//
+// Registered as an `onNewCopy` hook *after* `History.add`, so by the time it runs the original
+// item is already in history. When one or more automations match, their scripts are chained
+// (each transforms the previous output) and a single new item is produced: it becomes the
+// current clipboard contents and the most-recent history entry, while the original copy is
+// preserved.
+@MainActor
+final class AutomationProcessor {
+  static let shared = AutomationProcessor()
+
+  // Injectable so tests can stub script execution.
+  var scriptRunner: ScriptRunning = ScriptRunner.shared
+
+  private let logger = Logger(label: "org.p0deje.Maccy")
+
+  // Synchronous entry point for the clipboard hook.
+  func process(_ item: HistoryItem) {
+    guard shouldProcess(item), let text = item.text else { return }
+
+    let source = item.application
+    Task { @MainActor in
+      if let result = await buildResult(forText: text, source: source) {
+        History.shared.add(result)
+        Clipboard.shared.copy(result)
+      }
+    }
+  }
+
+  // Skip Maccy's own writes (prevents re-trigger loops) and non-text items.
+  func shouldProcess(_ item: HistoryItem) -> Bool {
+    !item.fromMaccy && (item.text?.isEmpty == false)
+  }
+
+  // Testable core: matches, chains scripts, and builds the resulting item without publishing
+  // it. Returns nil when nothing matched, a script failed, or the output was empty/unchanged.
+  func buildResult(forText text: String, source: String?) async -> HistoryItem? {
+    let matching = Defaults[.automations].filter { $0.isEnabled && $0.matches(text) }
+    guard !matching.isEmpty else { return nil }
+
+    var current = text
+    var isHTML = false
+
+    for automation in matching {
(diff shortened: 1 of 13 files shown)


Maccy/Settings/en.lproj/AutomationsSettings.strings
```diff
@@ -0,0 +1,14 @@
+"Title" = "Automations";
+"Untitled" = "Untitled automation";
+"SelectOrAddHint" = "Select an automation to edit, or add one with +.";
+"Name" = "Name";
+"NamePrompt" = "GitHub PR → rich link";
```
dansimau, line 5: Use a more generic example/placeholder name here

Maccy/Settings/AutomationsSettingsPane.swift
```diff
@@ -0,0 +1,172 @@
...
+  private func editor(_ automation: Binding<Automation>) -> some View {
+    Form {
+      TextField(text: automation.name, prompt: Text("NamePrompt", tableName: "AutomationsSettings")) {
+        Text("Name", tableName: "AutomationsSettings")
+      }
+
+      TextField(text: automation.regexp, prompt: Text("RegexpPrompt", tableName: "AutomationsSettings")) {
+        Text("Regexp", tableName: "AutomationsSettings")
+      }
+      .font(.body.monospaced())
```
dansimau, line 98: Remove monospaced font here

Maccy/Observables/AutomationProcessor.swift
```diff
@@ -0,0 +1,96 @@
...
+    var isHTML = false
+
+    for automation in matching {
+      guard case let .runScript(config) = automation.action, !config.scriptName.isEmpty else {
+        continue
+      }
+
+      do {
+        let output = try await scriptRunner.run(scriptName: config.scriptName, input: current)
+        let cleaned = trimTrailingNewlines(output)
```
dansimau, line 56: This creates a problem if users want to intentionally trailing newlines in their script output.

Probably we should just strip a single trailing newline, which allows people to use `echo` and variants as they normally would.

p0deje:
Thank you for your PR. I'd rather you reach out first to discuss this before opening, since I don't believe Maccy is an app that should have this kind of automation. It's possible to implement this via Shortcuts or maybe a standalone app. Having this functionality is going to open all sorts of feature requests for automation people might want, and it's going to bloat the app. I'd rather keep it lightweight and integrate with 3rd party apps that manipulate the clipboard (e.g., Maccy works well with Pure Paste, which removes formatting from copies).

If you're up for collaborating and making this generally available via better Shortcuts integration, I'd appreciate that.

dansimau:
@p0deje I'm pretty sure this exact functionality cannot be replicated with Shortcuts (a Shortcut can't trigger on a clipboard event).

Will think about alternatives.

Thanks for creating and maintaining Maccy.

p0deje:
There seems to be no way to trigger Shortcut on a new copy, but I wonder if you really want to do that. In your example, is the intention to format all GitHub links or just some of them, which you'd like to paste into Markdown?

Either way, if Maccy migrates from SiriKit to App Intents framework, it should be possible to set up a script that periodically looks at the Maccy history, converts all GitHub links into markdown, and then updates the history in Maccy.

dansimau:
@p0deje in my workflow specifically I want the clipboard automation to be completely automatic (i.e. not a separate trigger or keyboard shortcut).

> In your example, is the intention to format all GitHub links or just some of them, which you'd like to paste into Markdown?

Yes, one thing I've found testing this PR over the past couple of months is that in this implementation the script replaces the pasteboard data with _only_ HTML (meaning when pasting into a plain text context such as a terminal you get raw HTML, which is not a great experience). I am going to look into enhancing this, e.g. by adding **both** the original plain text + HTML to the clipboard.

In any case, it's getting finicky enough that a separate script or app makes sense.

PR #1430, opened 2026-06-22
Improve popup window fitting

### LINKED ISSUE #1427: Keep the popup readable and inside the active screen
## Summary

The popup can feel cramped when a large result set or an open preview needs more vertical space. It can also cross display boundaries when opened near a screen edge, especially with the cursor-based popup position.

Popup window fitting

## Proposed behavior

- Let the popup grow for larger result sets, with a fixed maximum preferred height.
- Give the preview enough minimum height to show content and metadata.
- Clamp popup and slideout frames inside visible screen bounds.
- When the popup opens at the cursor or status item, prefer that screen before falling back to frame center or intersection.

## Localization

No new user-facing strings are needed for this change.

## Notes

The goal is not to change the popup style, only to keep the existing UI readable and fully visible on multi-display setups.

allonli:
## Summary

Improves popup sizing and placement so the existing UI remains readable for long result lists, open previews, and multi-display edge cases.

Closes #1427

## What changed

- Clamps popup frames inside visible screen bounds when opening and resizing.
- Uses the cursor or status item point as the preferred screen when clamping near display edges.
- Clamps slideout frames inside visible screen bounds.
- Raises the preview minimum height so text and metadata have room when the preview is open.
- Allows result lists to grow dynamically, capped at the actual height of the first 9 visible rows.
- Calculates that 9-row cap from each row's real height, including taller image thumbnail rows.

## Screenshot

Popup window fitting

## Localization

No new user-facing strings are introduced.

## Tests

- `swiftlint --strict`
- `swiftlint 0.60.0 lint --strict --reporter xcode`
- `xcodebuild test -scheme Maccy -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO -only-testing:MaccyTests/PreviewBehaviorTests -only-testing:MaccyTests/HistoryItemDecoratorTests/testListItemHeightUsesThumbnailHeightForImages`
- `xcodebuild build -scheme Maccy -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 
-<img width="128px" src="https://maccy.app/img/maccy/Logo.png" alt="Logo" align="left" />
+<img width="128px" src="https://maccy.app/img/maccy/Logo.png" alt="Logo" />
 
 # [Maccy](https://maccy.app)
 


allonli:
Follow-up for the failed Bitrise run: I narrowed the popup height change so normal history lists stay bounded by the saved window height again, while keeping the visible-screen clamping and preview minimum-height fix. The previous dynamic growth made existing UI tests see a different set of visible history items and was too broad for this PR.\n\nVerified locally with:\n\n- `swiftlint --strict`\n- `xcodebuild test -scheme Maccy -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO -only-testing:MaccyTests/PreviewBehaviorTests`\n- `git diff --check`\n\nI also tried the three failing UI tests locally, but this machine cannot expose the app status item to XCUITest, so those attempts fail before the popup opens and are not useful signal for this fix.

allonli:
Updated the popup height behavior based on follow-up testing: the popup can grow dynamically again, but the list content is capped at the actual height of the first 9 visible rows. This uses each row's measured type: normal text rows use the standard item height, while image rows use the resized thumbnail height plus the same vertical padding used by the SwiftUI row.\n\nVerified locally with SwiftLint 0.60.0, current SwiftLint, focused unit tests for the 9-row cap and image-row height calculation, and a Debug build.

PR #1433, opened 2026-06-22
Keep popup window inside visible screen

### LINKED ISSUE #1432: Popup can open off-screen near cursor screen edges
## What happens

When the popup position is set to the cursor, opening Maccy near the bottom edge of a display can place part of the popup outside the visible screen area. In multi-display setups, the popup can also end up clamped onto a neighboring display instead of staying on the display that contains the cursor.

## Expected behavior

The popup should use the screen that contains the cursor as the placement screen, limit the popup size to that screen's visible frame, and clamp the popup origin so the full window stays visible on that same screen.

## Notes

This is a window placement bugfix and does not add or change any user-facing strings, so no localization updates are needed.

allonli:
## Summary

- Keep cursor-based popup placement on the screen that contains the cursor.
- Clamp the cursor popup origin to that screen's visible frame, so the window stays fully visible near bottom and side edges.
- Limit cursor popup size to the cursor screen's visible frame before positioning.
- Add unit coverage for bottom-edge cursor placement, side-edge cursor placement, and visible-frame size clamping.

Fixes #1432.

## Test plan

- `xcodebuild build -project Maccy.xcodeproj -scheme Maccy -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO`
- `xcodebuild test -project Maccy.xcodeproj -scheme Maccy -destination 'platform=macOS' -only-testing:MaccyTests/PopupPositionTests CODE_SIGNING_ALLOWED=NO`
- `swiftlint lint --strict`

## Localization

No user-facing strings were added or changed.

### DIFF
--- a/Maccy.xcodeproj/project.pbxproj
+++ b/Maccy.xcodeproj/project.pbxproj
@@ -66,6 +66,7 @@
 		DA19691F2C3F5F0600258481 /* KeyHandlingView.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA19691E2C3F5F0600258481 /* KeyHandlingView.swift */; };
 		DA1969212C3F6C6800258481 /* HistoryItemAction.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA1969202C3F6C6800258481 /* HistoryItemAction.swift */; };
 		DA1EDE432045B35300479723 /* HistoryDecoratorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DA1EDE422045B35300479723 /* HistoryDecoratorTests.swift */; };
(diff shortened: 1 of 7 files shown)


allonli:
CI note: the latest Bitrise run for this PR is failing in the existing `MaccyUITests` target (`Xcode Test for Mac`), for example `testCopyRTF` cannot find the expected `bar` static text. SwiftLint passes. I also checked the current upstream `master` status for 4cd1f97, and its Bitrise `push`/`pr` contexts are already failing in the same `Xcode Test for Mac` step with `MaccyUITests` failures, so this looks like the current UI test environment rather than a failure specific to the cursor-placement changes here. The focused local validation for this patch is listed in the PR body.

PR #1434, opened 2026-06-25
Fix history fast-scroll freeze

p3ngu1nx:
## Summary

Fix a main-thread freeze that can happen when scrolling quickly through a long history list, especially near the bottom of the popup.

The hang I reproduced was in SwiftUI list layout placement while the popup was also measuring the full scroll content height and image rows were changing size as thumbnails arrived.

## Changes

- Replace the scroll-content `GeometryReader` measurement with deterministic height calculation from row heights.
- Use stable row identity in `MultipleSelectionListView` and remove row-local `.id(...)` churn.
- Give image rows a fixed reserved height so thumbnail generation does not resize rows during fast scrolling.
- Generate thumbnails/previews off the main actor and cancel pending thumbnail work when rows disappear.
- Avoid loading universal-clipboard file URL image data for list thumbnails, which also avoids file-access prompts from row rendering.
- Keep `HistoryItemDecorator` hashing stable by hashing only its identity.

## Validation

- `git diff --check`
- `xcodebuild test -scheme Maccy -only-testing:MaccyTests/HistoryItemDecoratorTests -destination 'platform=macOS,arch=arm64' -derivedDataPath .tmp/DerivedData-FastScrollPR-Decorator CODE_SIGN_STYLE=Manual CODE_SIGN_IDENTITY=- DEVELOPMENT_TEAM=`
- `xcodebuild -scheme Maccy -configuration Release -derivedDataPath .tmp/DerivedData-FastScrollPR-Release -destination 'generic/platform=macOS' CODE_SIGN_STYLE=Manual CODE_SIGN_IDENTITY=- DEVELOPMENT_TEAM= build`
- `codesign --verify --deep --strict --verbose=4 .tmp/DerivedData-FastScrollPR-Release/Build/Products/Release/Maccy.app`
- Manual/runtime validation: fast-scrolled a long history list to the bottom and back; post-scroll sample showed the main thread idle in the event loop rather than stuck in SwiftUI lazy-stack layout.

### DIFF
--- a/Maccy/Clipboard.swift
+++ b/Maccy/Clipboard.swift
@@ -147,12 +147,20 @@ class Clipboard {
 
   @objc
   @MainActor
-  func checkForChangesInPasteboard() { // swiftlint:disable:this cyclomatic_complexity
-    guard pasteboard.changeCount != changeCount else {
+  func checkForChangesInPasteboard() { // swiftlint:disable:this cyclomatic_complexity function_body_length
+    let currentChangeCount = pasteboard.changeCount
+    guard currentChangeCount != changeCount else {
       return
     }
 
-    changeCount = pasteboard.changeCount
+    let pasteboardTypes = Set(pasteboard.types ?? [])
+    // clearContents() increments changeCount, but the following setData/setString does not.
+    // If polling lands between them, wait for data before consuming the change.
+    guard !pasteboardTypes.isEmpty else {
+      return
+    }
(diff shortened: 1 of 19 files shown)


p0deje:
This is a draft without any linked issues that are being fixed here.  I'll close this. Let's discuss implementation before opening a PR if you'd like to improve scrolling performance.

Issue #1240, opened 2025-10-21
The memory usage is a little bit high


kanbykanby:
### Before Submitting Your Bug Report

- [x] I have verified that there isn't already an issue reporting the same bug to prevent duplication.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Maccy Version (see 'About' window)

2.5.1 

### macOS Version

26

### Maccy Settings

```Shell
➜  ~ defaults read org.p0deje.Maccy
{
    "KeyboardShortcuts_delete" = "{\\"carbonModifiers\\":2048,\\"carbonKeyCode\\":51}";
    "KeyboardShortcuts_pin" = "{\\"carbonKeyCode\\":35,\\"carbonModifiers\\":2048}";
    "KeyboardShortcuts_popup" = "{\\"carbonKeyCode\\":9,\\"carbonModifiers\\":768}";
    "LaunchAtLogin__hasMigrated" = 1;
    "NSStatusItem Preferred Position Item-0" = 579;
    "NSStatusItem VisibleCC Item-1" = 0;
    "NSWindow Frame SUUpdateAlert" = "-1240 488 620 398 -1920 0 1920 1049 ";
    "NSWindow Frame com.sindresorhus.Settings.FrameAutosaveName" = "-578 528 510 473 -1920 0 1920 1049 ";
    SUAutomaticallyUpdate = 0;
    SUEnableAutomaticChecks = 1;
    SUHasLaunchedBefore = 1;
    SULastCheckTime = "2025-10-21 06:35:47 +0000";
    SUSendProfileInfo = 0;
    enabledPasteboardTypes =     (
        "public.png",
        "public.tiff",
        "public.html",
        "public.rtf",
        "public.utf8-plain-text"
    );
    historySize = 999;
    ignoredPasteboardTypes =     (
        "net.antelle.keeweb",
        "Pasteboard generator type",
        "de.petermaurer.TransientPasteboardType",
        "com.typeit4me.clipping",
        "com.agilebits.onepassword"
    );
    imageMaxHeight = 50;
    maxMenuItems = 0;
    menuIcon = scissors;
    migrations =     {
        "2024-07-01-version-2" = 1;
    };
    pasteByDefault = 1;
    popupPosition = cursor;
    previewDelay = 2000;
    showApplicationIcons = 0;
    showFooter = 1;
    showRecentCopyInMenuBar = 0;
    showSearch = 1;
[...]

Issue #368, opened 2022-02-04
Feature enhancement: set clear clipboard history for text, images, files based on length of time passed, optionally set by pasteboard type


villeodell:
Hi there! Thanks for this app. I wonder if you would consider this enhancement..

### Feature request / enhancement:

1. Ability to clear the clipboard history based on the amount of time passed (e.g, mins, hours, days, or months). 
    1. This is how Alfred implements something similar: ![Untitled](https://user-images.githubusercontent.com/38110594/152494548-172cc9b4-1f4b-4f52-ae36-e2ed9a795dd4.png)

1. Even better would be the optional additional ability to set this by pasteboard type. That way one could configure it such that clipboard items from one apps would be kept for just 1 day, and for others 1 month.

### Why this would be useful:

-  I think many users might care more about how old things are then how many of them there are. For example, I want to know that the Maccy will have everything from the last 7 days no matter how many items there were in that time period. 
-  It would also give me some assurance that if I might have copied some sensitive info to the clipboard, I could rely on the fact that it would be somewhat short lived -- this benefit would be especially enhanced with #2 above (I could set sensitive apps to 1 hour)

Related:

- #178

nicholas-hibshman1:
Strongly support this feature! 

The way I've decided to work around this for now is enabling the "Clear history on quit" setting in Advanced. This is a decent workaround, but if I restart my computer in the middle of the day, it can be a little unfortunate to have lost what I was working on before the restart.

p0deje:
I think it would be nice to support the time-based limit (e.g. last 7 days), but I don't follow why somebody would need to keep images for 1 month and text for 1 week. It seems like it could be used for some advanced use-case and the clipboard manager might not be the best solution for it.

ymost:
I actually would benefit a lot from different options for different types of items. Here is my use case:
I have a lot of text snippets which I use repeatedly over the course of months and even years, so ideally for text I would like a very long time limit. However, occasionally I copy images, which I typically only need to use again on that same day. When all these images stay in the history for a long time, it takes up a lot of memory and sometimes it even makes the app unbearably  slow, so I resort to deleting them manually.
So I guess what I really need is for the time limit to be based on the size of each item in memory, to get rid of the heavy items more quickly. Basing it on the format (text/image) is probably close enough.

p0deje:
[...]

PR #1436, opened 2026-06-26
Add time-based history retention ("keep for N days")

danielsyauqi:
## Summary

Implements time-based history retention, as discussed in #368. A new optional setting, **Keep for** (`historyKeepDays`), automatically removes unpinned history items whose last copy is older than a configured number of days. Pinned items are never removed.

The setting defaults to `0`, which disables the time-based limit, so existing behaviour is unchanged unless the user opts in.

## Motivation

Per the discussion in #368, the maintainer noted that a time-based limit was desirable but should follow the in-memory storage work (#384). Because this implementation **deletes** expired items from the SwiftData store, it reduces the number of items loaded into memory over long-running sessions rather than adding to the footprint, so it also acts as a partial mitigation for the memory growth reported in #384 and #1240.

To keep the first iteration simple, format-specific retention (separate limits for text vs. images) is intentionally left out; it can be layered on later if there is demand.

## Changes

- `Defaults.Keys.historyKeepDays` (`Int`, default `0` = disabled).
- `History.purgeExpiredItems()` removes unpinned items with `lastCopiedAt` older than the cutoff. It is invoked:
  - on launch (`History.load()`),
  - after each new copy (`History.add()`),
  - when the setting changes, and
  - on an hourly timer, so the limit is enforced while the app stays open and idle (the case the count-based limit cannot cover).
- A **Keep for** control in the Storage settings pane, alongside Size, showing "Disabled" when set to `0`.
- English strings for the new control.
- Unit tests covering removal of expired items, preservation of recent and pinned items, and the disabled (`0`) case.

## Testing

- `xcodebuild ... build` succeeds.
- The new tests pass when run directly.
- The full `MaccyTests` run shows no new failures introduced by this change.

## Notes for reviewers

- The new tests are added to `HistoryTests`, which is currently listed under `skippedTests` in `Maccy.xctestplan`; they therefore will not run in CI as-is. They pass when the class is run directly. Happy to adjust the test plan if preferred.
- Only English (`en.lproj`) strings are included. The other localizations fall back to English until translated.

### DIFF
--- a/Maccy/Extensions/Defaults.Keys+Names.swift
+++ b/Maccy/Extensions/Defaults.Keys+Names.swift
@@ -18,6 +18,7 @@ extension Defaults.Keys {
     "enabledPasteboardTypes", default: Set(StorageType.all.types)
   )
   static let highlightMatch = Key<HighlightMatch>("highlightMatch", default: .bold)
+  static let historyKeepDays = Key<Int>("historyKeepDays", default: 0)
   static let ignoreAllAppsExceptListed = Key<Bool>("ignoreAllAppsExceptListed", default: false)
   static let ignoreEvents = Key<Bool>("ignoreEvents", default: false)
   static let ignoreOnlyNextEvent = Key<Bool>("ignoreOnlyNextEvent", default: false)
--- a/Maccy/Observables/History.swift
+++ b/Maccy/Observables/History.swift
@@ -59,6 +59,12 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
   @ObservationIgnored
   private var sessionLog: [Int: HistoryItem] = [:]
 
+  // Enforces the time-based retention limit while the app stays open and idle,
+  // covering the case where no new copies arrive to trigger a purge.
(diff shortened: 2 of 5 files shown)


Sublio:
Hi @danielsyauqi. I think there might be a problem with "Keep for" stepper. Let's say I want to keep items for "7 Days" and I start counting up -> first is "1" and all items that are older than 1 day going to be purged by the time stepper changed it's value. So by the time I step up to "keep for 7 days" I've already purged all my items. Maybe we can introduce some delay when we count stepper up/down?

p0deje:
> Hi @danielsyauqi. I think there might be a problem with "Keep for" stepper. Let's say I want to keep items for "7 Days" and I start counting up -> first is "1" and all items that are older than 1 day going to be purged by the time stepper changed it's value. So by the time I step up to "keep for 7 days" I've already purged all my items. Maybe we can introduce some delay when we count stepper up/down?

I think it would make sense to follow the same approach we use for size - changing it doesn't purge history until the new copy arrives.

Issue #1259, opened 2025-11-18
Introduce a way to paste image from the clipboard directly into finder


andrefrsousa:
### Before Submitting Your Feature Request

- [ ] Check that there isn't already a similar feature request to avoid creating a duplicate.
- [ ] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Problem

It would be great if Maccy could support pasting images from the clipboard directly into Finder. Currently, Maccy focuses on text snippets, but many users copy images (screenshots, graphics, etc.) and would benefit from a seamless workflow where an image copied to the clipboard can be quickly pasted into finder if CMD+V is used.

Issue #1245, opened 2025-10-29
Save image to file


jhult:
### Before Submitting Your Feature Request

- [x] Check that there isn't already a similar feature request to avoid creating a duplicate.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Problem

Sometimes I copy an image to my clipboard but then want to save that image to a file.

### Solution

Allow saving an image to a file. Perhaps this could be an option by right clicking on the file or a keyboard shortcut (when the image is selected).

twknab:
What about copying it again and just pasting to your desired location? E.g, desktop?

jhult:
Can I paste directly to the macOS desktop? If so, I'm unsure how. My typical use case is using Google Voice in the browser which doesn't support pasting images, only uploading via a custom file picker.

[This workaround](https://github.com/p0deje/Maccy/issues/761#issuecomment-2395585068) gets me close. It would be nice if this could be built-in though. Also, I have to ensure the image is the active item on the clipboard.

Issue #761, opened 2024-04-10
Drag & Drop support (for images)


nevara:
### Before Submitting Your Feature Request

- [X] Check that there isn't already a similar feature request to avoid creating a duplicate.
- [X] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Problem

It would be great to have drag & drop support for images to pull them into websites or other apps for example.

p0deje:
It's not currently possible to implement it due to the limitations of NSMenu. I'll keep that in mind whenever I start working on re-writing Maccy from NSMenu.

gibfahn:
@p0deje is this something that would be more doable in the 2.0 version? Not sure whether that still uses NSMenu.

p0deje:
@gibfahn Yes, it would be possible in 2.0 since it doesn't use NSMenu.

dtonda8:
It doesn't seem to work on 2.0 (at least on my end). Can anyone confirm that it is working on? If so, I will look into my installation.

p0deje:
It's not implemented yet.

gibfahn:
May or may not be useful for anyone else, but my workaround for this is [this zsh function](https://github.com/gibfahn/dot/blob/50d681657b1b02f25bd1ca40fa49c4378505262e/dotfiles/.config/zsh/deferred/deferred.zsh#L234-L249):

```zsh
# Save the current clipboard to a png file, copy the path to the clipboard, and show the file
# in Finder. This allows you to easily drag the file, or paste it with a ⇧⌘g in a file
# picker window.
clipboard_image_save() {
  local screenshot_path
  screenshot_path=$HOME/tmp/screen_shot_recording/screenshot_$(date "+%Y-%m-%d_%H-%M-%S").png
  osascript -e "set png_data to the clipboard as «class PNGf»
	set the_file to open for access POSIX path of (POSIX file \"$screenshot_path\") with write permission
	write png_data to the_file
	close access the_file"

  echo "Created $screenshot_path"
  pbcopy <<<$screenshot_path
  # Reveal in Finder.
  open -R $screenshot_path
}
```

LoganDark:
Also just thought of this. Would be nice :)

clovelt:
This would be amazing! No apps in the market do this atm. Clipboard2File works, but only on browsers and not too well.

PR #1438, opened 2026-06-30
Allow dragging an image item out to a file

maloqab:
## What

Adds drag-and-drop support for image items: you can now drag an image out of Maccy's history into Finder (or any destination that accepts files), and it is written to disk as a PNG.

This addresses #761 (and is the practical answer to the recurring "save/paste image to a file/folder" requests, e.g. #1245 / the dupe #1259) — Maccy stores images but previously had no way to get a copied image back out as a file.

## How

- `HistoryItem.pngImageData` — a PNG-encoded copy of the item's image, normalizing TIFF/JPEG/HEIC sources to PNG so the dropped file is always valid.
- `HistoryItemDecorator.imageFileItemProvider()` — builds an `NSItemProvider` using a **file representation**. This is the sandbox-friendly mechanism: the system grants write access to the drop destination, so **no new entitlements are required**. The PNG is written to a unique temp directory per drag.
- `HistoryItemView` — attaches `.onDrag` only when the item actually contains an image (`item.hasImage`). Text and file items are completely unaffected.

No UI strings were added (drag has no label), so there are no localization changes.

## Testing

- `xcodebuild -scheme Maccy` builds cleanly (Debug, macOS 14 deployment target).
- Verified the drag payload programmatically by exercising the same API Finder uses on drop (`loadFileRepresentation(forTypeIdentifier: "public.png")`): it yields an `Image.png` file whose contents have valid PNG signature bytes.
- I have **not** yet done a live drag-into-Finder test on a running build — would appreciate a maintainer sanity-checking the actual gesture, or I can follow up with a confirmation.

## Notes

- Dropped files are named after the item's title (falling back to `Image.png`). Happy to adjust the naming scheme (e.g. include dimensions or a timestamp) if preferred.

### DIFF
--- a/Maccy/Models/HistoryItem.swift
+++ b/Maccy/Models/HistoryItem.swift
@@ -165,6 +165,18 @@ class HistoryItem {
     return NSImage(data: data)
   }
 
+  // PNG-encoded copy of the item's image, used when dragging it out to a file.
+  // Normalizes TIFF/JPEG/HEIC sources to PNG so the dropped file is always valid.
+  var pngImageData: Data? {
+    guard let image = image,
+          let tiff = image.tiffRepresentation,
+          let bitmap = NSBitmapImageRep(data: tiff) else {
+      return nil
+    }
+
+    return bitmap.representation(using: .png, properties: [:])
+  }
+
   var rtfData: Data? { contentData([.rtf]) }
   var rtf: NSAttributedString? {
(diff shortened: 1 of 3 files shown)


maloqab:
Withdrawing this — after more thought I'm going a different (non-Maccy) route for my use case. Apologies for the noise, and thanks for the great app!

PR #1441, opened 2026-07-01
Minor performance fixes

weisJ:
- Cache decoded images in `HistoryItem`
- Lazily generate and cache preview text for `HistoryItemDecorator`
- Add missing call to `cleanup` when removing duplicate items.
- Avoid fetching history from storage where possible.

### DIFF
--- a/Maccy/Extensions/Defaults.Keys+Names.swift
+++ b/Maccy/Extensions/Defaults.Keys+Names.swift
@@ -3,7 +3,7 @@ import Defaults
 
 struct StorageType {
   static let files = StorageType(types: [.fileURL])
-  static let images = StorageType(types: [.png, .tiff])
+  static let images = StorageType(types: [.png, .tiff, .jpeg, .heic])
   static let text = StorageType(types: [.html, .rtf, .string])
   static let all = StorageType(types: files.types + images.types + text.types)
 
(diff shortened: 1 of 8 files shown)


Maccy/Extensions/Defaults.Keys+Names.swift
```diff
@@ -3,7 +3,7 @@ import Defaults
 
 struct StorageType {
   static let files = StorageType(types: [.fileURL])
-  static let images = StorageType(types: [.png, .tiff])
+  static let images = StorageType(types: [.png, .tiff, .jpeg, .heic])
```
p0deje, line 6: We'll need a migration to add these to `enabledPasteboardTypes` - right now I have only png/tiff:


```
defaults read org.p0deje.Maccy enabledPasteboardTypes
(
    "public.tiff",
    "public.file-url",
    "public.rtf",
    "public.html",
    "public.utf8-plain-text",
    "public.png"
)
```

weisJ: Good point. Will add a migration.

weisJ: Done

weisJ:
Latest changes: Render long text using `NSTextView`. It easily handles extremely long text without UI lag.

Maccy/Extensions/Defaults.Keys+Names.swift
p0deje: The current migration adds jpeg/heic even if user disabled images completely. I think you should be checking and not adding new types in this case.

weisJ: Fixed.

p0deje (approved):

PR #1442, opened 2026-07-01
Add button to copy OCR text from image

weisJ:
Button on the left.
image

After clicking it.
image


MaccyTests/ClipboardTests.swift
```diff
@@ -222,6 +222,12 @@ class ClipboardTests: XCTestCase {
     XCTAssertEqual(pasteboard.string(forType: .source), "com.foo.bar")
   }
 
+  @MainActor
+  func testCopyString() {
+    clipboard.copyInMaccy("foo")
+    XCTAssertEqual(pasteboard.string(forType: .string), "foo")
```
p0deje, line 228: Might be worth checking the source is Maccy as well

p0deje (approved):
This is great. I remember talking about making a new kind of selection (e.g. Shift+Enter) on images that would instead copy its text, but I was hesitant to add a new shortcut - this is much clearer.

I see there are localization conflicts; feel free to merge once resolved.

Issue #1294, opened 2025-12-23
High ram usage


muhammedyousif:
### Before Submitting Your Bug Report

- [x] I have verified that there isn't already an issue reporting the same bug to prevent duplication.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Maccy Version (see 'About' window)

2.6.1

### macOS Version

Tahoe 26.1

### Maccy Settings

```Shell
{
    "KeyboardShortcuts_delete" = "{\\"carbonKeyCode\\":51,\\"carbonModifiers\\":2048}";
    "KeyboardShortcuts_pin" = "{\\"carbonKeyCode\\":35,\\"carbonModifiers\\":2048}";
    "KeyboardShortcuts_popup" = "{\\"carbonKeyCode\\":9,\\"carbonModifiers\\":2048}";
    "NSStatusItem VisibleCC Item-1" = 0;
    "NSWindow Frame com.sindresorhus.Settings.FrameAutosaveName" = "1025 777 710 441 0 0 2560 1409 ";
    SUEnableAutomaticChecks = 1;
    SUHasLaunchedBefore = 1;
    SULastCheckTime = "2025-12-22 18:07:48 +0000";
    SUUpdateGroupIdentifier = 1845891376;
    enabledPasteboardTypes =     (
        "public.utf8-plain-text",
        "public.tiff",
        "public.file-url",
        "public.html",
        "public.rtf",
        "public.png"
    );
    historySize = 250;
    migrations =     {
        "2024-07-01-version-2" = 1;
    };
    pasteByDefault = 1;
    removeFormattingByDefault = 1;
    searchMode = fuzzy;
    showFooter = 1;
    showSearch = 1;
    showTitle = 1;
    suppressClearAlert = 0;
    windowSize = "[450,800]";
}
```

### Description

<img width="1106" height="31" alt="Image" src="https://github.com/user-attachments/assets/4289abf2-9644-4fdc-897d-eb035e63db0b" />

Ram usage is almost 300 mb and real memory is 128 mb

hkyuei:
<img width="1660" height="86" alt="Image" src="https://github.com/user-attachments/assets/926b1d1d-3613-4408-9681-b20df3ae2359" />
!!!

hkyuei:
Memory Leak Analysis Report

  Root Cause: Recursive withObservationTracking Pattern

[...]

Issue #1129, opened 2025-07-15
It occupies a large amount of disk space，also memory


tianxiayu007:
### Before Submitting Your Bug Report

- [ ] I have verified that there isn't already an issue reporting the same bug to prevent duplication.
- [ ] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Maccy Version (see 'About' window)

2.3.0(54)

### macOS Version

15.5

### Maccy Settings

```Shell
{
    "KeyboardShortcuts_delete" = "{\\"carbonKeyCode\\":51,\\"carbonModifiers\\":2048}";
    "KeyboardShortcuts_pin" = "{\\"carbonKeyCode\\":35,\\"carbonModifiers\\":2048}";
    "KeyboardShortcuts_popup" = "{\\"carbonKeyCode\\":8,\\"carbonModifiers\\":768}";
    "NSStatusItem Visible Item-1" = 0;
    "NSWindow Frame com.sindresorhus.Settings.FrameAutosaveName" = "603 451 510 232 0 0 1920 1055 ";
    SUEnableAutomaticChecks = 0;
    SUHasLaunchedBefore = 1;
    SULastCheckTime = "2025-05-29 07:46:22 +0000";
    SUUpdateGroupIdentifier = 3670883842;
    clearOnQuit = 0;
    enabledPasteboardTypes =     (
        "public.file-url",
        "public.tiff",
        "public.png",
        "public.html",
        "public.rtf",
        "public.utf8-plain-text"
    );
    highlightMatch = color;
    historySize = 50;
    menuIcon = clipboard;
    migrations =     {
        "2024-07-01-version-2" = 1;
    };
    pasteByDefault = 1;
    popupPosition = cursor;
    removeFormattingByDefault = 1;
    showApplicationIcons = 0;
    showFooter = 1;
    showRecentCopyInMenuBar = 0;
    showSearch = 1;
    showTitle = 1;
    suppressClearAlert = 1;
    windowPosition = "[0.8411458333333334,0.9498495486459378]";
    windowSize = "[450,800]";
}
```

### Steps to Reproduce

It occupies a large amount of disk space，also memory

<img width="1644" height="414" alt="Image" src="https://github.com/user-attachments/assets/c70474c1-16cb-4a83-a6af-184a1285e00f" />

[...]

Issue #1088, opened 2025-04-28
Very high ram usage (I saw upto 5 gb), even after clearing all the history.


SaurabhSinghAivot:
### Before Submitting Your Bug Report

- [x] I have verified that there isn't already an issue reporting the same bug to prevent duplication.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Maccy Version (see 'About' window)

2.3.0

### macOS Version

15.4.1

### Maccy Settings

```Shell

```

p0deje:
Duplicate of #1077

PR #1446, opened 2026-07-08
fix: reduce ram usage

sesky4:
## Summary

Hi, I found Maccy uses ~2.4 GB of memory at startup on my machine with a large clipboard history (hundreds of items with images/RTF blobs). Others have reported the same issue (#1240 #1294 #1129 #1088). This PR reduces startup memory.

## Issues

1. **Unbounded fetch** — `History.load()` uses a `FetchDescriptor<HistoryItem>()` with no `fetchLimit`, eagerly loading every row and all `contents` relationships into memory.
2. **Inline blob storage** — `HistoryItemContent.value` stores image/RTF blobs inline in SQLite, so every fetch hydrates all of them into `__DataStorage._bytes`.
3. **Full-table fetch on every copy** — `findSimilarItem()` does another unbounded `FetchDescriptor<HistoryItem>()` per clipboard change for dedup.
4. **Eager thumbnail generation** — `ensureThumbnailImage()` decodes full-resolution `NSImage` for every row during view construction, even when the popup isn't visible.

## Changes

- `HistoryItemContent.value`: add `@Attribute(.externalStorage)` so blobs are stored out-of-line and loaded lazily.
- `History.load()`: separate pinned (unlimited) and unpinned (`fetchLimit = min(size, 50)`) fetches, each sorted by the user's sort order.
- `findSimilarItem()`: search the in-memory `all` array instead of doing a full-table fetch.
- `HistoryItemView`: gate `ensureThumbnailImage()` on `scenePhase == .active`.

## Testing

macOS 26.5.1, 1,140 items in DB (SQLite file ~2.8 GB).

| Metric | Before | After |
|---|---|---|
| Physical footprint | 2.4 GB | **59 MB** |
| RSS | ~2.4 GB | **116 MB** |

Filling 999 slots with fresh text copies kept memory stable at ~59 MB. Pins, search, delete, paste stack all work.

## Backward compatibility

`@Attribute(.externalStorage)` triggers a transparent Core Data migration on first launch. Existing data is preserved — no data loss.

### DIFF
--- a/Maccy/ApplicationImageCache.swift
+++ b/Maccy/ApplicationImageCache.swift
@@ -1,8 +1,6 @@
 class ApplicationImageCache {
   static let shared = ApplicationImageCache()
 
-  private let universalClipboardIdentifier: String =
-  "com.apple.finder.Open-iCloudDrive"
   private let fallback = ApplicationImage(bundleIdentifier: nil)
   private var cache: [String: ApplicationImage] = [:]
 
@@ -22,10 +20,6 @@ class ApplicationImageCache {
   }
 
   private func bundleIdentifier(for item: HistoryItem) -> String? {
-    if item.universalClipboard {
-      return universalClipboardIdentifier
-    }
-
     if let bundleIdentifier = item.application {
       return bundleIdentifier
     }
--- a/Maccy/Clipboard.swift
+++ b/Maccy/Clipboard.swift
@@ -1,6 +1,7 @@
 import AppKit
 import Defaults
 import Sauce
+import SwiftData
 
 class Clipboard {
   static let shared = Clipboard()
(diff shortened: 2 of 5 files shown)


weisJ:
> 1. **Unbounded fetch** — `History.load()` uses a `FetchDescriptor<HistoryItem>()` with no `fetchLimit`, eagerly loading every row and all `contents` relationships into memory

Your fix just effectively sets the history limit to 50. How are you handling if someone wants to search for content older than the last 50 entries?

> 2. **Inline blob storage** — `HistoryItemContent.value` stores image/RTF blobs inline in SQLite, so every fetch hydrates all of them into `__DataStorage._bytes`.

Do you have data that `Attribute(.externalStorage)` on its own reduces the memory footprint? Where do you have the information from that it will not load the data into memory?

> 3. **Full-table fetch on every copy** — `findSimilarItem()` does another unbounded `FetchDescriptor<HistoryItem>()` per clipboard change for dedup.

Already handled in #1441. Note that while this improves performance when copying I found to real effect of it on memory usage.

> 4. **Eager thumbnail generation** — `ensureThumbnailImage()` decodes full-resolution `NSImage` for every row during view construction, even when the popup isn't visible.

Thumbnails are generated as soon as an item is displayed. Not on view creation.

PR #1447, opened 2026-07-09
Custom pin reordering through drag and drop and new item editor

weisJ:
Allows reordering pins through drag and drop inside the history list.

https://github.com/user-attachments/assets/f1d7ffde-c9e9-4a10-b6fd-8f87b2305d5e

By default pins are added to the end of the list of pines. Note that with this change pins are no longer sorted automatically by last used etc as the rest of the history. An option would be to have a separate setting for how pins are sorted with a `Manual` mode, in which case drag and drop is enabled (and disabled for all other modes).

This is already compatible with the multi-selection mechanism which is currently still disabled.

Add an edit button to the toolbar which opens a modal dialog for editing items. This replaces the editing mechanism from the settings and removes `PinsSettingsPane`.

image

The pin section is only available for pinned items, but every item can be edited.

### DIFF
--- a/Maccy/Extensions/Defaults.Keys+Names.swift
+++ b/Maccy/Extensions/Defaults.Keys+Names.swift
@@ -53,6 +53,7 @@ extension Defaults.Keys {
   static let migrations = Key<[String: Bool]>("migrations", default: [:], suite: preferencesSuite)
   static let numberOfUsages = Key<Int>("numberOfUsages", default: 0, suite: preferencesSuite)
   static let pasteByDefault = Key<Bool>("pasteByDefault", default: false, suite: preferencesSuite)
+  static let pinOrder = Key<PinOrder>("pinOrder", default: PinOrder(), suite: preferencesSuite)
   static let pinTo = Key<PinsPosition>("pinTo", default: .top, suite: preferencesSuite)
   static let popupPosition = Key<PopupPosition>("popupPosition", default: .cursor, suite: preferencesSuite)
   static let popupScreen = Key<Int>("popupScreen", default: 0, suite: preferencesSuite)
--- a/Maccy/Extensions/Defaults.Keys+Names.swift
+++ b/Maccy/Extensions/Defaults.Keys+Names.swift
@@ -54,6 +54,7 @@ extension Defaults.Keys {
   static let numberOfUsages = Key<Int>("numberOfUsages", default: 0, suite: preferencesSuite)
   static let pasteByDefault = Key<Bool>("pasteByDefault", default: false, suite: preferencesSuite)
   static let pinOrder = Key<PinOrder>("pinOrder", default: PinOrder(), suite: preferencesSuite)
+  static let pinSortBy = Key<Sorter.PinBy>("pinSortBy", default: .lastCopiedAt, suite: preferencesSuite)
   static let pinTo = Key<PinsPosition>("pinTo", default: .top, suite: preferencesSuite)
   static let popupPosition = Key<PopupPosition>("popupPosition", default: .cursor, suite: preferencesSuite)
   static let popupScreen = Key<Int>("popupScreen", default: 0, suite: preferencesSuite)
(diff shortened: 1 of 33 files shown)


weisJ:
Added setting to change the sorting order for pins

image
image

If custom is currently selected and the user selects another sorting mechanism a warning is displayed that this will overwrite the custom sorting (if the new sorting order is different from the custom one).

weisJ:
Cleaned up the implementation and commit history.

Maccy/Extensions/Defaults.Keys+Names.swift
```diff
@@ -52,6 +52,8 @@ extension Defaults.Keys {
   static let migrations = Key<[String: Bool]>("migrations", default: [:])
   static let numberOfUsages = Key<Int>("numberOfUsages", default: 0)
   static let pasteByDefault = Key<Bool>("pasteByDefault", default: false)
+  static let pinOrder = Key<PinOrder>("pinOrder", default: PinOrder())
+  static let pinSortBy = Key<Sorter.PinBy>("pinSortBy", default: .lastCopiedAt)
```
p0deje: I think it's fair to drop it completely - it's very easy to re-order pins now and we can remove the logic that preserved pins order after it's copied.

Maccy/Observables/PinManager.swift
```diff
@@ -0,0 +1,188 @@
...
+    order.reconcile(with: pinnedItems.compactMap(\.item.pin))
+    Defaults[.pinOrder] = order
+  }
+
+  private func sortPinnedItems() {
+    pinnedItems = sorter.sortPins(pinnedItems, key: \.item)
+  }
+}
+
+extension Array where Element == HistoryItemDecorator {
```
p0deje: Can we move it to Extensions folder?

Maccy/Views/Info.plist
p0deje, line 1: Is this file needed?

p0deje (commented):
I haven't looked through all the code yet, but why do we need Legacy* views?

weisJ:
The required api for the smoothest experience is only available on macOS 26. Earlier versions need a workaround.

Maccy/Views/ReorderableContainer.swift
```diff
@@ -0,0 +1,219 @@
...
+        } preview: {
+          dragContext.dragPreview(item)
+        }
+        .dragConfiguration(
+          DragConfiguration(
+            operationsWithinApp: .init(allowCopy: false, allowMove: true),
+            operationsOutsideApp: .init(allowCopy: false, allowMove: false)
+          )
+        )
+        .onDragSessionUpdated { session in
```
weisJ, line 122: This method is only available on macOS 26. It is used to reliably detect when a drag operation has ended, even if that is outside the the window.

For earlier versions the legacy implementation works around this by installing a global monitor to listen for mouse up, which seems to work but really is not the ideal solution, which is why I didn't want to use it for every macOS version (the drag preview is also not as nice for the legacy implementation).

Maccy/Views/Info.plist
weisJ: No, that was an accidental leftover. I removed it.

Maccy/Observables/PinManager.swift
weisJ: Moved

Maccy/Extensions/Defaults.Keys+Names.swift
weisJ: Sure. Removed it.

weisJ:
Unsure why `testIgnoreAllApplicationsExcept` and `testIgnoreApplication` are failing. They pass for me locally.

Maccy/Views/ReorderableContainer.swift
p0deje: Worth adding a comment explaining the macOS 26 and legacy implementation differences.

p0deje (commented):
This works great in my testing, but I'd like to address a few things before the merge:

1. Images should support editing. Even though the title/content should not be editable, it should still be possible to change the pin key.
2. Edit item modal should probably become the first responder. Right now, if you press Enter during editing, it selects the item in Maccy list rather than saving the edited contents. Likewise, pressing Escape closes Maccy instead of just closing the edit modal.
3. When editing an item with rich content, it's removed even if only the pin key is changed. It would be better if formatting were preserved unless the user actually changes the contents.
4. When you pin the first item and then unpin it, it doesn't show the CMD+1 shortcut on it anymore.
5. Let's add accessibility labels for the edit button and controls in the spirit of #1448.
6. Let's use the `square.and.pencil` icon instead - it's quite hard to click the current `pencil`.

PR #1448, opened 2026-07-09
VoiceOver accessibility improvements for the clipboard history panel

joelnicolastorello-lgtm:
# VoiceOver accessibility improvements for the clipboard history panel

## A bit about me

I'm a junior developer taking some of my first real steps into open source — and into
accessibility work specifically. I lost my sight in 2020, and VoiceOver has been how
I've done everything on a Mac since, including, now, contributing code. This is one of
the first accessibility PRs I've ever attempted on GitHub, so please be gentle 🙂

## Why this PR

Maccy is genuinely great, and it's an app I use constantly. But using it with VoiceOver
today ranges from "confusing" to "basically not possible" in a few key spots: there's no
way to tell which history item is selected, several buttons just announce as generic
system icon names with no indication of what they actually do, and arrow-key navigation
through your clipboard history is completely silent.

I'll be honest about the scope of who this matters to: this PR is essentially invisible
to 99% of the people who'll ever look at this repo. But for the 100% of people who
navigate macOS with a screen reader, it's the difference between being able to use Maccy
at all, or not.

## What this changes

- **History rows are now real accessible elements**, not just painted highlights. Each
  row exposes its full content (title, source app, pinned status), a proper button
  trait, and two actions — Pin/Unpin and Delete — reachable directly from VoiceOver.
- **Arrow-key navigation now announces the selected item** (title, app it was copied
  from, and whether it's pinned). Before this, moving through your history with the
  keyboard produced no feedback at all.
- **The quick-action toolbar (Pin, Delete, copy extracted text, remove from paste
  stack) now has real labels.** These were announcing as raw system icon names (e.g.
  the pin icon just read as "pin marker") instead of saying what the button does —
  something I only caught by actually using it live with VoiceOver, not by reading
  the code.
- **The status bar icon now has a label** ("Maccy — clipboard history"), so it's
  identifiable when scanning the menu bar instead of being an anonymous icon.
- **Search field**: the "clear search" button now has a label, and the decorative
  magnifying glass icon is hidden from the accessibility tree instead of being read
  as clutter.
- **The preview toggle button** now announces "Show preview" / "Hide preview"
  depending on its current state, instead of nothing.
- **The footer's Clear / Clear all buttons** are now properly hidden from
  accessibility when visually hidden — previously both existed in the accessibility
  tree simultaneously (one invisible), which meant a screen reader user could
  accidentally trigger "Clear all" while intending to trigger "Clear."
- **8 settings controls** (sliders, pickers, and keyboard shortcut recorders) that
  had no accessible label — VoiceOver would announce just the raw value with zero
  context about what setting it belonged to.
- **Spanish translations** for every new accessibility string, alongside the English
  originals.

## How I tested it

I didn't just read the source and guess. I built a debug version of the app and
tested every one of these changes live, with real VoiceOver, on my own Mac — copying
text, navigating history, pinning items, opening Settings, all of it. A few of the
bugs above (especially the toolbar icon-labeling one) only became obvious once I was
actually navigating with a screen reader instead of reading Swift — the kind of thing
that's genuinely hard to catch from code review alone.

## What's *not* in this PR

There's an open question I looked into but couldn't responsibly resolve here: whether
VoiceOver's cursor reliably reaches the history popup at all when it opens (it's a
non-activating panel, which is exactly the reason it doesn't steal focus from whatever
app you're about to paste into). I found and reverted an attempted fix for this after
realizing it would silently break pasting into the wrong app — not something I was
willing to ship. My own live test suggests VoiceOver *does* reach the panel in
practice, but I'd rather flag this as unresolved than claim more confidence than I
actually have.

## Being upfront

I know a PR like this, from someone this new both to the project and to accessibility
work in general, most likely won't get merged exactly as-is — maybe some of these
choices don't match how you'd want to approach it, maybe the scope should be split up,
maybe there's context about the codebase I'm missing. All of that's completely fine.
I mostly just wanted Maccy — an app I already rely on — to be a little more usable for
people like me.

Thank you, genuinely, for taking the time to read this far. 🙏


Maccy/Observables/NavigationManager.swift
```diff
@@ -32,6 +32,23 @@ class NavigationManager { // swiftlint:disable:this type_body_length
...
+        var parts = [item.hasImage ? NSLocalizedString("history_item_image_accessibility_generic", comment: "") : item.title]
+        if let application = item.application {
+          parts.append(application)
+        }
+        if item.isPinned {
+          parts.append(NSLocalizedString("history_item_pinned_accessibility_value", comment: ""))
+        }
+        announceForAccessibility(parts.joined(separator: ", "))
+      }
+
```
weisJ, line 48: Can you make announceForAccessibility take a closure that produces a string? this way we can avoid this computation if voiceover is not enabled.

Maccy/Views/ConfirmationView.swift
```diff
@@ -7,6 +7,14 @@ struct ConfirmationView<Content: View>: View {
...
     if let confirmation = item.confirmation, let suppressConfirmation = item.suppressConfirmation {
       content()
+        .accessibilityAddTraits(.isButton)
+        .accessibilityAction {
+          if suppressConfirmation.wrappedValue {
+            item.action()
+          } else {
+            item.showConfirmation = true
+          }
+        }
```
weisJ: This pattern of marking something as a button and then having `onTapGesture` an `accessibilityAction` with the same closure is repeated a few times. I think it is worth while to extract it into a separate extension method (maybe named `buttonAction`?).

Maccy/Views/FooterView.swift
```diff
@@ -27,8 +27,10 @@ struct FooterView: View {
       ZStack {
         FooterItemView(item: footer.items[0])
           .opacity(clearOpacity)
+          .accessibilityHidden(clearOpacity == 0)
```
weisJ: This pattern of `opacity` joined with `accessibilityHidden` also lends itself to extraction into a helper method `invisible` taking a boolean. Of course `clearOpacity` and `clearAllOpacity` would need to be converted to booleans `showClear` and `showClearAllI`.

Maccy/Views/HeaderView.swift
```diff
@@ -34,9 +34,13 @@ struct HeaderView: View {
           tableName: "PreviewItemView",
           replacementKey: "previewKey"
         )
+        .accessibilityLabel(
+          Text(controller.state.isOpen ? "preview_toggle_hide_accessibility_label" : "preview_toggle_show_accessibility_label")
+        )
```
weisJ: Maybe the better approach here is to modify `KeyboardShortcutHelpModifier` to also set the `accessibilityLabel`. Currently this would result in the label just being `Press {previewKey} to toggle preview.`, but this could be solved by splitting the key `PreviewKey` (in the table `PreviewItemView`) into `OpenPreview` and `ClosePreview` and choosing either for the `key` parameter of `shortcutKeyHelp`.

Changing the modifier would have the nice effect that would also apply to the toolbar buttons.

Maccy/Views/ListItemView.swift
```diff
@@ -114,6 +123,14 @@ struct ListItemView<Title: View, ID: Hashable>: View {
     // The slight opcaity white background is a workaround
     .background(isSelected ? Color.accentColor.opacity(0.8) : .white.opacity(0.001))
     .clipShape(selectionAppearance.rect(cornerRadius: Popup.cornerRadius))
+    .accessibilityElement(children: .ignore)
+    .accessibilityLabel(Text(accessibilityLabel))
+    .accessibilityAddTraits(.isButton)
+    .accessibilityAddTraits(isSelected ? .isSelected : [])
+    .accessibilityValue(selectionIndex.map { Text("\($0 + 1)") } ?? Text(""))
```
weisJ: Let's extract the string `"\(selectionIndex + 1)"` as a computed property, so it can be reused here an above.

Maccy/Views/ToolbarView.swift
```diff
@@ -114,6 +114,9 @@ struct ToolbarView: View {
             Image(systemName: "text.viewfinder")
           }
           .help(Text("CopyExtractedText", tableName: "PreviewItemView"))
+          // Icon-only button relying solely on the SF Symbol's own generic description
+          // ("text.viewfinder") - VoiceOver had no way to know this copies extracted text.
+          .accessibilityLabel(Text("CopyExtractedText", tableName: "PreviewItemView"))
```
weisJ: If we extend `shortcutKeyHelp` to allow specifying `name: nil` then we can also reuse it for the `help` together with `accessibilityLabel` combo.

Maccy/FloatingPanel.swift
```diff
@@ -75,6 +75,20 @@ class FloatingPanel<Content: View>: NSPanel, NSWindowDelegate {
...
+    // is running - but Clipboard.paste() works by posting a synthetic Cmd+V to whatever app
+    // is frontmost at that moment (Clipboard.swift, `sourceApp`/`paste()`), relying entirely
+    // on the previously-active app staying frontmost because this panel never activates.
+    // There is no code anywhere that re-activates that app before pasting, so calling
+    // `NSApp.activate` here would make the synthetic paste land on Maccy itself instead of
+    // the app the user meant to paste into - silently breaking the app's core feature for
+    // exactly the users this fix is meant to help. Fixing this for real needs either (a)
+    // capturing the previously-frontmost app and reactivating it right before `paste()`
+    // fires, with the timing risk that implies, or (b) a different mechanism entirely -
+    // needs a design discussion with the maintainer before writing code, not a one-line fix.
```
weisJ: What exactly is reverted here?

weisJ (commented):
Thank you for stepping up to implement this. I left few comments which should make the implementation cleaner and also make it easier to keep new code accessible.

joelnicolastorello-lgtm:
Thank you so much for the detailed review — this is genuinely useful feedback and I appreciate you taking the time. I've pushed a commit addressing all six points:

1. **`announceForAccessibility`** now takes the announcement as an `@autoclosure`, so the string-building work is skipped entirely when VoiceOver isn't running.
2. **Extracted `buttonAction`** (in `View+ButtonAction.swift`): marks a view as a button and wires the same closure to both `onTapGesture` and `accessibilityAction`. Used in `ConfirmationView` and `ListItemView` in place of the duplicated pattern.
3. **Extracted `invisible`** (in `View+Invisible.swift`): combines `opacity` + `accessibilityHidden`. Converted `FooterView`'s `clearOpacity`/`clearAllOpacity` to booleans (`showClear`/`showClearAll`) as you suggested — turns out the crossfade wasn't actually animated to begin with, so this was a clean swap.
4. **`KeyboardShortcutHelpModifier`** now also sets `accessibilityLabel` from the same resolved string used for `.help()`. Split `PreviewKey` into `OpenPreview`/`ClosePreview` in the `PreviewItemView` table so the label reflects the current toggle state. This replaced my one-off `accessibilityLabel` calls on Pin/Delete/preview-toggle.
5. **Extracted `selectionNumber`** as a computed property, now shared between the visible badge and `accessibilityValue`.
6. **`shortcutKeyHelp`'s `name` is now optional** (defaults to `nil`), so `CopyExtractedText` — which has no keyboard shortcut — goes through the same modifier instead of a separate `.help()` + `.accessibilityLabel()` pair.

On the `FloatingPanel.swift` question: sorry for the confusion — there was nothing to revert in this PR's actual history. I tried `NSApp.activate()` locally while investigating whether VoiceOver reliably reaches the popup, realized it would silently break pasting into the wrong app, and reverted it *before* ever committing that version. The comment was describing that investigation, not undoing anything visible in this diff. I rewrote it to make that clear.

Let me know if any of these choices don't fit how you'd rather see it done — happy to adjust.

weisJ:
> On the `FloatingPanel.swift` question: sorry for the confusion — there was nothing to revert in this PR's actual history. I tried `NSApp.activate()` locally while investigating whether VoiceOver reliably reaches the popup, realized it would silently break pasting into the wrong app, and reverted it _before_ ever committing that version. The comment was describing that investigation, not undoing anything visible in this diff. I rewrote it to make that clear.

You mentioned that voice over seems to reach the popup for you. Is that reliable? If yes I think we can leave the version as is. Otherwise I think we could incorporate a compatibility mode for VoiceOver. I will also test this locally so see whether I can reproduce a possible issue here.

joelnicolastorello-lgtm:
On reliability: so far it's been consistent for me. I've tested opening the popup multiple times in a row, both via the keyboard shortcut and by clicking the menu bar icon, and VoiceOver has reached it every time — I haven't hit a silent/unreachable case yet. I've also just been using Maccy day-to-day since testing this PR (it's genuinely become one of my favorite tools), so this isn't a one-off test, it's real daily use. That said, I don't want to overstate my confidence from a sample size of one person on one machine — really glad to hear you're going to try reproducing it locally too, that's a much better way to settle it than my testing alone.

And separately from the technical question — thank you, genuinely. This is one of the first proposals I've ever put up on GitHub, as a junior developer just starting to find his footing in open source contributions. Getting a review this thoughtful, from someone who clearly cares about getting accessibility right rather than just waving a PR through, means a lot more than I think you'd expect. Thank you for the corrections you suggested earlier too, and please forgive whatever mistakes I made along the way — I'm still learning how all of this works.

Maccy/Views/HistoryItemView.swift
```diff
@@ -37,6 +37,38 @@ struct HistoryItemView: View {
...
+      parts.append(application)
+    }
+    if item.isPinned {
+      parts.append(NSLocalizedString("history_item_pinned_accessibility_value", comment: ""))
+    }
+    if let index = visualIndex {
+      parts.append(String(format: NSLocalizedString("history_item_selected_accessibility_value", comment: ""), index + 1, appState.navigator.selection.count))
+    }
+    return parts.joined(separator: ", ")
+  }
```
p0deje: It would be great to move this to a `HistoryItemDecorator`, so the view doesn't have to work both with `item` and `item.item`.

Maccy/Views/ListItemView.swift
```diff
@@ -40,12 +40,19 @@ struct ListItemView<Title: View, ID: Hashable>: View {
...
+  var accessibilityLabel: String = ""
   @ViewBuilder var title: () -> Title
 
   @Default(.showApplicationIcons) private var showIcons
   @Environment(AppState.self) private var appState
   @Environment(ModifierFlags.self) private var modifierFlags
 
+  // Use the same selection number for the visible badge and accessibility value.
+  private var selectionNumber: String? {
+    selectionIndex.map { "\($0 + 1)" }
```
p0deje, line 53: The `selectionIndex` is an integer and I don't think the `map` function exists on the integers.

Maccy/FloatingPanel.swift
```diff
@@ -80,6 +80,7 @@ class FloatingPanel<Content: View>: NSPanel, NSWindowDelegate {
     let finalHeight = max(min(height, size.height), miniumHeight)
     setContentSize(NSSize(width: finalWidth, height: finalHeight))
     setFrameOrigin(popupPosition.origin(size: frame.size, statusBarButton: statusBarButton))
+
```
p0deje: Please, remove as it's not needed.

Maccy/VoiceOver.swift
```diff
@@ -0,0 +1,19 @@
+import AppKit
+
+/// Posts an announcement when VoiceOver is enabled.
+///
+/// The autoclosure avoids constructing announcements when they cannot be heard.
```
p0deje: I think this comment is not valid anymore.

p0deje (commented):
First of all, let me thank you for taking the time and effort to work on this. This is a non-trivial piece of work, and I don't know how much time you spent on it. I'll do my best to make sure it's not in vain.

I've left a couple of minor comments, but once they are addressed, I'll pull this and give it a short test. I honestly don't know how to test the accessibility aspects, so I'll take your word for it, as you clearly have more experience in this area. From a code perspective, all changes make perfect sense. I just want to make sure nothing breaks with them.

On a side note, it would be great if there were a way to make sure accessibility is not an afterthought. I, for instance, will likely forget to add an accessibility label to a new UI element, so a linter or a shared view helper that propagates a visible text label to an accessibility label would be handy. I'm not sure it exists, and I think this PR should be merged without this, but I wish it were harder to miss these aspects.

Maccy/VoiceOver.swift
weisJ: Done

Maccy/FloatingPanel.swift
weisJ: Removed

Maccy/Views/ListItemView.swift
weisJ: It is an optional integer

Maccy/Views/HistoryItemView.swift
weisJ: Moved to `HistoryItemDecorator`

p0deje (approved):

p0deje:
@joelnicolastorello-lgtm This is now mergd and will be available in the next release. Thank you very much!

PR #1454, opened 2026-07-16
fix: Fix flaky UI tests

Sublio:
Fix flaky UI tests on macOS 26

• Isolate test runs from the developer's personal preferences: reset windowSize and pinTo to defaults in the enable-testing launch path.
• Add history-scroll-view accessibility identifier to the history scroll view so tests can locate and scroll it.
• Make hover() produce real pointer movement (two points) so hover-selection registers under XCUITest — fixes CI-only failures like testCopyWithEnter.
• Add scrollIntoViewIfNeeded and hoverAndClick helpers; scroll rows into view before clicking so clicks don't fall through to the footer.
• Extract closePopupByClickingOutside helper and use it for close-on-outside-click tests.

### DIFF
--- a/Maccy/AppDelegate.swift
+++ b/Maccy/AppDelegate.swift
@@ -31,6 +31,8 @@ class AppDelegate: NSObject, NSApplicationDelegate {
                  userDriver: SPUStandardUserDriver(hostBundle: Bundle.main, delegate: nil),
                  delegate: nil)
       .automaticallyChecksForUpdates = false
+      Defaults.reset(.windowSize)
+      Defaults.reset(.pinTo)
     }
     #endif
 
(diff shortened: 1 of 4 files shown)


Maccy/AppDelegate.swift
```diff
@@ -31,6 +31,8 @@ class AppDelegate: NSObject, NSApplicationDelegate {
                  userDriver: SPUStandardUserDriver(hostBundle: Bundle.main, delegate: nil),
                  delegate: nil)
       .automaticallyChecksForUpdates = false
+      Defaults.reset(.windowSize)
```
p0deje: We should not do that. If you run a UI test, it's going to reset the window size for your existing _production_ version of Maccy. Instead, return a stubbed window when `enable-testing` is set.

Maccy/AppDelegate.swift
```diff
@@ -31,6 +31,8 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
+      Defaults.reset(.pinTo)
```
p0deje: Ditto.

MaccyUITests/MaccyUITests.swift
p0deje, line 1: Please fix formatting (indentation)

p0deje (changes requested):

PR #1456, opened 2026-07-17
Fix popup hotkey re-registration leak

GuangDai:
## Summary

- keep the popup global shortcut registered for the process lifetime
- route complete shortcut presses through the existing popup state machine
- keep only the `flagsChanged` monitor needed for release-to-select behavior

## Root cause

Opening the popup disabled `.popup`, and closing it enabled the shortcut again. In KeyboardShortcuts 2.0.2, that pair unregisters the Carbon hotkey and registers a new one. The Carbon backing allocations accumulated across popup open/close cycles.

Keeping the global registration alive removes that allocation cycle. Toggle, ordinary repeated presses, and release-to-select continue to use the existing popup state machine.

## Known test difference

`testOpenAndSelectThirdItemRepeatedPress` posts multiple synthetic key-down events without intervening key-up events. With the Carbon hotkey left registered, those extra raw events do not reach the local AppKit monitor, so that test remains red. The minimal leak fix is intentionally kept here; compatibility with that synthetic repeat sequence can be handled separately if it reflects required runtime behavior.

## Validation

- ported from the fork's measured fix (`04298ab2`), which observed 783 live Carbon hotkey allocations using 43.5 MB after extended use
- verified against the pinned KeyboardShortcuts 2.0.2 source: `disable` unregisters, `enable` calls `RegisterEventHotKey`, and handled hotkey events return `noErr`
- independently AI-reviewed against current upstream `master`
- `git diff --check` passes
- the branch contains one production-code file and no docs or CI changes

### DIFF
--- a/Maccy/Observables/Popup.swift
+++ b/Maccy/Observables/Popup.swift
@@ -61,8 +61,8 @@ class Popup {
     guard eventsMonitor == nil else { return }
 
     self.eventsMonitor = NSEvent.addLocalMonitorForEvents(
-      matching: [.flagsChanged, .keyDown],
-      handler: handleEvent
+      matching: .flagsChanged,
+      handler: handleFlagsChanged
     )
   }
 
@@ -78,7 +78,6 @@ class Popup {
 
   func reset() {
     state = .toggle
-    KeyboardShortcuts.enable(.popup)
   }
 
   func close() {
@@ -114,52 +113,10 @@ class Popup {
     if isClosed() {
       open(height: height)
       state = .opening
-      KeyboardShortcuts.disable(.popup)  // Handle events via eventsMonitor. Re-enable on popup close
       return
     }
 
-    // Maccy was not opened via shortcut. We assume toggle mode and close it
-    close()
-  }
-
-  private func handleEvent(_ event: NSEvent) -> NSEvent? {
-    switch event.type {
-    case .keyDown:
-      return handleKeyDown(event)
-    case .flagsChanged:
-      return handleFlagsChanged(event)
-    default:
-      return event
-    }
-  }
-
-  private func handleKeyDown(_ event: NSEvent) -> NSEvent? {
-    if isHotKeyCode(Int(event.keyCode)) {
-      if let item = History.shared.pressedShortcutItem {
-        AppState.shared.navigator.select(item: item)
-        Task { @MainActor in
-          AppState.shared.history.select(item)
-        }
-        return nil
-      }
-
-      if state == .opening {
-        state = .cycle
-        // Next 'if' will highlight next item and then return nil
-      }
-
-      if state == .cycle {
-        AppState.shared.navigator.highlightNext(allowCycle: true)
-        return nil
-      }
-
-      if state == .toggle && isHotKeyModifiers(event.modifierFlags) {
-        close()
-        return nil
-      }
-    }
-
-    return event
+    handleRepeatedHotKeyDown()
   }
 
   private func handleFlagsChanged(_ event: NSEvent) -> NSEvent? {
@@ -180,24 +137,30 @@ class Popup {
     return event
   }
 
-  private func isHotKeyCode(_ keyCode: Int) -> Bool {
-    guard let shortcut = KeyboardShortcuts.Name.popup.shortcut else {
-      return false
-    }
-
-    return shortcut.key?.rawValue == keyCode
-  }
-
-  private func isHotKeyModifiers(_ modifiers: NSEvent.ModifierFlags) -> Bool {
-    guard let shortcut = KeyboardShortcuts.Name.popup.shortcut else {
-      return false
-    }
-
-    return modifiers.intersection(.deviceIndependentFlagsMask) ==
-      shortcut.modifiers.intersection(.deviceIndependentFlagsMask)
-  }
-
   private func allModifiersReleased(_ event: NSEvent) -> Bool {
     return event.modifierFlags.isDisjoint(with: .deviceIndependentFlagsMask)
+  }
+
+  private func handleRepeatedHotKeyDown() {
+    if let item = History.shared.pressedShortcutItem {
+      AppState.shared.navigator.select(item: item)
+      Task { @MainActor in
+        AppState.shared.history.select(item)
+      }
+      return
+    }
+
+    if state == .opening {
+      state = .cycle
+    }
+
+    if state == .cycle {
+      AppState.shared.navigator.highlightNext(allowCycle: true)
+      return
+    }
+
+    if state == .toggle {
+      close()
+    }
   }
 }


GuangDai:
While working on the preview path in my fork, I ran into three separate and important preview/UI performance issues. They remain **out of scope for this PR**, but I want to record them as possible follow-ups.

I rechecked the commit history through the [final fork snapshot immediately before the Swift 6 / strict-concurrency migration](https://github.com/GuangDai/Maccy/commit/1f25102bf94bbdf8d0437347aa89e96ecf60e9be). The actual final chains were:

1. **Large text can trigger a SwiftUI/CoreText render and layout storm.** The final pre-migration state bounded preview text with a configurable limit (default 3,000 characters; 0 maps to effectively full text) ([final text-limit implementation](https://github.com/GuangDai/Maccy/commit/36ab3b0dc7f980e0d8e2fb7f81d5bac2a7fb5ea8)). One correction to my earlier wording: the experiment that removed the history-items transition ([8700a5a3](https://github.com/GuangDai/Maccy/commit/8700a5a3a79bdd8e201a5ef5db21a03eabbdeb2a)) was explicitly reverted ([fc60568f](https://github.com/GuangDai/Maccy/commit/fc60568f6a27023147815a845bf725469c01dacd)) because it increased peak stalls. The layout-driving animation that ultimately had to go was the preview window's animated frame-resize path; the fork replaced it with one immediate frame update plus an opacity-only fade ([463feb15](https://github.com/GuangDai/Maccy/commit/463feb1569eefcb94adfa3cfe5daddaf5f38e625)).

2. **High-resolution images should be downsampled for the preview viewport.** The final path used an ImageIO downsampling primitive ([3a5241ac](https://github.com/GuangDai/Maccy/commit/3a5241ac13f1bce2229760e0f6c024d727cd115e)) through an off-main `ImageProcessor` actor and decorator integration ([actor](https://github.com/GuangDai/Maccy/commit/8f006d64a8dc8064de816a859a21f5fd59a4203e), [integration](https://github.com/GuangDai/Maccy/commit/10853086736f77eeb4e6fa7b1bd4fdc024a8c415), [shared processor/cache](https://github.com/GuangDai/Maccy/commit/1672b57dd573b302695594cd5f1c673930c4b490)). Preview decode was capped at an 800-pixel longest side by default ([545d4d46](https://github.com/GuangDai/Maccy/commit/545d4d46e474c1b6a8b9f7717cb7cabbe39c7146)) and then made configurable ([36ab3b0d](https://github.com/GuangDai/Maccy/commit/36ab3b0dc7f980e0d8e2fb7f81d5bac2a7fb5ea8)); 0 means screen-sized rather than decoding a huge native-resolution bitmap.

3. **Rapid mouse movement can make the preview chase the hovered selection and visibly jitter.** The final solution was layered: hover selection stopped driving `scrollTo` / `LazyVStack` layout invalidation ([03f8f67a](https://github.com/GuangDai/Maccy/commit/03f8f67aefa4d315017194cb17c55083571558c6)); selection changes cancelled stale preview decodes ([3599b2fb](https://github.com/GuangDai/Maccy/commit/3599b2fbfba44460c93b71a46ed9b546dc8c7684)); the displayed `previewedItem` was decoupled from the live `leadHistoryItem`, with cancel-on-change retargeting/debounce ([ee2cd34d](https://github.com/GuangDai/Maccy/commit/ee2cd34d5b31cd22f6c646eb6be96b7bb85dbdda)); and preview geometry changed instantly with only a composited opacity fade ([463feb15](https://github.com/GuangDai/Maccy/commit/463feb1569eefcb94adfa3cfe5daddaf5f38e625)). The final default retarget delay was 200 ms and remained configurable down to 0 ([a9ddf3fb](https://github.com/GuangDai/Maccy/commit/a9ddf3fb1a7b84a9d307313802bf5dcb17425858)).

The fork has diverged substantially since these changes landed. These commits document behavior that survived to the final pre-migration snapshot; they are references and evidence, not patches that can be cherry-picked. Each issue would need to be revisited independently and adapted into a small upstream-compatible change.

GuangDai:
CleanShot 2026-07-17 at 19 53 49

If we can deal with these three problems, we can make maccy much smoother like my branch.

GuangDai:
One more candid note after checking the current upstream path: OCR is eager, not search-triggered. Every accepted clipboard entry is passed through `generateTitle()` ([clipboard ingest](https://github.com/p0deje/Maccy/blob/59e781732f8501d2652ef68823226b00a30c48a0/Maccy/Clipboard.swift#L220-L230)); if the item contains an image, that immediately schedules a `.fast` `VNRecognizeTextRequest` ([image branch](https://github.com/p0deje/Maccy/blob/59e781732f8501d2652ef68823226b00a30c48a0/Maccy/Models/HistoryItem.swift#L88-L94), [Vision request](https://github.com/p0deje/Maccy/blob/59e781732f8501d2652ef68823226b00a30c48a0/Maccy/Models/HistoryItem.swift#L216-L229)). This happens before history deduplication, so duplicate image copies can also start OCR, and changing `showSpecialSymbols` calls `generateTitle()` again for every current item ([settings update path](https://github.com/p0deje/Maccy/blob/59e781732f8501d2652ef68823226b00a30c48a0/Maccy/Observables/History.swift#L87-L92)). There is no OCR enable/disable setting or lazy search-time trigger.

Honestly, I do not think OCR earns its cost as an always-on built-in clipboard feature. My expectation—not a measured statistic—is that in well over 90% of image clipboard usage, users will never need to find those images by recognized text. Spending compute and energy on every image copy for a rarely used capability looks like a poor resource trade-off to me. For that reason I [removed the Vision OCR image-title feature from my fork](https://github.com/GuangDai/Maccy/commit/ef95792424cd74a9917d95d2433bed74bf8411ba). If OCR remains upstream, I think it should at least be opt-in and/or lazy rather than running for every copied image.

weisJ:
Do you have any data indicating that hotkey registration causes actual significant memory usage?

> 1. **Large text can trigger a SwiftUI/CoreText render and layout storm.** [...]

In my experience there is no significant lag during search/delete/pin etc. Can you give an example, where this is the case for you. The animation exists very deliberately.

> 2. **High-resolution images should be downsampled for the preview viewport.** [...]

What is the rational behind this proposed change? SwiftUI already handles downsampling images under the hood.

> 3. **Rapid mouse movement can make the preview chase the hovered selection and visibly jitter.** [...]

Again, do you have actual evidence that this is a problem? Usually SwiftUI already handles coalescing layout invalidation quite well.

> One more candid note after checking the current upstream path: [...]
> 
> Honestly, I do not think OCR earns its cost as an always-on built-in clipboard feature. [...]

OCR happens on a background thread. Having OCR run before deduplication is also not a very urgent issue as images in the clipboard are near impossible to be deduplicated anyways. As above, do you have data supporting image OCR taking up significant resources? With the next update the OCR text from images can directly be copied from the UI, which - in my opinion - makes this a valuable feature.
Though, if you provide a PR adding a toggle to the settings, I generally don't see a problem why it couldn't be merged. If you do please make sure to provide translations for any added strings appearing in the UI.

GuangDai:
CleanShot 2026-07-18 at 08 39 18


large text cause freeze. 

large text: you can generate any large text using chatgpt. 

example file:

[large text.txt](https://github.com/user-attachments/files/30144145/large.text.txt)

sample file here

[Maccy.sample.txt](https://github.com/user-attachments/files/30144117/Maccy.sample.txt)

And memory leak. If you used for a long time (10+ hours) and always trigger with shortcuts, then run leak $PID(Maccy's PID). You will find the evidence.

GuangDai:
And my fork.

CleanShot 2026-07-18 at 08 48 30

[Maccy.sample.txt](https://github.com/user-attachments/files/30144169/Maccy.sample.txt)

GuangDai:
Then there's the issue of downsampling. I'm not familiar with Swift features; I just followed my previous development approach and did it directly. So it might be incorrect.

But I implemented multilayer cache. So I store downsampled image in memory, and leave origin image in disk to save memory.

Next, regarding OCR, I prefer that if a feature isn't used but runs frequently, then it's unnecessary. So I removed it. Of course, this is just a personal habit of mine and my friend's; I'm not sure how many people actually use OCR. But I still maintain my opinion that this feature is unnecessary.

GuangDai:
Problem 3 and Problem 1 are directly related; improving Problem 1 will greatly alleviate Problem 3.

However, the behavior will differ from my current fork. In my fork, I made the following modifications based on my personal habits:

1. Original: As long as you trigger Preview, moving the mouse, even briefly skipping an unrelated item, will still trigger Preview.

2. Mine: Triggering Preview and then moving the mouse will only refresh the preview area if it hovers over an item for more than the configured preview time (I set 200ms as default).

GuangDai:
And the memory leak issue ,as you've seen,  the app performs terribly with large text, and I frequently need to use ChatGPT for my studies. Furthermore, verifying a clear memory leak requires a long run, so I'm temporarily unwilling to provide this evidence myself, as it significantly impacts usability.

GuangDai:
[Maccy-dry-run-818f03d-2.6.1-60.zip](https://github.com/user-attachments/files/30144344/Maccy-dry-run-818f03d-2.6.1-60.zip)

my fork app. If you worry about any safety, security issue with this app. (I know there was a fake maccy) You can fork my fork, check it with AI, and run github workflow to build yourself.

weisJ:
I would appreciate more focused replies to my answers.

You are comparing the preview speed from the latest released version of Maccy to your fork, which already includes changes from this repository. Please give a comparison between the current master branch and your version. The current version on master already handles large text much better.

> And memory leak. If you used for a long time (10+ hours) and always trigger with shortcuts, then run leak $PID(Maccy's PID). You will find the evidence.

Did you check that Keyboard shortcut registration is actually at fault?

> However, the behavior will differ from my current fork. In my fork, I made the following modifications based on my personal habits [...]

This does not answer my questions. Please show evidence (a comparison between current master and your version) where the delay has an impact on the UI speed.

GuangDai:
> I would appreciate more focused replies to my answers.
> 
> You are comparing the preview speed from the latest released version of Maccy to your fork, which already includes changes from this repository. Please give a comparison between the current master branch and your version. The current version on master already handles large text much better.
> 
> > And memory leak. If you used for a long time (10+ hours) and always trigger with shortcuts, then run leak $PID(Maccy's PID). You will find the evidence.
> 
> Did you check that Keyboard shortcut registration is actually at fault?
> 
> > However, the behavior will differ from my current fork. In my fork, I made the following modifications based on my personal habits [...]
> 
> This does not answer my questions. Please show evidence (a comparison between current master and your version) where the delay has an impact on the UI speed.

Did you actually inspect the sample file? Did you even bother to check where my fork started? Or did you seriously assume that I invented a problem first and then submitted a random PR without doing the work?

I have already provided detailed explanations and supporting context, yet it is becoming increasingly obvious that you may not have read any of it. So I will ask you directly: do you actually understand the sample file? Do you understand exactly where I forked from?

I have closed the PR. At this point, I no longer care.

A code review is not just firing off questions without first reading the code, checking the evidence, or understanding the context. My time matters too, and I am not going to keep wasting it compensating for a reviewer who cannot be bothered to do the most basic due diligence.

weisJ:
For future reference here is evidence that keyboard shortcut registration and unregistration does not cause any measurable memory usage.

````swift
import AppKit
import KeyboardShortcuts
import SwiftUI

@main
struct MaccyApp: App {
  var body: some Scene {
    Window("Maccy Shortcut Memory Test", id: "shortcut-memory-test") {
      ShortcutMemoryTestView()
    }
    .windowResizability(.contentSize)
  }
}

private struct ShortcutMemoryTestView: View {
  @StateObject private var cycler = ShortcutCycler()

  var body: some View {
    Toggle(
      isOn: Binding(
        get: { cycler.isRunning },
        set: cycler.setRunning
      )
    ) {
      Text(cycler.isRunning ? "Stop shortcut cycling" : "Start shortcut cycling")
    }
    .toggleStyle(.button)
    .controlSize(.large)
    .padding(32)
    .onAppear {
      NSApp.activate(ignoringOtherApps: true)
    }
    .onDisappear {
      cycler.setRunning(false)
    }
  }
}

private final class ShortcutCycler: ObservableObject {
  @Published private(set) var isRunning = false

  private let queue = DispatchQueue(
    label: "org.p0deje.Maccy.shortcut-memory-test",
    qos: .userInitiated
  )
  private var cancellationToken: CancellationToken?

  init() {
    // Initialize the shortcut on the main thread before the test starts.
    KeyboardShortcuts.disable(.popup)
  }

  func setRunning(_ shouldRun: Bool) {
    guard shouldRun != isRunning else { return }

    isRunning = shouldRun

    if shouldRun {
      let cancellationToken = CancellationToken()
      self.cancellationToken = cancellationToken

      queue.async {
        var count = 0
        while !cancellationToken.isCancelled {
          autoreleasepool {
            KeyboardShortcuts.enable(.popup)
            KeyboardShortcuts.disable(.popup)
          }
          count += 1
          if count % 100 == 0 {
            print(count)
          }
        }

        KeyboardShortcuts.disable(.popup)
      }
    } else {
      cancellationToken?.cancel()
      cancellationToken = nil
    }
  }

  deinit {
    cancellationToken?.cancel()
  }
}

private final class CancellationToken {
  private let lock = NSLock()
  private var cancelled = false

  var isCancelled: Bool {
    lock.withLock { cancelled }
  }

  func cancel() {
    lock.withLock {
      cancelled = true
    }
  }
}
````
Clicking the button will repeatedly enable and disable the keyboard shortcut (several million types within a few minutes). Watching the activity monitor reveals constant memory consumption for Maccy during the complete time.
As for the large text problem feel free to try out the changes in #1441. Rendering a copy of 10x your example file is almost instant without any truncation of the preview text.

PR #1458, opened 2026-07-21
Bound decoded image memory with an LRU cache

musnows:
## Problem

Maccy's memory usage grows monotonically while browsing history and never drops when the popup is closed. With an image-heavy history it easily reaches several hundred MB (600+ MB RSS observed with a ~400 MB storage database).

Root causes:

1. **Decoded images are cached forever.** Every `HistoryItemDecorator` keeps its `thumbnailImage` and screen-sized `previewImage` until the item is deleted or the history is cleared. A single preview is the size of the whole screen (`previewImageSize` uses `NSScreen.forPopup.visibleFrame`), i.e. ~20 MB on a Retina display, so scrolling through history with the preview pane open accumulates tens of these.
2. **Resized images retain the original.** `NSImage.resized(to:)` used `NSImage(size:flipped:drawingHandler:)`, whose closure captures `self` — every ~200 KB thumbnail kept its full-size original (often a multi-MB uncompressed TIFF) alive.
3. Nothing is released when the popup closes.

## Solution

The raw image data already lives in the on-disk SwiftData store, so decoded images can be treated as a regenerable cache:

- **New `ImageCache` (LRU)** bounds decoded images in memory: at most **5 screen-sized previews** and **50 thumbnails**. When a limit is exceeded, the least recently generated image is evicted (the owning decorator's reference is cleared) and regenerated on demand next time the item is displayed.
- **`NSImage.resized(to:)`** now rasterizes into a standalone `NSBitmapImageRep` (at the screen's backing scale factor, so Retina sharpness is preserved) instead of a drawing handler that retains the original image.
- **Popup close releases all decoded images** (`History.releaseDecodedImages()`); they are regenerated on demand on next open.
- The preview pane's *Dimensions* label now uses the pixel size recorded during decoding (`imagePixelSize`) instead of decoding the original image again.
- `sessionLog` is bounded to the 100 most recent entries (it grew for the whole session and can hold items with decoded image data).

Also fixes a latent bug: after an image is evicted, the generation task references are cleared so the image can actually be regenerated.

## Testing

- Added `testImageCacheEvictsPreviewsBeyondLimit` (eviction keeps only the N most recent previews) and `testCleanupImagesReleasesCachedImages` to `HistoryItemDecoratorTests`; all decorator tests pass.
- Full `MaccyTests` suite: only the pre-existing `ClipboardTests` failures remain (they fail identically on unmodified master in this environment — they depend on the system pasteboard).
- With these changes, decoded image memory is bounded to roughly 120 MB worst case (5 screen-sized previews + 50 thumbnails) and returns to baseline when the popup is closed, instead of growing without limit.

## Known limitation

SwiftData does not expose Core Data's re-fault API, so raw (compressed) image data that was accessed during a session stays in the `ModelContext` until restart. Fully addressing that would require moving image blobs out of the database into file storage — intentionally out of scope for this PR.

### DIFF
--- a/Maccy/Observables/History.swift
+++ b/Maccy/Observables/History.swift
@@ -172,6 +172,12 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
     limitHistorySize(to: Defaults[.size] - 1)
 
     sessionLog[Clipboard.shared.changeCount] = item
+    // sessionLog is only needed to detect copies that modify a recent item,
+    // so keep it bounded instead of growing it for the whole session.
+    if sessionLog.count > 100 {
+      let staleKeys = sessionLog.keys.sorted().dropFirst(sessionLog.count - 100)
+      staleKeys.forEach { sessionLog.removeValue(forKey: $0) }
+    }
 
     var itemDecorator: HistoryItemDecorator
     if let pin = item.pin {
@@ -288,6 +294,15 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
   @MainActor
   private func cleanup(_ item: HistoryItemDecorator) {
     item.cleanupImages()
+  }
+
+  // Releases all decoded images (thumbnails and previews).
+  // The raw data stays in the on-disk database and is decoded
+  // on demand when an item is displayed again.
+  @MainActor
+  func releaseDecodedImages() {
+    all.forEach { $0.cleanupImages() }
+    ImageCache.shared.removeAll()
   }
 
   private func currentModifierFlags() -> NSEvent.ModifierFlags {
--- a/Maccy/Observables/HistoryItemDecorator.swift
+++ b/Maccy/Observables/HistoryItemDecorator.swift
@@ -45,6 +45,10 @@ class HistoryItemDecorator: Identifiable, Hashable, HasVisibility {
   var thumbnailImageGenerationTask: Task<(), Error>?
   var previewImage: NSImage?
   var thumbnailImage: NSImage?
+  // Pixel dimensions of the original image, captured during decoding so that
+  // the preview view does not need to decode the original image again just
+  // to display its dimensions.
+  var imagePixelSize: NSSize?
   var applicationImage: ApplicationImage
 
   // 10k characters seems to be more than enough on large displays
@@ -85,6 +89,9 @@ class HistoryItemDecorator: Identifiable, Hashable, HasVisibility {
     }
     thumbnailImageGenerationTask = Task { [weak self] in
       self?.generateThumbnailImage()
+      // Clear the task reference so the image can be regenerated
+      // after it has been evicted from the cache.
+      self?.thumbnailImageGenerationTask = nil
     }
   }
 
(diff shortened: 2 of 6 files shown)


Maccy/Observables/History.swift
```diff
@@ -172,6 +172,12 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
     limitHistorySize(to: Defaults[.size] - 1)
 
     sessionLog[Clipboard.shared.changeCount] = item
+    // sessionLog is only needed to detect copies that modify a recent item,
+    // so keep it bounded instead of growing it for the whole session.
+    if sessionLog.count > 100 {
+      let staleKeys = sessionLog.keys.sorted().dropFirst(sessionLog.count - 100)
+      staleKeys.forEach { sessionLog.removeValue(forKey: $0) }
+    }
```
weisJ, line 180: This seems unrelated to images. At least keep it in a separate commit.

Maccy/Observables/Popup.swift
```diff
@@ -83,6 +83,11 @@ class Popup {
 
   func close() {
     AppState.shared.appDelegate?.panel.close()  // close() calls reset
+    // Release decoded images once the popup is closed; they are
+    // regenerated on demand when the popup is opened again.
+    Task { @MainActor in
+      History.shared.releaseDecodedImages()
+    }
```
weisJ, line 90: This feels unnecessary given that the cache only stores only a few images now anyway.

Maccy/Observables/HistoryItemDecorator.swift
```diff
@@ -45,6 +45,10 @@ class HistoryItemDecorator: Identifiable, Hashable, HasVisibility {
   var thumbnailImageGenerationTask: Task<(), Error>?
   var previewImage: NSImage?
   var thumbnailImage: NSImage?
+  // Pixel dimensions of the original image, captured during decoding so that
+  // the preview view does not need to decode the original image again just
+  // to display its dimensions.
+  var imagePixelSize: NSSize?
```
weisJ, line 51: Please factor changes to related to the precomputed `imagePixelSize` into a separate commit.

weisJ (changes requested):
As I see it the image cache currently doesn't do anything. Images are stored in there but never retrieved? Right now you just evict all images when the popup closes which forces them to be regenerated again when opening defeating the whole point of even precomputing the images.

I would expect that the image generation tasks should check the cache to retrieve already generated images. Moreover, the cache should not be evicted when the popup closes (otherwise there is no actual use in the cache).

weisJ:
> 1. **Decoded images are cached forever.** Every `HistoryItemDecorator` keeps its `thumbnailImage` and screen-sized `previewImage` until the item is deleted or the history is cleared. A single preview is the size of the whole screen (`previewImageSize` uses `NSScreen.forPopup.visibleFrame`), i.e. ~20 MB on a Retina display, so scrolling through history with the preview pane open accumulates tens of these.

Correction. The screen size is the maximum size of the preview image. Images which are smaller than that will never be resized. In general only the thumbnail images will cause any new allocations at all. It would be great if you could provide some data that this change actually reduces any memory pressure.

musnows:
Sorry, this is my first submission for maccy, and it indeed doesn't quite meet the standards. I will fix it and conduct memory tests locally. After drawing an accurate conclusion, I will update this PR

PR #1459, opened 2026-07-21
Raise popup window level above Chrome autofill

### LINKED ISSUE #1403: Maccy popup appears underneath the Chrome autofill dropdown

2.5.1 (58)

### macOS Version

15.7.4 (24G517)

### Maccy Settings

```Shell
{
    "KeyboardShortcuts_delete" = "{\\"carbonModifiers\\":2048,\\"carbonKeyCode\\":51}";
    "KeyboardShortcuts_pin" = "{\\"carbonModifiers\\":2048,\\"carbonKeyCode\\":35}";
    "KeyboardShortcuts_popup" = "{\\"carbonKeyCode\\":8,\\"carbonModifiers\\":768}";
    "NSWindow Frame SUUpdateAlert" = "438 422 620 398 0 0 1496 933 ";
    "NSWindow Frame com.sindresorhus.Settings.FrameAutosaveName" = "493 360 510 473 0 0 1496 933 ";
    SUAutomaticallyUpdate = 1;
    SUEnableAutomaticChecks = 1;
    SUHasLaunchedBefore = 1;
    SULastCheckTime = "2025-11-25 18:06:27 +0000";
    SUUpdateGroupIdentifier = 2124363697;
        "public.file-url",
        "public.png",
        "public.utf8-plain-text",
        "public.html",
        "public.tiff",
        "public.rtf"
    );
    migrations =     {
        "2024-07-01-version-2" = 1;
    };
    pasteByDefault = 1;
    removeFormattingByDefault = 1;
    showFooter = 1;
    showSearch = 1;
    showTitle = 1;
    windowSize = "[450,800]";
}
```

### Description

Image

### Steps to Reproduce

1. Go here: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/tel
2. Get Chrome to suggest something in the dropdown, by having entered a phone number previously.
3. Open Maccy

MohammedAlmajhali:
## Problem

Fixes #1403.

When Chrome’s autofill / suggestion dropdown is open, opening Maccy shows the clipboard popup **underneath** that dropdown, so the history list is partially or fully obscured.

## Root cause

`FloatingPanel` sets `level = .statusBar` (window layer **25**). Chrome’s autofill popup is created at layer **999**, which sits above `.statusBar` and also above `.popUpMenu` (**101**), so Maccy loses the z-order contest while autofill is visible.

`.statusBar` was previously chosen so the panel stays above Spotlight; that still holds for any level ≥ 25.

## Fix

Raise the panel level to `.screenSaver` (**1000**), which sits just above Chrome’s autofill layer (999) while remaining above status items / Spotlight:

```swift
// Chrome autofill uses window layer 999; screenSaver (1000) sits just above it
// while still covering status items / Spotlight. See #1403.
level = .screenSaver
```

## Verification

* Built and ran a Debug build with the change applied
* Reproduced on Chrome autofill (phone input suggestions): Maccy popup now appears **above** the autofill dropdown
* Confirmed popup still appears above Spotlight
* Outside click / resign-key still dismisses the popup

### DIFF
--- a/Maccy/FloatingPanel.swift
+++ b/Maccy/FloatingPanel.swift
@@ -37,7 +37,9 @@ class FloatingPanel<Content: View>: NSPanel, NSWindowDelegate {
 
     animationBehavior = .none
     isFloatingPanel = true
-    level = .statusBar
+    // Chrome autofill uses window layer 999; screenSaver (1000) sits just above it
+    // while still covering status items / Spotlight. See #1403.
+    level = .screenSaver
     collectionBehavior = [.auxiliary, .stationary, .moveToActiveSpace, .fullScreenAuxiliary]
     titleVisibility = .hidden
     titlebarAppearsTransparent = true


weisJ:
I don't like that we have to do this, clearly this is an issue with Chrome. They should use an appropriate level for that auto-complete popup. Nonetheless, this is probably the easiest fix. Thank you for the contribution.

PR #1460, opened 2026-07-22
Restore macOS ⌃K kill-to-end-of-line at the first item

### LINKED ISSUE #1055: Do not swallow ctrl+k

Version 2.3.0 (54)

### macOS Version

15.2

### Maccy Settings

```Shell
Nothing special, just default
```

### Description

_No response_

### Steps to Reproduce

Currently maccy seems to take ctrl+k to do something else, which prevent the default behavior of kill all right on macos, maccy should not use this keybinding.

MohammedAlmajhali:
## Problem

Fixes #1055

⌃K was always bound to vim-style “move to previous”, so it swallowed the macOS/readline kill-to-end-of-line behavior in the search field. At the first item (or with no results) that navigation is a no-op, so ⌃K appeared to do nothing while typing.

## Root cause

`KeyChord` maps ⌃K to `moveToPrevious`, and `KeyHandlingView` always handled that path. The search field never received a real kill-to-end action.

## Fix

Middle ground suggested by @p0deje :

- If the first visible history item is selected, or there are no visible items → ⌃K kills from the caret to the end of the search query
- Otherwise → keep navigating up with ⌃K

Kill updates `searchQuery` (same approach as ⌃H/⌃W) so SwiftUI does not revert field-editor edits.

## Verification

- Smoke-built on tag `2.5.1`
- Verified manually:
  - With results, on the first item: ⌃K deletes from caret to end of search
  - With results, not on the first item: ⌃K moves selection up
  - With no search results: ⌃K still kills to end of search

### DIFF
--- a/Maccy/KeyChord.swift
+++ b/Maccy/KeyChord.swift
@@ -103,10 +103,9 @@ enum KeyChord: CaseIterable {
     case (.upArrow, []),
          (.p, [.control]):
       self = .moveToPrevious
-    case (.k, [.control]):
-      // With no search results, let the text field keep macOS kill-to-end-of-line (⌃K).
-      // Otherwise keep vim-style navigation. See https://github.com/p0deje/Maccy/issues/1055
-      self = AppState.shared.history.firstVisibleItem == nil ? .ignored : .moveToPrevious
+    case (.k, [.control]) where !AppState.shared.navigator.isFirstItemHighlighted:
+      // See https://github.com/p0deje/Maccy/issues/1055
+      self = .moveToPrevious
     case (.upArrow, [.command, .shift]),
          (.upArrow, [.option, .shift]),
          (.p, [.control, .option, .shift]):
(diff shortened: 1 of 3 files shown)


p0deje (changes requested):
This needs to be refactored to get merged:

1. Ctrl+J key chord should be detected via `Maccy/KeyChord.swift`.
2. Let's only delete to the end of input when there are 0 search results.
3. To remove, you need to return `.ignored` from `Maccy/Views/KeyHandlingView.swift`.

MohammedAlmajhali:
@p0deje Updated per your review:

1. ⌃K is decided in `KeyChord.swift` (`.ignored` vs `.moveToPrevious`)
2. Kill-to-end only when there are **0** search results
3. `KeyHandlingView` returns `.ignored` so the search field performs the delete

Verified manually on both cases (with results → navigate; with no results → kill-to-end).

Issue #1470, opened 2026-07-29
Clipboard history and preview pane layout issues


Yamonov:
### Before Submitting Your Bug Report

- [x] I have verified that there isn't already an issue reporting the same bug to prevent duplication.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Maccy Version (see 'About' window)

2.7.0(61)

### macOS Version

26.5.2

### Maccy Settings

```Shell
{
    "KeyboardShortcuts_delete" = "{\\"carbonKeyCode\\":2,\\"carbonModifiers\\":4096}";
    "KeyboardShortcuts_pin" = "{\\"carbonKeyCode\\":1,\\"carbonModifiers\\":4096}";
    "KeyboardShortcuts_popup" = "{\\"carbonKeyCode\\":9,\\"carbonModifiers\\":4096}";
    "KeyboardShortcuts_togglePreview" = "{\\"carbonKeyCode\\":49,\\"carbonModifiers\\":4096}";
    "LaunchAtLogin__hasMigrated" = 1;
    NSNavPanelExpandedSizeForOpenMode = "{800, 448}";
    NSOSPLastRootDirectory = {length = 772, bytes = 0x626f6f6b 04030000 00000510 40000000 ... 58010000 00000000 };
    "NSStatusItem Preferred Position Item-0" = 1596;
    "NSStatusItem VisibleCC Item-1" = 0;
    "NSWindow Frame NSNavPanelAutosaveName" = "868 661 800 448 0 0 2560 1415 ";
    "NSWindow Frame com.sindresorhus.Preferences.FrameAutosaveName" = "986 554 587 382 0 0 2560 1415 ";
    "NSWindow Frame com.sindresorhus.Settings.FrameAutosaveName" = "1039 627 510 520 0 0 2560 1409 ";
    SUEnableAutomaticChecks = 1;
    SUHasLaunchedBefore = 1;
    SULastCheckTime = "2026-07-29 03:57:50 +0000";
    SUUpdateGroupIdentifier = 778032096;
    avoidTakingFocus = 1;
    clearOnQuit = 1;
    clearSystemClipboard = 1;
    enabledPasteboardTypes =     (
        "public.heic",
        "public.html",
        "public.jpeg",
        "public.utf8-plain-text",
        "public.png",
        "public.tiff",
        "public.rtf"
    );
    historySize = 500;
    ignoreEvents = 0;
    ignoredApps =     (
[...]

PR #1471, opened 2026-07-29
Layout fixes

weisJ:
Fixes the issues described in #1470. Not sure what the best minimum height is. For now I chose whatever space pins, footer, header etc takes up plus space for 3 history items.

### DIFF
--- a/Maccy/Observables/Popup.swift
+++ b/Maccy/Observables/Popup.swift
@@ -43,6 +43,11 @@ class Popup {
   var extraTopHeight: CGFloat = 0
   var extraBottomHeight: CGFloat = 0
   var footerHeight: CGFloat = 0
+  
+  var minimumHeight: CGFloat {
+    // Reserve space for 3 items
+    return suitableHeight(for: 3 * Popup.itemHeight)
+  }
 
   private var eventsMonitor: Any?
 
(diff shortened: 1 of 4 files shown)


p0deje:
@weisJ Can you share few screenshots showing what it looks like now?

weisJ:
Pinned items, no unpinned (top/bottom pins).
With footer:
image
image

Without footer:
image
image

Pinned items, with unpinned (top/bottom pins).
With footer:
image
image

Without footer:
image
image

No pinned items, with unpinned:
With footer:
image

Without footer:
image

Completely empty:
image

Maccy/Observables/AppState.swift
```diff
@@ -8,7 +8,7 @@ import SwiftUI
 class AppState: Sendable {
   static let shared = AppState(history: History.shared, footer: Footer())
 
-  let multiSelectionEnabled = false
+  let multiSelectionEnabled = true
```
p0deje: I think you want this reverted 😅

Maccy/Observables/Popup.swift
```diff
@@ -43,6 +43,11 @@ class Popup {
   var extraTopHeight: CGFloat = 0
   var extraBottomHeight: CGFloat = 0
   var footerHeight: CGFloat = 0
+  
+  var minimumHeight: CGFloat {
```
p0deje, line 47: We can use the default window height instead (800)

Maccy/Observables/AppState.swift
weisJ: Yes, fixed.

Maccy/Observables/Popup.swift
weisJ: 800 is very large. With footer enabled and a few pins it takes up almost 3/4 of the vertical space of my MacBook screen. I think people should have the option to choose a smaller height than the default.

p0deje: Fair enough, agreed.

PR #1472, opened 2026-07-29
fix: crash re-invoking bottom-most pinned item at full capacity (#1466)

### LINKED ISSUE #1466: Maccy 2.6.1 crashes when invoking the bottom-most pinned item at full history capacity
### Before submitting

- [x] I searched existing open issues and did not find a matching report.
- [x] I reviewed the FAQ/documentation.

### Environment

- Maccy 2.6.1, build 60 (Mac App Store edition)
- macOS 26.5.2 (25F84)
- MacBookPro16,1, Intel x86_64

### Description

Maccy reliably crashes when I open the popup with `⌘⇧C` and invoke a pinned item using its assigned `⌘letter` shortcut.

The item initially had `⌘E`. I unpinned and re-pinned it, which assigned `⌘D`; the same crash recurred. Selecting the same item with keyboard navigation and Return works, so the failure appears specific to the direct pinned-item shortcut path rather than to a particular shortcut letter.

The pinned record was copied from a Chromium-based app. Without including its clipboard contents, its pasteboard types are:

- `public.html`
- `public.utf8-plain-text`
- `org.chromium.internal.source-rfh-token`
- `org.chromium.source-url`

### Steps to reproduce

1. Copy rich/plain text from a Chromium-based app.
2. Open Maccy with `⌘⇧C`.
3. Select that item and pin it with `⌥P`.
4. Close and reopen Maccy with `⌘⇧C`.
5. Press the assigned direct shortcut, such as `⌘D` or `⌘E`.

### Expected behavior

The pinned item is loaded into the clipboard and Maccy closes normally.

### Actual behavior

Maccy quits unexpectedly.

This reproduced after assigning two different pin letters. Six Maccy crash reports were generated on the same day.

### Crash signature

- Exception: `EXC_BAD_INSTRUCTION`
- Signal: `SIGILL`
- Crashed thread: main thread
- Latest incident ID: `23C67C48-B4DA-45E1-BADD-46CFDBC14C95`

The App Store binary is stripped, so the report contains offsets rather than useful Maccy symbols. A complete `.ips` report is available if helpful.

### Possible existing fix / regression evidence

Post-2.6.1 commit `aef7c524d8991648804c081ef05ddd9776229b54` adds both observed Chromium metadata types to `HistoryItem.transientTypes`:

- `org.chromium.source-url`
- `org.chromium.internal.source-rfh-token`

I applied that commit to the 2.6.1 tag and added a regression test confirming Chromium metadata does not prevent an otherwise equivalent clipboard item from superseding the existing item. All 15 focused `HistoryItemTests` pass. This is a strong lead, but I have not yet claimed it as an end-to-end fix against the affected installed clipboard history.

Sublio:
Attempt to fix #1466 
STR:
1. Settings → Appearance → Pin to = Bottom.
2. Settings → Storage → Size = a small number (e.g. 5) to fill history quickly.
3. Copy one text snippet, open Maccy (⌘⇧C), select it, and pin it (⌥P). With bottom pinning it sits at the very bottom.
4. Copy 5 more distinct snippets so the unpinned history is exactly full (5/5).
5. Open Maccy again and invoke the bottom pinned item — via its ⌘<letter> shortcut, or hover + Enter.

Fix is: all.insert(itemDecorator, at: min(removedItemIndex, all.count))
Regression test to cover this case: HistoryTests/testReaddingBottomMostPinnedItemAtFullCapacity

### DIFF
--- a/Maccy/Observables/History.swift
+++ b/Maccy/Observables/History.swift
@@ -179,9 +179,10 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
     var itemDecorator: HistoryItemDecorator
     if let pin = item.pin {
       itemDecorator = HistoryItemDecorator(item, shortcuts: KeyShortcut.create(character: pin))
-      // Keep pins in the same place.
       if let removedItemIndex {
-        all.insert(itemDecorator, at: removedItemIndex)
+        // If pin to bottom -> last element should be inserted to the removedItemIndex - 1
+        // Or to the last all array place.
+        all.insert(itemDecorator, at: min(removedItemIndex, all.count))
       }
     } else {
       itemDecorator = HistoryItemDecorator(item)
--- a/MaccyTests/HistoryTests.swift
+++ b/MaccyTests/HistoryTests.swift
@@ -6,6 +6,7 @@ import Defaults
 class HistoryTests: XCTestCase {
   let savedSize = Defaults[.size]
   let savedSortBy = Defaults[.sortBy]
+  let savedPinTo = Defaults[.pinTo]
   let history = History.shared
 
   override func setUp() {
@@ -13,12 +14,14 @@ class HistoryTests: XCTestCase {
     history.clearAll()
     Defaults[.size] = 10
     Defaults[.sortBy] = .firstCopiedAt
+    Defaults[.pinTo] = .bottom
   }
 
   override func tearDown() {
     super.tearDown()
     Defaults[.size] = savedSize
     Defaults[.sortBy] = savedSortBy
+    Defaults[.pinTo] = savedPinTo
   }
 
   func testDefaultIsEmpty() {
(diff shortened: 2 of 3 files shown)


MaccyTests/HistoryTests.swift
```diff
@@ -226,6 +226,35 @@ class HistoryTests: XCTestCase {
     XCTAssertFalse(history.items.contains(items[5]))
   }
 
+  func testReaddingBottomMostPinnedItemAtFullCapacity() {
+    // Regression test for a crash when re-copying (invoking) the bottom-most
+    // pinned item while history is at full capacity and pins are sorted to the
+    // bottom. The stale insert index used to trap with an out-of-bounds insert.
+    let savedPinTo = Defaults[.pinTo]
+    Defaults[.pinTo] = .bottom
+    defer { Defaults[.pinTo] = savedPinTo }
```
p0deje: Let's use setUp/tearDown for this.

MaccyTests/HistoryTests.swift
```diff
@@ -226,6 +226,35 @@ class HistoryTests: XCTestCase {
     XCTAssertFalse(history.items.contains(items[5]))
   }
 
+  func testReaddingBottomMostPinnedItemAtFullCapacity() {
+    // Regression test for a crash when re-copying (invoking) the bottom-most
+    // pinned item while history is at full capacity and pins are sorted to the
+    // bottom. The stale insert index used to trap with an out-of-bounds insert.
```
p0deje, line 235: Add a link to Github issue.

Maccy/Observables/History.swift
```diff
@@ -179,9 +179,10 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
     var itemDecorator: HistoryItemDecorator
     if let pin = item.pin {
       itemDecorator = HistoryItemDecorator(item, shortcuts: KeyShortcut.create(character: pin))
-      // Keep pins in the same place.
       if let removedItemIndex {
-        all.insert(itemDecorator, at: removedItemIndex)
+        //If pin to bottom -> last elem should be inserted to the removedItemIndex - 1
```
p0deje:
```suggestion
        // If pin to bottom -> last element should be inserted to the removedItemIndex - 1
```

Sublio: Fixed

MaccyTests/HistoryTests.swift
Sublio: Done

MaccyTests/HistoryTests.swift
Sublio: Done

MaccyTests/HistoryTests.swift
```diff
@@ -226,6 +227,35 @@ class HistoryTests: XCTestCase {
...
+    // Issue link: https://github.com/p0deje/Maccy/issues/1466
+    // `pinTo` is restored to its saved value in `tearDown`.
```
p0deje: This is clearly not true!

Maccy/Observables/History.swift
```diff
@@ -179,9 +179,10 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
     var itemDecorator: HistoryItemDecorator
     if let pin = item.pin {
       itemDecorator = HistoryItemDecorator(item, shortcuts: KeyShortcut.create(character: pin))
-      // Keep pins in the same place.
       if let removedItemIndex {
-        all.insert(itemDecorator, at: removedItemIndex)
+        //If pin to bottom -> last element should be inserted to the removedItemIndex - 1
```
p0deje: Please add a space after a comment symbol




```suggestion
        // If pin to bottom -> last element should be inserted to the removedItemIndex - 1
```

MaccyTests/HistoryTests.swift
Sublio: Added Defaults[.pinTo] = .top in tearDown

Maccy/Observables/History.swift
Sublio: Done

PR #1492, opened 2026-08-07
Fix String.shortened(to:) returning one character too many

YuriNachos:
## Problem

`String.shortened(to:)` returned one character too many. It used a closed range `self[...index(startIndex, offsetBy: maxLength)]`, which includes the end index — so `"abcdef".shortened(to: 3)` returned `"abcd"` (4 characters) instead of `"abc"`.

## Fix

Use a half-open range `self[startIndex..<index(startIndex, offsetBy: maxLength)]`, so the result contains at most `maxLength` characters.

## Test

New `ShortenedTests` covers: returns at most `maxLength`; no-op when within limit; no-op for a large cap; and a loop asserting the result never exceeds `maxLength` for lengths 0…8.


p0deje:
Thank you!

PR #1498, opened 2026-08-08
Fix orphaned history item content cleanup

### LINKED ISSUE #1496: HistoryItemContent rows are orphaned instead of cascade-deleted on eviction — store grew to 1.89 GB in 11 days (2.7.0)

Related but not duplicate: #1368 fixed only the duplicate-insertion path (closed with 2.7.0). #274 reported the same underlying behaviour back in 2021 and was fixed then for Core Data; it appears to have **regressed after the SwiftData migration**. The dominant leak path — normal history eviction — is still present in 2.7.0.

2.7.0 (leak also confirmed on 2.6.1)

### macOS Version

26.3 (25D125), Apple Silicon

### Maccy Settings

```
enabledPasteboardTypes = ("public.html", "public.tiff", "public.rtf", "public.png", "public.file-url", "public.utf8-plain-text");
historySize = 100;
sortBy = lastCopiedAt;
```

### Description

`HistoryItemContent` rows are never deleted when their parent `HistoryItem` is evicted. Instead their `item` reference is set to `NULL`, leaving orphaned rows that accumulate forever. Since these rows hold the actual pasteboard payloads (including images), the store grows without bound.

**Measured on a normal install after 11 days of uninterrupted use:**

| | |
|---|---|
| `Storage.sqlite` size | **1.89 GB** |
| `ZHISTORYITEM` rows | 100 (respecting `historySize`) |
| `ZHISTORYITEMCONTENT` rows | 15,339 |
| — of which orphaned (`ZITEM IS NULL`) | **15,062 (98.2%)** |
| — of which live | 277 |
| `ZHISTORYITEM` max primary key | 21,631 |

So ~21,500 evicted items left their content behind. Every orphan had `ZITEM IS NULL` and **zero** rows pointed at a non-existent `ZHISTORYITEM`, which indicates the relationship is being **nullified rather than cascaded** on delete, despite the declared rule:

https://github.com/p0deje/Maccy/blob/2.7.0/Maccy/Models/HistoryItem.swift#L74-L75

```swift
@Relationship(deleteRule: .cascade, inverse: \HistoryItemContent.item)
var contents: [HistoryItemContent] = []
```

Payload breakdown before cleanup — images dominate because macOS puts screenshots on the pasteboard as uncompressed TIFF:

| Type | Rows | Size |
|---|---|---|
| `public.tiff` | 152 | 1439 MB |
| `public.png` | 1171 | 327 MB |
| everything else (text, html, urls…) | ~14,000 | ~10 MB |

Of the 1768 MB held by image rows, 1742 MB (98.5%) sat in orphaned rows. Purging the image payloads plus the entries left empty by that purge took the store from 1.89 GB to 20 MB, with the 67 remaining history entries intact and `PRAGMA integrity_check` returning `ok`.

Re-measuring 1.5 h of ordinary use later showed **+811 new orphans** (13,620 → 14,431), so the leak is continuous rather than a one-off artefact of some past `clearAll()`.

### Suspected code paths

1. **`History.limitHistorySize(to:)` → `History.delete(_:)`** — the dominant path. It calls `Storage.shared.context.delete(item.item)`, relying on the cascade rule that empirically does not fire.
   https://github.com/p0deje/Maccy/blob/2.7.0/Maccy/Observables/History.swift#L121-L124
   https://github.com/p0deje/Maccy/blob/2.7.0/Maccy/Observables/History.swift#L271-L289

2. **`History.clearAll()`** — deletes only the parent model and never touches `HistoryItemContent`:
   https://github.com/p0deje/Maccy/blob/2.7.0/Maccy/Observables/History.swift#L249-L262
   ```swift
   try? Storage.shared.context.delete(model: HistoryItem.self)
   ```
   Note `clear()` right above it *does* delete both models explicitly, so the asymmetry looks unintentional. (SwiftData's batch `delete(model:where:)` does not run cascade rules either, so the explicit second delete in `clear()` is required — `clearAll()` is missing its counterpart.)

An explicit `item.contents.forEach(context.delete)` before deleting the parent in `delete(_:)`, plus the missing `HistoryItemContent` delete in `clearAll()`, should close both paths. A one-time migration purging `ZITEM IS NULL` rows would be needed to reclaim space for existing users — this was also requested in #274.

### Steps to Reproduce

1. Use Maccy normally until more than `historySize` items have been copied (copying a few screenshots makes the size effect obvious immediately).
2. Quit Maccy and inspect the store:

```sh
sqlite3 "$HOME/Library/Containers/org.p0deje.Maccy/Data/Library/Application Support/Maccy/Storage.sqlite" \
  "SELECT (SELECT COUNT(*) FROM ZHISTORYITEM)                             AS items,
          (SELECT COUNT(*) FROM ZHISTORYITEMCONTENT)                      AS contents,
          (SELECT COUNT(*) FROM ZHISTORYITEMCONTENT WHERE ZITEM IS NULL)  AS orphaned;"
```

`orphaned` grows monotonically and never returns to 0.

### Expected Behaviour

Evicting or clearing a `HistoryItem` removes its `HistoryItemContent` rows, so the store stays proportional to `historySize`.

### Note on privacy impact

Because the orphaned rows retain the original pasteboard payloads, content a user believes has aged out of history (or was explicitly cleared via `clearAll()`) remains readable on disk indefinitely. This was raised as a security concern in #274 as well. Users who copy passwords or tokens are likely unaware that this data persists.

hhe48203-ctrl:
Fixes #1496

## Summary

- Delete history content explicitly before deleting its parent item, instead of relying on SwiftData cascade behavior.
- Remove all live content during clear operations and purge any pre-existing orphaned content.
- Run a one-time startup cleanup for persisted orphaned rows, retrying on a later launch if it fails.
- Avoid persisting discarded content while resolving duplicate history entries.

## Tests

- Added orphan-count assertions for duplicate merging, size eviction, individual deletion, clearing unpinned history, and clearing all history.
- Added coverage for cleaning an existing orphan without deleting live content.

## Validation

- `xcodebuild test -quiet -project Maccy.xcodeproj -scheme Maccy -destination 'platform=macOS,arch=arm64' -only-testing:MaccyTests/HistoryTests CODE_SIGNING_ALLOWED=NO CODE_SIGN_IDENTITY=''` — passed (14 tests).
- The full `MaccyTests` suite has two existing system-pasteboard failures in this local environment: `ClipboardTests/testIgnoreAllApplicationsExcept` and `ClipboardTests/testIgnoreApplication`. Both reproduce unchanged on commit `02dd0a3`.

### DIFF
--- a/Maccy/AppDelegate.swift
+++ b/Maccy/AppDelegate.swift
@@ -122,6 +122,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     }
   }
 
+  @MainActor
   private func migrateUserDefaults() {
     ensureMigration(key: "2024-07-01-version-2") {
       // Start 2.x from scratch.
@@ -144,6 +145,12 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         types.formUnion(StorageType.images.types)
       }
       Defaults[.enabledPasteboardTypes] = types
+    }
+
+    if Defaults[.migrations]["2026-08-08-cleanup-orphaned-history-item-contents"] != true {
+      if (try? Storage.shared.cleanupOrphanedContents()) != nil {
+        Defaults[.migrations]["2026-08-08-cleanup-orphaned-history-item-contents"] = true
+      }
     }
 
     // The following defaults are not used in Maccy 2.x
--- a/Maccy/Observables/History.swift
+++ b/Maccy/Observables/History.swift
@@ -136,16 +136,10 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
   @discardableResult
   @MainActor
   func add(_ item: HistoryItem) -> HistoryItemDecorator {
-    if #available(macOS 15.0, *) {
-      try? History.shared.insertIntoStorage(item)
-    } else {
-      // On macOS 14 the history item needs to be inserted into storage directly after creating it.
-      // It was already inserted after creation in Clipboard.swift
-    }
-
     var removedItemIndex: Int?
     if let existingHistoryItem = findSimilarItem(item) {
       if isModified(item) == nil {
+        deleteContents(of: item)
         item.contents = existingHistoryItem.contents
       }
       item.firstCopiedAt = existingHistoryItem.firstCopiedAt
@@ -160,7 +154,7 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
       if let removedItemIndex {
         cleanup(all[removedItemIndex])
       }
-      Storage.shared.context.delete(existingHistoryItem)
+      deleteFromStorage(existingHistoryItem)
       if let removedItemIndex {
         all.remove(at: removedItemIndex)
       }
@@ -173,6 +167,13 @@ class History: ItemsContainer { // swiftlint:disable:this type_body_length
     // Remove exceeding items. Do this after the item is added to avoid removing something
     // if a duplicate was found as then the size already stayed the same.
     limitHistorySize(to: Defaults[.size] - 1)
+
+    if #available(macOS 15.0, *) {
+      try? insertIntoStorage(item)
+    } else {
+      // On macOS 14 the history item needs to be inserted into storage directly after creating it.
+      // It was already inserted after creation in Clipboard.swift
+    }
 
     sessionLog[Clipboard.shared.changeCount] = item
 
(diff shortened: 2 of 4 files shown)


p0deje:
Merged with improvements in #1509. Thank you.

PR #1500, opened 2026-08-08
Fix off-by-one index bound in clipboard-history Shortcuts actions

YuriNachos:
## Problem

The three clipboard-history Shortcuts actions — **Delete**, **Select**, and **Get Item from Clipboard History** — crash with an index-out-of-range error when `Number` is set to exactly one past the last item (e.g. `Number = 4` when 3 items exist). `Number = 0` (and any value below 1) crashes the same way.

## Root cause

Each intent converts the 1-based `number` to a 0-based `index` (`index = number - 1`) and then guards the subscript with `items.count >= index`. With 3 items and `number = 4` → `index = 3`:

```
items.count(3) >= index(3)  →  true  →  guard passes  →  items[3]  →  Fatal error: Index out of range
```

The largest valid 0-based subscript for a 3-element array is `2`, so the guard is off by one. The lower end is also unprotected: `number = 0` → `index = -1` → `3 >= -1` passes and traps on `items[-1]`.

## Fix

Make the bound reject both ends, using the existing `AppIntentError.notFound` path (no new error and no change to the valid-path behavior):

- `Delete` / `Select`:
  ```swift
  guard index >= 0, items.count > index else { throw AppIntentError.notFound }
  ```
- `Get`:
  ```swift
  if index >= 0, AppState.shared.history.items.count > index { item = … }
  ```
  (out of range leaves `item` nil, so the existing `guard let item else { throw .notFound }` applies)

| `number` (3 items) | result before | result after |
| --- | --- | --- |
| `1`–`3` | resolves the item | resolves the item (unchanged) |
| `4` (one past last) | **crash** (Index out of range) | `.notFound` |
| `0` / negative | **crash** | `.notFound` |

No intent signatures, parameter types, or valid-path behavior are changed.

## Test

Adds `MaccyTests/ClipboardHistoryItemIntentsTests.swift` — a headless XCTest that builds a 3-item history and, for each of the three intents, asserts:

- `number = 4` (one past last) → `AppIntentError.notFound`, no mutation,
- `number = 3` (last valid) → resolves the correct last item,
- `number = 0` → `AppIntentError.notFound`, no mutation.

Verified **RED → GREEN**: on the pre-fix code the `number = 4` case crashes with `Fatal error: Index out of range`; after the fix it throws `.notFound` cleanly and the last valid item resolves.

## Notes for review

- The bound fix is the entire behavior change; the diff in `Maccy/Intents/` is one operator per file.
- The full `MaccyTests` suite has two pre-existing, unrelated failures on this machine (`ClipboardTests.testIgnoreApplication` and `testIgnoreAllApplicationsExcept`) — they depend on the system clipboard-changed hook, which does not fire under the `xcodebuild` test host, and they fail identically on `master` without this change.

### DIFF
--- a/Maccy.xcodeproj/project.pbxproj
+++ b/Maccy.xcodeproj/project.pbxproj
@@ -38,6 +38,7 @@
 		4762D6972467226100B3A2BA /* Localizable.strings in Resources */ = {isa = PBXBuildFile; fileRef = 4762D6992467226100B3A2BA /* Localizable.strings */; };
 		CA110ED10000000000000001 /* CollectionSurroundingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = CA110ED10000000000000002 /* CollectionSurroundingTests.swift */; };
 		C0FFEE000000000000000003 /* ShortenedTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0FFEE000000000000000004 /* ShortenedTests.swift */; };
+		DECAFB00000000000000000A /* ClipboardHistoryItemIntentsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DECAFB00000000000000000B /* ClipboardHistoryItemIntentsTests.swift */; };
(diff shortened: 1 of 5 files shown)


p0deje:
This is clearly wrong because the offset is already taken into account. If you have 4 items and take number 4, there is no crash.

PR #1501, opened 2026-08-09
Fix Throttler delaying the first invocation by minimumDelay

YuriNachos:
# Fix Throttler delaying the first invocation by minimumDelay

`Throttler` delayed the very first invocation by the full `minimumDelay` instead of
firing it immediately. Every call — including the first — waited the whole delay, so the
throttled action always lagged on first use.

## Steps to reproduce

1. `Throttler` is used by `History` to throttle search-as-you-type (`History.swift`, `Throttler(minimumDelay: 0.2)`).
2. On the first keystroke the time since the previous run is effectively infinite (`previousRun` starts at `Date.distantPast`), so the search should run immediately.
3. Instead the first search result always lags by the full `0.2s`.

## Root cause

`Maccy/Throttler.swift:31` measured elapsed time in the wrong direction:

```swift
// before — previousRun.timeIntervalSinceNow == (previousRun - now), always negative
let delay = previousRun.timeIntervalSinceNow > minimumDelay ? 0 : minimumDelay
```

`previousRun.timeIntervalSinceNow` is `previousRun − now`, a **negative** number (the previous
run is in the past, or `Date.distantPast` initially). Comparing `negative > minimumDelay` is
always `false`, so `delay` was always `minimumDelay` and the action never fired immediately —
contradicting the method's own comments ("If the time since the previous run is more than the
required minimum delay => execute the workItem immediately").

## Fix is

```swift
let delay = Date().timeIntervalSince(previousRun) > minimumDelay ? 0 : minimumDelay
```

Measure elapsed time since the previous run as `now − previousRun` (positive). On the first call
(`distantPast`) the value is enormous → `delay = 0` (immediate); a call within `minimumDelay` →
`delay = minimumDelay` (deferred). Based on the original craftappco throttler, which negates the
same expression (`-previousRun.timeIntervalSinceNow`); the negation was dropped here.

## Regression test

`MaccyTests/ThrottlerTests.swift`:

- `testFirstInvocationFiresImmediately` — the first call must fire within `minimumDelay / 2`.
  **Failed (RED)** against the old line (the block fired at ~`minimumDelay`, exceeding the 0.15s
  wait, 3 retries); **passes (GREEN, ~0.001s)** after the fix.
- `testSubsequentInvocationWithinMinimumDelayIsDeferred` — a second call made within
  `minimumDelay` of the previous run is deferred by ~`minimumDelay`.
- `testInvocationAfterMinimumDelayFiresImmediately` — a call made after `minimumDelay` has elapsed
  fires immediately again (the repeat-caller immediate path). Also **RED** on the old line, **GREEN**
  after the fix.

## Verification (macOS, ad-hoc signing `CODE_SIGNING_ALLOWED=NO`)

- `xcodebuild test -only-testing:MaccyTests` → 81 tests. The only failures are the pre-existing,
  host-specific `ClipboardTests/testIgnoreApplication` and `ClipboardTests/testIgnoreAllApplicationsExcept`
  (hardcoded `// Finder is on Bitrise`, fail off-Bitrise) — unchanged by this PR.
- `swiftlint lint --strict` on touched files → 0 violations.

## Diff

`Maccy/Throttler.swift` (1 line) + `MaccyTests/ThrottlerTests.swift` (new) + project wiring.

### DIFF
--- a/Maccy.xcodeproj/project.pbxproj
+++ b/Maccy.xcodeproj/project.pbxproj
@@ -38,6 +38,7 @@
 		4762D6972467226100B3A2BA /* Localizable.strings in Resources */ = {isa = PBXBuildFile; fileRef = 4762D6992467226100B3A2BA /* Localizable.strings */; };
 		CA110ED10000000000000001 /* CollectionSurroundingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = CA110ED10000000000000002 /* CollectionSurroundingTests.swift */; };
 		C0FFEE000000000000000003 /* ShortenedTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0FFEE000000000000000004 /* ShortenedTests.swift */; };
+		C0DECED0000000000000000D /* ThrottlerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C0DECED0000000000000000E /* ThrottlerTests.swift */; };
 		DA009931256411F90030E697 /* appcast.xml in Resources */ = {isa = PBXBuildFile; fileRef = DA00992C256411F90030E697 /* appcast.xml */; };
 		DA009932256411F90030E697 /* README.md in Resources */ = {isa = PBXBuildFile; fileRef = DA00992D256411F90030E697 /* README.md */; };
 		DA009934256411F90030E697 /* LICENSE in Resources */ = {isa = PBXBuildFile; fileRef = DA00992F256411F90030E697 /* LICENSE */; };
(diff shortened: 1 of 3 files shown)


p0deje:
This is wrong and should not be implemented like this - the goal of the throttler is to wait until the next key press.

weisJ:
I think this is actually valid.
````swift
print(previousRun, previousRun.timeIntervalSinceNow, delay)
````
produces
````
0001-01-01 00:00:00 +0000 -63922120831.47771 0.2
2026-08-10 08:40:31 +0000 -0.29634201526641846 0.2
2026-08-10 08:40:31 +0000 -0.29934096336364746 0.2
2026-08-10 08:40:32 +0000 -2.042220950126648 0.2
2026-08-10 08:40:34 +0000 -1.0901830196380615 0.2
2026-08-10 08:40:35 +0000 -4.0037349462509155 0.2
2026-08-10 08:40:39 +0000 -2.12409508228302 0.2
2026-08-10 08:40:42 +0000 -2.096688985824585 0.2
````
`previousRun.timeIntervalSinceNow` is always negative as `previousRun` happened before `now`.
Notice how even after waiting a few seconds between key strokes the delay is always `0.2`.

PR #1509, opened 2026-08-12
Prevent and cleanup orphaned `HistoryItemContent`

p0deje:
Based on #1498 with a few improvements.

### DIFF
--- a/Maccy/AppDelegate.swift
+++ b/Maccy/AppDelegate.swift
@@ -137,6 +137,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     }
   }
 
+  @MainActor
   private func migrateUserDefaults() {
     ensureMigration(key: "2024-07-01-version-2") {
       // Start 2.x from scratch.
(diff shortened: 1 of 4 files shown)


Maccy/AppDelegate.swift
```diff
@@ -161,6 +162,12 @@ class AppDelegate: NSObject, NSApplicationDelegate {
       Defaults[.enabledPasteboardTypes] = types
     }
 
+    let orphanCleanupKey = "2026-08-08-cleanup-orphaned-history-item-contents"
+    if Defaults[.migrations][orphanCleanupKey] != true {
+      _ = try? Storage.shared.cleanupOrphanedContents()
+      Defaults[.migrations][orphanCleanupKey] = true
+    }
```
weisJ: This could use the `ensureMigration` helper I introduced. Probably also want to run the cleanup in a background task, as it might hang if there are many orphans in the database.

p0deje: I totally missed `ensureMigration`, will use it.
As for the background task, I'd rather let it block once than try to deal with concurrency on accessing Core Data context in separate actors. I don't ever feel safe doing this - at least not until we're on Swift 6.

PR #1524, opened 2026-08-31
Strip U+FFFC from titles to avoid CoreText hang on macOS 26

### LINKED ISSUE #1520: 100% CPU hang on macOS 26: titles with 2+ U+FFFC (repro for #1382)
## Summary

On macOS 26, Maccy pegs one CPU core at 100% immediately on launch and never
draws its window — no crash report, no way out from inside the app. The trigger
is a stored history entry whose title contains **two or more U+FFFC**
(OBJECT REPLACEMENT CHARACTER) together with at least one non-Latin character.
`ListItemTitleView`'s `.lineLimit(1) + .truncationMode(.middle)` then sends
CoreText into an infinite loop.

This is the missing reproduction condition for #1382, where the fix stalled
because it couldn't be reproduced.

Environment: macOS 26.6.2 (25G83), arm64, Maccy 2.7.1 (Homebrew cask).

## Where the U+FFFC comes from

Rich-text copies with inline images. Each inline image becomes one U+FFFC in the
`public.utf8-plain-text` pasteboard flavor, so copying a chat message with 2+
inline images/emoticons from an IM client produces exactly this shape. Mine came
from Baidu Hi (`com.baidu.hi.mac`); Slack, WeChat, Notes and Mail can all emit it.

`HistoryItem.generateTitle()` takes `previewableText` verbatim, so those U+FFFC
end up in `title` and are persisted to `ZHISTORYITEM.ZTITLE`. That makes the hang
**sticky** — every subsequent launch re-renders the same row and re-hangs.

It also reproduces at runtime: with Maccy idle at 0% CPU, a single `pbcopy` of
such a string took it straight to 100%.

## Reproduction

Two ways.

**In Maccy** — with Maccy running, copy a string containing two U+FFFC plus CJK:

```sh
python3 -c "import subprocess; subprocess.run(['pbcopy'], input='￼复制粘贴测试￼'.encode())"
```

CPU goes to 100% within a second and stays there. It will now also hang on
every launch until the entry is removed from the store.

**Standalone** — `swiftc -O repro.swift -o repro && ./repro`:

```swift
import SwiftUI
import AppKit

struct Row: View {
  let s: String
  var body: some View {
    HStack(spacing: 0) {
      Spacer().frame(width: 10)
      Text(verbatim: s)
        .lineLimit(1)
        .truncationMode(.middle)
        .padding(.trailing, 5)
      Spacer()
    }
    .frame(minHeight: 24)
    .frame(maxWidth: .infinity, alignment: .leading)
  }
}

// The `if` matters: it emits a _ConditionalContent, which is the shape
// ListItemTitleView builds. Measuring a bare Row does NOT hang.
struct Wrap: View {
  let s: String
  let wrapped: Bool
  var body: some View {
    if wrapped {
      VStack(spacing: 0) { ForEach(0..<1, id: \.self) { _ in Row(s: s) } }
    } else {
      Row(s: s)
    }
  }
}

func probe(_ label: String, _ s: String, wrapped: Bool = true, timeout: Double = 5) {
  let sem = DispatchSemaphore(value: 0)
  let t0 = Date()
  let watchdog = Thread {
    if sem.wait(timeout: .now() + timeout) == .timedOut {
      print("\(label): *** HANG (>\(Int(timeout))s, 100% CPU) ***")
      fflush(stdout)
      exit(3)
    }
  }
  watchdog.start()

  let hv = NSHostingView(rootView: Wrap(s: s, wrapped: wrapped))
  hv.frame = NSRect(x: 0, y: 0, width: 440, height: 24)
  _ = hv.fittingSize

  sem.signal()
  print(String(format: "%@: %.2f ms", label, Date().timeIntervalSince(t0) * 1000))
  fflush(stdout)
}

let ofc = "\u{FFFC}"

probe("warmup",           "x", wrapped: false)
probe("one U+FFFC + CJK", ofc + "复制粘贴测试")
probe("two U+FFFC + CJK", ofc + "复制粘贴测试" + ofc)   // <-- hangs here

print("ALL DONE (no hang)")
```

Output on macOS 26 (5/5 runs identical):

```
warmup: 66.70 ms
one U+FFFC + CJK: 2.75 ms
two U+FFFC + CJK: *** HANG (>5s, 100% CPU) ***
```

## Why #1382 couldn't be reproduced

The hang needs the pathological string **and** the right view-tree shape.
In the repro above, measuring a bare `Row` does not hang; wrapping it so the
builder emits a `_ConditionalContent` does. `ListItemTitleView` has exactly
that shape (`if let attributedTitle { … } else { … }`).

So launching with an ordinary history won't show it — you need a clipboard entry
that actually carries multiple U+FFFC. That's a narrow enough condition that
it's easy to miss, and it explains the "couldn't reproduce" on #1382.

## Condition matrix

Each row run 2–5 times; stable.

| Variation | Result |
|---|---|
| 2× U+FFFC + CJK (`复制粘贴测试`) | **hang** |
| 2× U+FFFC + Japanese (`日本語テスト`) | **hang** |
| 2× U+FFFC + Hangul (`한국어`) | **hang** |
| 2× U+FFFC + emoji (`🚀🚀`) | **hang** |
| 3× U+FFFC + CJK | **hang** |
| 1× U+FFFC + CJK | ok |
| 2× U+FFFC + ASCII only | ok |
| 2× U+FFFC + Cyrillic (`привет мир`) | ok |
| 2× U+FFFC + accented Latin (`café naïve`) | ok |
| CJK, no U+FFFC | ok |
| 2× U+FFFC + CJK, `.truncationMode(.tail)` | ok |
| 2× U+FFFC + CJK, `.truncationMode(.head)` | ok |
| 2× U+FFFC + CJK, no `.lineLimit(1)` | ok |

Independent of: proposed width (440, 2000, `.greatestFiniteMagnitude` all hang);
whether the text is short enough to need no truncation at all; `.drawingGroup()`;
`.fixedSize`; `LazyVStack` vs plain `VStack`.

## Stack

Hot main-thread frames from `sample` on the spinning process:

```
20  __NSCoreTypesetterTruncateLine           (UIFoundation)
14  TTruncator::TruncateWithTokenHandler     (CoreText)
13  TTruncator::TruncateStartChars           (CoreText)
 8  TTruncator::TruncateEndChars             (CoreText)
 7  TLine::InsertRun                         (CoreText)
 7  TLine::InsertPartialRun                  (CoreText)
 6  TLine::UpdateCachedMetricsForRun         (CoreText)
 5  TTruncator::CreateToken                  (CoreText)
 5  TLine::FindRunWithCharIndex              (CoreText)
 5  TLine::ClearLine                         (CoreText)
```

Reached via:

```
NSHostingView.layout()
  → ViewGraphRootValueUpdater.render
    → StackLayout.UnmanagedImplementation.placeChildren
      → StyledTextLayoutEngine.sizeThatFits
        → ResolvedStyledText.StringDrawing.sizeThatFits
          → NSAttributedString.MetricsCache.metrics
            → -[NSAttributedString boundingRectWithSize:options:context:]
              → __NSStringDrawingEngine
                → __NSCoreTypesetterTruncateLine   ← never returns
```

Same frames #1382 identified. The underlying defect is in CoreText/UIFoundation,
not in Maccy — but since an ordinary copy can brick the app with no in-app
recovery, a workaround on Maccy's side seems worth having.

## Recovery for anyone hitting this

```sh
pkill -9 -x Maccy
sqlite3 "$HOME/Library/Containers/org.p0deje.Maccy/Data/Library/Application Support/Maccy/Storage.sqlite" \
  "UPDATE ZHISTORYITEM SET ZTITLE = replace(ZTITLE, char(65532), '') WHERE ZTITLE LIKE '%'||char(65532)||'%';"
```

Clipboard contents are untouched — only the display title changes, so pasting
still yields the original data. (Back up `Storage.sqlite` first.)

## Possible fixes

Both verified against the repro:

| | Result |
|---|---|
| A — `.truncationMode(.tail)` on macOS 26 (what #1382 does) | no hang |
| B — strip U+FFFC in `generateTitle()`, keep `.middle` | no hang |

B preserves middle truncation, which is genuinely useful for long URLs and paths
where the distinguishing part is at the end. U+FFFC has no visible glyph in a
plain-text title anyway, so dropping it costs nothing — and it also keeps the
character out of search/fuzzy-match input. The two are independent; both could
be applied.

Happy to send a PR for B if that's the preferred direction.

Primer-Fan:
Fixes #1520.

## What hangs

A title holding two or more U+FFFC (OBJECT REPLACEMENT CHARACTER) next to
non-Latin text sends `__NSCoreTypesetterTruncateLine` into an infinite loop when
`ListItemTitleView` lays it out with `.lineLimit(1)` + `.truncationMode(.middle)`.
One core sits at 100%, the window never draws, and there's no in-app way out.

U+FFFC is the placeholder for an inline attachment, so any rich-text copy with
embedded images carries one per attachment in its plain-text flavor — a chat
message with two inline emoticons is enough. Because `generateTitle()` persisted
the character, the hang came back on every launch.

## Changes

**`String+UnsafeForTitleLayout.swift`** (new) — `containsScalarsUnsafeForTitleLayout`
and `removingScalarsUnsafeForTitleLayout()`, backed by a set that currently holds
only U+FFFC.

**`HistoryItem.generateTitle()`** — strip after `shortened(to: 1_000)`, before the
`showSpecialSymbols` substitutions, so the existing `⏎`/`⇥`/`·` behaviour is
untouched.

**`Storage.sanitizeTitles()` + migration `2026-08-31-sanitize-history-item-titles`** —
heals titles already in the store. Without it, an affected user still can't launch
the app after upgrading; they'd have to edit `Storage.sqlite` by hand, which is
what I had to do. It runs in `migrateUserDefaults()` from
`applicationDidFinishLaunching`, before the panel is built and the history is
first rendered. Follows the shape of the existing
`2026-08-12-cleanup-orphaned-history-item-contents` migration.

## Two implementation notes

**Filtering is per Unicode scalar, not per Character.** `"\u{FFFC}\u{0301}"` is a
single `Character` — U+FFFC plus a combining mark form one grapheme cluster — so
`filter { $0 != "\u{FFFC}" }` doesn't match it and the hang survives. I confirmed
this is not theoretical:

```
2x (FFFC + U+0301) + CJK:  HANG
  after Character filter:  HANG    ← looks right, doesn't fix it
  after scalar filter:     ok
```

**Only U+FFFC is in the set.** I swept 55 candidates — every C0 control character,
U+FFFD, ZWSP, BOM, LS/PS, soft hyphen, bidi controls, word joiner, interlinear
annotation (U+FFF9–FFFB), private use, invisible math operators. U+FFFC was the
only one that hangs, so nothing else is dropped. The set makes it easy to extend
if another one turns up.

Only the display title is affected; `HistoryItemContent` is untouched, so pasting
still yields the original data including its attachments. Dedup compares contents
via `supersedes()`, not titles, so matching is unaffected.

## Testing

`UnsafeForTitleLayoutTests` (17 assertions) covers leading/trailing/interior/
multiple occurrences, all-unsafe input, the grapheme-cluster case, and that
U+FFFD, ZWSP, BOM, emoji, flag sequences, combining marks, and `⏎`/`⇥`/`·` all
survive. Two cases added to `HistoryItemTests` cover the full `generateTitle()`
path.

I should be straight about what I could and couldn't run: this machine has only
CommandLineTools, no full Xcode, so **I could not run `xcodebuild` or the XCTest
suite** — CI will be the first real run of them. What I did verify by compiling
the two new source files directly with `swiftc`:

- all 17 test assertions pass as plain checks against the real extension source
- every string previously confirmed to hang renders fine after sanitization
  (CJK, Japanese, Hangul, emoji, multiple occurrences, the combining-mark case)
- an unsanitized control still hangs in the same harness, so the check isn't vacuous
- `plutil -lint` passes on the modified `project.pbxproj`, and the four new object
  IDs resolve correctly in the build graph

Please give the Xcode build and the test suite a look, since I couldn't.

One thing I'd flag for your judgement: I picked the migration date `2026-08-31`
for the key. If this lands on a different date you may want to rename it, though
since the key only needs to be unique it works either way.

Happy to split the migration into its own PR, drop it, or adjust the naming if
you'd prefer a smaller change.


Primer-Fan:
CI is red here, but the failure is pre-existing and unrelated to this change. I
pulled the Bitrise log for this build (`1b296a4e`) and every other recent build
to check.

The only failing suite is `ClipboardTests`, and it's the same two tests every
time (each retried 3× by the test plan, so 6 reported failures):

```
ClipboardTests testIgnoreApplication            × 3
ClipboardTests testIgnoreAllApplicationsExcept  × 3
Executed 95 tests, with 6 failures
```

Every other suite passes. Same two tests, same counts, on the three other open
fork PRs — all touching unrelated code:

| PR | Topic | Failing tests |
|---|---|---|
| #1517 | numeric settings inputs | same 2 × 3 |
| #1521 | pins position | same 2 × 3 |
| #1522 | in-memory-only history | same 2 × 3 |
| #1524 | this one | same 2 × 3 |

Last green build was `master` on 2026-08-13; every build since then has failed
the same way, so it looks like a change on the Bitrise runner rather than in the
repo — `master` itself hasn't moved since 2026-08-12 (`d994f91`).

Both tests hardcode `Defaults[.ignoredApps] = ["com.apple.dt.Xcode", "com.apple.finder"]`
with the comment `// Finder is on Bitrise`, and `Clipboard` resolves the source
app from `NSWorkspace.shared.frontmostApplication`. The runner's TCC database in
this build lists only `com.apple.dt.Xcode`, `com.apple.dt.Xcode-Helper`,
`com.apple.Terminal`, `com.apple.screensharing.agent`, the Bitrise guest agent
and the UI-test runner — no Finder. If Finder is no longer frontmost on the
image, `testIgnoreApplication` stops matching the ignore list (its inverted
expectation gets fulfilled) and `testIgnoreAllApplicationsExcept` never gets its
copy through (2 s timeout), which is exactly the two failure modes in the log.

I realise #1507 proposed making these two host-independent and you closed it as
not worth it, so I'm not relitigating that — just flagging that it's currently
gating every fork PR, and I have no way to get this one green on my own.

For what it's worth, everything this PR adds did pass in the same run:

```
Test Suite 'UnsafeForTitleLayoutTests' passed
  Executed 11 tests, with 0 failures

HistoryItemTests testTitleWithObjectReplacementCharacters      passed
HistoryItemTests testTitleWithOnlyObjectReplacementCharacters  passed
```

SwiftLint passed (✓ 3.80 sec) with both new files linted, and the build compiled
cleanly, which also confirms the hand-edited `project.pbxproj` entries are wired
up correctly.

Happy to rebase whenever the runner issue is sorted, or to adjust anything in the
change itself.

PR #1525, opened 2026-09-01
Fix unselected clipboard items collapsing to ellipsis ("...") on macOS <26

Epikaigle:
## Problem

On macOS 15 (Sequoia), unselected text entries in the history list collapse and render solely as an ellipsis (`...`). Only the currently hovered/selected item or items matching an active search query display their full text.

## Root Cause

In `Maccy/Views/ListItemTitleView.swift`, `.drawingGroup()` was added as a workaround for text flipping upside down on macOS 26 (#1113):

```swift
} else {
  title()
    .accessibilityIdentifier("copy-history-item")
    .lineLimit(1)
    .truncationMode(.middle)
    // Workaround for macOS 26 to avoid flipped text
    // https://github.com/p0deje/Maccy/issues/1113
    .drawingGroup()
}
```

On macOS 15, applying `.drawingGroup()` to the SwiftUI `Text` view inside the list row prevents SwiftUI from computing the intrinsic width during offscreen layout passes for inactive/unselected rows. Because the proposed layout width resolves to zero, `.truncationMode(.middle)` collapses the entire string to `...`.

When a search query is active, `item.attributedTitle` is populated, so it enters the `if let attributedTitle` branch (which does not use `.drawingGroup()`), causing the text to render normally during searches.

## Fix

Guard `.drawingGroup()` with `#available(macOS 26.0, *)` so the workaround is only active on macOS 26+ where the flipped text bug occurs, leaving macOS 15 and earlier on the standard SwiftUI text layout path.


p0deje:
@Epikaigle Can you please show a screenshot of the problem before/after the fix?

Epikaigle:
Sure! Here is the visual comparison on macOS 15 (Sequoia):

### Before (with `.drawingGroup()` active)
*All unselected entries collapse and render only as `...`, while only the currently hovered/selected item displays text:*

Before

---

### After (with `.drawingGroup()` guarded by `#available(macOS 26.0, *)`)
*All clipboard items render their title and text properly as expected:*

After

PR #1526, opened 2026-09-01
Add snippet support

### LINKED ISSUE #1424: Add snippet support
### Before Submitting Your Feature Request

- [x] Check that there isn't already a similar feature request to avoid creating a duplicate.

### Problem

Pinning items is useful, however, the list of pinned items can get very long and there's no way to even categorize them.
Please consider adding snippet support in the form of folders, in which we can group specific commands for example. An example of this is Clipy: 

https://github.com/Clipy/Clipy

### Solution

_No response_

luizanao:
## Summary
- add persistent, ordered snippet folders with multiline plain-text snippets
- add a Snippets preferences pane for creating, editing, deleting, and reordering folders and snippets
- integrate a searchable snippet browser into the popup with a configurable global shortcut, number shortcuts, arrow navigation, and existing copy/paste modifier behavior

image
image

## Testing
- `xcodebuild -project Maccy.xcodeproj -scheme Maccy -configuration Debug build CODE_SIGNING_ALLOWED=NO`
- `xcodebuild -project Maccy.xcodeproj -scheme Maccy -configuration Debug test CODE_SIGNING_ALLOWED=NO -only-testing:MaccyTests/SnippetsTests`
- manually verified preferences CRUD, popup entry from the footer/global shortcut, keyboard focus, folder navigation, and snippet insertion

The full unit target was also exercised; two existing frontmost-application-dependent `ClipboardTests` fail in this local environment.

Closes #1424


luizanao:
cc @p0deje , apologies for the ping, just not sure exactly how this repo is maintained. This feature is really what is keeping me using clippy

p0deje:
My current take on the snippets is that if we add a way to group pins in folders, the rest can be implemented without any new extra UI. @weisJ has a PR that allows editing items' title/content inline and changing their order by drag&drop (#1447). Once his PR lands, we could extend drag&drop to allow creating folders when one pin is dropped onto another pin (similar to how iOS creates folders for apps). 

Would you be open to implementing the folder support on top of #1447? Maybe I'm missing something, but it seems like it should cover all the snippets' use cases. We'll still need to figure out the UI, though (e.g., how folders are expanded and collapsed).

PR #1527, opened 2026-09-02
Limit large text preview rendering

HerbertGao:
## Summary

- cap preview text at 10,000 characters before passing it to TextKit
- use `String.prefix` in `shortened(to:)` to avoid a full `String.count` traversal before truncation
- verify the exact preview prefix while preserving the complete stored clipboard text
- make source-application ignore tests deterministic by injecting fixed bundle IDs into the full pasteboard change path

## Motivation

`PreviewItemView` renders `HistoryItemDecorator.previewText`. On the base branch that property bypasses the existing 10,000-character limit on `text`, so multi-megabyte strings can be passed to `LargeTextPreviewView` in full.

The image preview branch uses `AsyncView`, but the large-text branch is an `NSViewRepresentable` path. Bounding the string passed into `NSTextView` avoids giving TextKit the complete clipboard payload for a visual preview.

## Repeated UI reproduction

A temporary XCUITest used the private one-line JSON sample from the original report (2,820,999 UTF-8 bytes, 2,210,051 code points, including 283,189 CJK code points). The sample contents and filename were not uploaded.

On `origin/master` at `39e0ba5`, XCUITest could not query the popup after the status-item click. Three attempts (the initial run plus two automatic retries) were right-censored by accessibility-query timeouts at 125.498, 113.735, and 113.872 seconds; the maximum observed Maccy CPU samples were approximately 100%. The planned explicit hover loop was never reached.

On patched `a7bf4a3`, three independent runs completed successfully in 27.120–28.917 seconds, including ten round trips between the large-text row and an adjacent row.

The UI runs were not randomized or interleaved: patched run 1 was followed by one base invocation containing three attempts (the initial run plus two automatic retries), then patched runs 2 and 3. XCUITest accessibility snapshotting can add overhead, so these are test-path results rather than ordinary UI timings. They show a consistent local association under this setup, not a causal profile. Detailed methodology and repeated standalone TextKit results are in the updated PR comment.

## Behavior boundary

Only the displayed preview is shortened. `HistoryItem.contents` remains unchanged and continues to supply the complete data when an item is copied or pasted.

`item.previewableText` is still fully materialized before truncation. This change limits TextKit input; it does not eliminate full UTF-8 decoding or RTF/HTML parsing.

## Validation

- `xcodebuild test -project Maccy.xcodeproj -scheme Maccy -destination 'platform=macOS' -only-testing:MaccyTests CODE_SIGNING_ALLOWED=NO`: 92 tests passed, 0 failures
- commit `a7bf4a3` passed the complete Bitrise workflow, including 36 UI tests
- the dedicated private-sample XCUITest A/B produced three base timeouts and three successful patched runs as described above
- a tester manually confirmed that a Release build containing the preview changes later committed as `26bf532` no longer stalled in their large-text scenario

### DIFF
--- a/Maccy/Clipboard.swift
+++ b/Maccy/Clipboard.swift
@@ -34,8 +34,12 @@ class Clipboard {
   private var disabledTypes: Set<NSPasteboard.PasteboardType> { supportedTypes.subtracting(enabledTypes) }
 
   private var sourceApp: NSRunningApplication? { NSWorkspace.shared.frontmostApplication }
+  private let sourceAppBundleIdentifier: () -> String?
 
-  init() {
+  init(sourceAppBundleIdentifier: @escaping () -> String? = {
+    NSWorkspace.shared.frontmostApplication?.bundleIdentifier
+  }) {
+    self.sourceAppBundleIdentifier = sourceAppBundleIdentifier
     changeCount = pasteboard.changeCount
   }
 
@@ -177,7 +181,8 @@ class Clipboard {
       return
     }
 
-    if let sourceAppBundle = sourceApp?.bundleIdentifier, shouldIgnore(sourceAppBundle) {
+    let sourceAppBundle = sourceAppBundleIdentifier()
+    if let sourceAppBundle, shouldIgnore(sourceAppBundle) {
       return
     }
 
@@ -224,7 +229,7 @@ class Clipboard {
       try? History.shared.insertIntoStorage(historyItem)
     }
 
-    historyItem.application = sourceApp?.bundleIdentifier
+    historyItem.application = sourceAppBundle
     historyItem.title = historyItem.generateTitle()
 
     onNewCopyHooks.forEach({ $0(historyItem) })
--- a/Maccy/Extensions/String+Shortened.swift
+++ b/Maccy/Extensions/String+Shortened.swift
@@ -1,9 +1,5 @@
 extension String {
   func shortened(to maxLength: Int) -> String {
-    guard count > maxLength else {
-      return self
-    }
-
-    return String(self[startIndex..<index(startIndex, offsetBy: maxLength)])
+    String(prefix(maxLength))
   }
 }
--- a/Maccy/Extensions/String+Shortened.swift
+++ b/Maccy/Extensions/String+Shortened.swift
@@ -1,5 +1,5 @@
 extension String {
   func shortened(to maxLength: Int) -> String {
-    String(prefix(maxLength))
+    String(prefix(max(0, maxLength)))
   }
 }
--- a/Maccy/Extensions/String+Shortened.swift
+++ b/Maccy/Extensions/String+Shortened.swift
@@ -1,5 +1,5 @@
 extension String {
   func shortened(to maxLength: Int) -> String {
-    String(prefix(max(0, maxLength)))
+    String(prefix(maxLength))
   }
 }
--- a/Maccy/Observables/HistoryItemDecorator.swift
+++ b/Maccy/Observables/HistoryItemDecorator.swift
@@ -44,14 +44,12 @@ class HistoryItemDecorator: Identifiable, Hashable, HasVisibility {
   var previewImageGenerationTask: Task<(), Error>?
   var thumbnailImageGenerationTask: Task<(), Error>?
   var previewImage: NSImage?
-  var previewText: String {
-    item.previewableText
-  }
+  // 10k characters seems to be more than enough on large displays
+  var previewText: String { item.previewableText.shortened(to: 10_000) }
   var thumbnailImage: NSImage?
   var applicationImage: ApplicationImage
 
-  // 10k characters seems to be more than enough on large displays
-  var text: String { previewText.shortened(to: 10_000) }
+  var text: String { previewText }
 
   var isPinned: Bool { item.pin != nil }
   var isUnpinned: Bool { item.pin == nil }
--- a/Maccy/Observables/HistoryItemDecorator.swift
+++ b/Maccy/Observables/HistoryItemDecorator.swift
@@ -45,7 +45,16 @@ class HistoryItemDecorator: Identifiable, Hashable, HasVisibility {
   var thumbnailImageGenerationTask: Task<(), Error>?
   var previewImage: NSImage?
   // 10k characters seems to be more than enough on large displays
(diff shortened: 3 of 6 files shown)


p0deje:
Can you please demonstrate the problem you're trying to fix? Since previews are now async, they should be able to handle larger texts without blocking UI. It would be great if you could record a video showing the performance problem.

HerbertGao:
Thanks for asking for a demonstration. I repeated the tests and am replacing the earlier one-off benchmark with the results below. The previous 26.82-second synthetic figure did not reproduce under repeated runs and should not be used.

### 1. Symptoms

The reported behavior was a minute-scale UI stall after opening Maccy from the status bar with a roughly 3 MB minified JSON item at the top of recent history. A Release build containing the preview changes later committed as `26bf532` no longer stalled for the same tester.

### 2. Reproduction

The minute-scale symptom above is the tester's manual report. Separately, I ran a local XCUITest A/B on a MacBookPro18,2 (M1 Max, 32 GB), macOS 27.0 (26A5421a), Xcode 27.0 (27A5252f), using the same private JSON sample involved in that report. I did not upload its contents or filename.

Sample shape:

- 2,820,999 UTF-8 bytes;
- 2,210,051 Unicode code points on one line;
- 283,189 CJK code points;
- longest non-whitespace run: 202,693 characters;
- valid JSON, with no U+FFFC or unexpected control characters.

The same temporary XCUITest was applied to base `39e0ba5` and patched `a7bf4a3`. Both used `enable-testing`, isolated preferences, and an in-memory SwiftData store. The test copied the sample, clicked the status item, waited for the popup, and then planned ten round trips between the large row and an adjacent row. The Maccy process was polled with `ps`; actual intervals ranged from 0.478 to 3.013 seconds. The RSS/CPU values below are maximum observed samples, not continuous process peaks.

### 3. Code path and measured evidence

#### Real Maccy XCUITest A/B

On the base revision, the Xcode test plan made three attempts (the initial run plus two automatic retries). Every attempt timed out while querying the popup after the status-item click, before the explicit hover loop was reached:

| Base attempt | Test duration at query timeout | Maximum observed Maccy RSS | Maximum observed Maccy CPU | Outcome |
| ---: | ---: | ---: | ---: | --- |
| 1 | 125.498 s | 206.2 MiB | 100.0% | UI snapshot timeout |
| 2 | 113.735 s | 278.8 MiB | 100.1% | UI snapshot timeout |
| 3 | 113.872 s | 263.4 MiB | 100.1% | UI snapshot timeout |

The repeated failure was `Failed to get matching snapshots: Timed out while evaluating UI query`. All three base durations are right-censored by that timeout. They show that XCUITest could not query the popup for more than a minute on the base revision; accessibility snapshotting can add overhead, so they are not measurements of ordinary user-interaction latency.

The patched revision completed in three independent runs:

| Patched attempt | Popup query | Ten row round trips | Full test | Maximum observed Maccy RSS | Maximum observed Maccy CPU |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 7.085 s | 10.248 s | 28.917 s | 152.4 MiB | 67.3% |
| 2 | 6.865 s | 9.965 s | 27.655 s | 162.5 MiB | 58.5% |
| 3 | 6.889 s | 9.878 s | 27.120 s | 161.6 MiB | 64.0% |

The base attempts never reached the explicit hover loop, so there is no base-versus-patched hover-duration ratio to report.

#### Repeated standalone TextKit check

I also rebuilt the standalone `NSTextView` benchmark and ran five independent processes per group. Within each layout mode, full/capped process order alternated. Each process created five text views with a 28,446,720-byte input or its 10,000-character prefix:

| Mode | Input | n | Median total time (range) | Median maximum observed process RSS (range) |
| --- | --- | ---: | ---: | ---: |
| no explicit `ensureLayout` | full | 5 | 0.132 s (0.128–0.151) | 197.4 MiB (197.4–197.5) |
| no explicit `ensureLayout` | capped | 5 | 0.074 s (0.061–0.100) | 51.2 MiB (51.1–51.2) |
| explicit `ensureLayout(documentRange)` | full | 5 | 66.794 s (65.893–68.600) | 1696.1 MiB (956.2–2482.2) |
| explicit `ensureLayout(documentRange)` | capped | 5 | 0.105 s (0.094–0.156) | 55.7 MiB (55.7–55.7) |

The forced-layout mode is intentionally stronger than Maccy's production path, which does not explicitly call `ensureLayout(documentRange)`. These standalone numbers are not hover timings or Maccy-process memory measurements.

The code-level difference is narrower than my earlier explanation:

- `AsyncView` covers the image branch; large text goes through `LargeTextPreviewView`, an `NSViewRepresentable`, and is assigned to `NSTextView.string` on the UI path.
- On the base revision, `HistoryItemDecorator.previewText` passes the full `item.previewableText` to that view.
- The UI runs were not randomized or interleaved: patched run 1 was followed by one base invocation containing three attempts (the initial run plus two automatic retries), then patched runs 2 and 3. They show a consistent local association between the bounded input and completion under this setup, but I did not capture an Instruments stack trace and am not claiming a complete causal profile.
- I found no evidence establishing an event-queue loop, autorelease issue, swap/page-fault chain, or memory leak, and I withdraw those earlier claims.

### 4. Fix

- Bound `HistoryItemDecorator.previewText` to 10,000 characters before passing it to `LargeTextPreviewView` / TextKit.
- Implement `shortened(to:)` with `String.prefix`, avoiding the previous full `String.count` traversal before slicing.
- Keep `HistoryItem.contents` unchanged; the regression test verifies the exact preview prefix and the complete original 2 MiB stored text.

`item.previewableText` is still fully materialized before truncation. This patch limits the amount handed to TextKit; it does not claim to remove full UTF-8 decoding or RTF/HTML parsing.

HerbertGao:
@p0deje Could you please take another look when you have a chance?

I updated both the PR description and my earlier comment after repeated testing with the original large, single-line JSON shape:

- base `39e0ba5`: all three XCUITest attempts hit right-censored popup accessibility-query timeouts at 113–125 seconds;
- patched `a7bf4a3`: three independent runs completed successfully;
- the earlier one-off synthetic benchmark and unsupported event-queue/swap explanations were explicitly corrected;
- the updated comment now separates manual observations, real Maccy XCUITest results, standalone TextKit measurements, and remaining evidence limits.

The current commit also passes the complete Bitrise workflow. Thanks!

Issue #1515, opened 2026-08-20
Separator between boots in persisted items.


tomgron:
### Before Submitting Your Feature Request

- [x] Check that there isn't already a similar feature request to avoid creating a duplicate.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Problem

I like the idea of having the copy history between boots of my mac but would be good if older items could be separated between boot and/or days, much like browser history. Also, clearing items from history could be limited to "older than 1 hour" or "older than 1 day" etc.

PR #1528, opened 2026-09-04
Add optional day separators for unpinned history

md786-dotcom:
This pull request adds one part of #1515.

It adds day separators only. It does not include boot separators. It does not include "clear older than".

## Changes

The Appearance pane has a new option: **Show day separators**. The option is off by default.

The unpinned list shows date headers when these conditions are true:

- The option is on.
- The search field is empty.
- The sort mode is by date.

The headers are:

- Today
- Yesterday
- The weekday and the date for older days (example: Monday, 1 Sep)

The list hides the first header when that day is today. This keeps the popup short.

Pinned items stay in the pin block. You cannot select a header. The headers do not use the ⌘1–⌘9 shortcuts. The list does not show headers during a search. The list does not show headers when the sort mode is "number of copies".

New files:

- `DayGrouper` puts adjacent unpinned items into groups. Each group is one calendar day.
- `DaySeparatorView` shows the header.
- `DayGrouperTests` tests same-day groups, midnight boundaries, the first-copied sort mode, and filtered lists.

The English strings are in `AppearanceSettings.strings`. Weblate can supply the translations for the other languages.

## Testing

- `DayGrouperTests` passed.
- SwiftLint passed on the changed Swift files.
- Manual tests passed:
  - The headers show for each day when the option is on.
  - The headers do not show during a search.
  - The list stays flat when the option is off.
  - Pins, the ⌘1–⌘9 shortcuts, and the arrow keys operate as before.

## Additional

**Local signing.** Xcode tried to sign the app with the upstream development team. I do not have that certificate. I signed the local build with my own development identity. This let me run the app for the manual tests. This pull request does not change the project team ID.

**Unrelated local failures.** Two tests failed when I ran the full suite from the terminal: `ClipboardTests.testIgnoreApplication` and `ClipboardTests.testIgnoreAllApplicationsExcept`. Those tests need Xcode or Finder as the frontmost application. The frontmost application was the terminal. This change does not cause those failures. A local Prettier hook also failed on `Maccy.xctestplan`. Prettier has no parser for that file type. That hook is local only.

Made with [Cursor](https://cursor.com)


md786-dotcom:
The Bitrise check failed with "Failed - Maccy". This looks like the same pre-existing CI issue described in #1524 

Locally, the full `MaccyTests` suite fails on the same two tests:

- `ClipboardTests.testIgnoreApplication`
- `ClipboardTests.testIgnoreAllApplicationsExcept`

Both tests assume Finder or Xcode is the frontmost application. The frontmost application was the terminal during my run. This matches the Bitrise runner analysis in #1524.

Tests added by this pull request passed:

- `DayGrouperTests` — 5/5 passed
- SwiftLint passed on the changed Swift files

This pull request does not change `ClipboardTests` or the clipboard ignore logic. I believe the Bitrise failure is unrelated to the day separator change, in line with #1524.

Issue #1097, opened 2025-05-16
Feature Request: Allow Larger Clipboard History and Optimize UI for Long Histories


victorwads:
### Before Submitting Your Feature Request

- [x] Check that there isn't already a similar feature request to avoid creating a duplicate.
- [x] I have seen the [FAQ](https://github.com/p0deje/Maccy?tab=readme-ov-file#faq).

### Problem

Today, Maccy enforces a hardcoded clipboard history limit of 999 entries. This restricts users who wish to keep extensive histories (e.g., 10,000 to 100,000 items) for productivity or research purposes.

Although the historySize can be increased manually via:
```
defaults write org.p0deje.Maccy historySize -int 90999
```

The app still struggles from 30k+ items because of how it loads all entries at once into the UI. For example:
* With historySize = 50000, the database grows accordingly.
* Upon opening Maccy, it attempts to render all 50,000 items, leading to freezes or sluggish scrolling.
* Even if a user only wants to view the last few clips, Maccy still loads everything by default.

> ⚠️ Example: With ~43,000 items saved, performance begins to degrade slightly. Yet at 50,000 items, usage is still feasible and acceptable — but only if the UI didn’t try to eagerly render everything.

### Solution

✅ Proposed Solution

This request has two parts:
1. Remove the artificial cap of 999 entries:
    * Allow users to configure a much higher value (e.g., up to 100,000 or more).
    * Add a warning prompt or tooltip if users go above a recommended threshold (e.g., 50,000 or 100,000) explaining potential performance impacts.
2. Lazy-load UI rendering:
    * When opening Maccy, render only the most recent N entries (e.g., 100).
    * Older entries remain stored in the database and can be accessed via search.
    * This keeps the interface responsive even with very large histories.
    * Optionally, expose a setting like visibleHistorySize for power users.

FoxMcWeezer:
It's 2025. We can figure out how to hold longer clipboard histories by now

p0deje:
@FoxMcWeezer Of course we can, looking forward to your PR!

KingSupernova31:
Duplicate of #310 I think?

weisJ:
Yes, thank you :)

PR #1530, opened 2026-09-07
Add optional unlimited history with explicit archive search

B1lli:
## Problem and behavior

Large retained histories currently populate the in-memory history list and participate in ordinary searches and deduplication. This adds substantial work even when the user only needs a recent clip. This PR adds an **opt-in extended history** setting with a recent cache and an explicit action to search older records.

- Keep the default retention limit of 200. Enabling extended history permits larger positive limits or `-1` for unlimited retention; reducing retention requires confirmation. Disabling the extension hides archive controls without silently changing the saved limit or deleting existing records.
- Keep the most recently copied 1,000 unpinned records plus pins in the display cache. Ordinary search retains the selected search mode within that cache.
- **Deep search** applies a stored-title predicate before pagination in a background SwiftData context. Each page contains at most 100 matches; **Next page** replaces the page, and **Recent**, changing the query, recording a new clipboard item, or closing the popup returns to recent search. The action is available even with zero recent matches.
- Preserve the existing database schema. Cache eviction does not delete stored records. Unlimited retention skips pruning; explicit deletion/clear actions still work. Avoid archive-wide startup cleanup/title rewriting, and sanitize legacy titles at the display boundary instead. Release evicted decorators by using weak observation callbacks.

Related: #310, #1097, #1446. This addresses retrieval as well as bounded loading, rather than making older records unreachable after a fetch limit.

## Comparison with #1262

#1262 implements virtualized browsing for unlimited history. At its current head (`cfb0fb6`), search filters the loaded window; [its author identified storage-backed search as follow-up work](https://github.com/p0deje/Maccy/pull/1262#discussion_r3687724405). A nonempty query searches only that loaded window; the search view does not load additional archive pages. This is a current implementation gap, not an inherent limitation of virtualization.

This PR chooses explicit archive retrieval: the user sees the recent-search scope and can query storage without scrolling. It does not provide continuous archive browsing. Many matches still require narrowing the query or visiting successive pages; there is currently no previous-page action or total match count.

## Validation and performance

Executed on macOS 15.6 / Mac16,12 / 10 CPU cores / 32 GiB RAM / Xcode 26.3. Both revisions use optimized XCTest builds (`-O`), separate file-backed synthetic text fixtures, and fresh test processes. Values are medians of three paired runs, alternating order. Baseline is current upstream `c376789`; upstream uses a retention limit of 1,000,000,000 and the candidate uses `-1`, so neither prunes the large fixtures. These are internal history-path timings, **not popup first-paint measurements**. Search calls `Search.search` directly in exact mode, excluding input throttling and UI updates. Add timings cover synchronous `History.add` for unique records, not clipboard ingestion, duplicate-copy behavior, or completed asynchronous UI work. Candidate additions run after its archive-search samples.

| Stored records | History load, upstream → this PR | Direct exact search | Add unique record | Process footprint after load |
| --- | --- | --- | --- | --- |
| 200, default mode | 25.58 → 21.19 ms | 0.73 → 0.67 ms | 2.28 → 2.79 ms | 94.42 → 93.94 MiB |
| 12,000 retained | 1,307.88 → 120.93 ms | 43.53 → 3.56 ms | 70.61 → 7.01 ms | 142.80 → 102.47 MiB |
| 50,000 retained | 7,526.96 → 126.16 ms | 182.78 → 3.45 ms | 330.65 → 7.20 ms | 521.14 → 101.55 MiB |

Normal-search scopes differ: upstream searches the retained list, while this PR searches the recent cache. Deep search over the complete title store took 117/107 ms per page at 12,000/50,000 records (pooled nearest-rank p95: 175/155 ms). A separate full finite-limit 50,000-record pair measured add medians of 484 ms upstream and 47 ms here; finite pruning still scales with stored history. No general speedup is claimed at the default size, and image-heavy workloads and disk-cold startup were not benchmarked.

[Reproduction instructions, benchmark harness, aggregation script, and raw per-operation results](https://gist.github.com/B1lli/3c616dcf3dacf0a7e51c273aa8a5f31d).

- 62 focused XCTest regressions passed, with retries disabled: HistoryTests, HistoryItemTests, SearchTests, SorterTests, and UnsafeForTitleLayoutTests. They cover retention, persistence, archive pagination, selection, pinning, paused-recording copies, and decorator release.
- Native UI: a recent query with zero matches retrieves 205 archived matches in pages of 100 / 100 / 5; the last page removes Next page. A single archived match also works. Settings validation, disabling the extension, and cancelling a retention reduction were exercised in an isolated test app.
- Release build, changed-file SwiftLint, localized-string validation, Prettier, and `git diff --check` passed.
- Independent fresh-context implementation review: PASS for overfitting, adjacent-path compatibility, and minimal scope; no blockers. Review findings about hidden-header height and stale archive references after deduplication were fixed and re-reviewed.

Limitations: deep search matches stored titles/OCR summaries by substring, not every byte of long clips or all normal fuzzy/regex modes. Copying a record outside the cache preserves that record and creates a recent copy; global archive deduplication is not implemented.

AI assistance: OpenAI Codex was used for implementation, tests, and independent review. The validation above was executed locally against synthetic data; no private clipboard contents are included.


B1lli:
Fixed the failing CI tests in e3628ba. Both application-filter tests assumed Xcode or Finder was the foreground application and relied on a two-second polling window. The same failures reproduce on unmodified upstream c376789.

The tests now use the actual foreground application and call the existing clipboard-change entry point synchronously. Both allow and deny paths remain asserted, including a non-empty, non-matching application list. Production code is unchanged; the separate polling integration test remains in place.

Validation: all 102 unit tests pass locally with retries disabled; strict SwiftLint and independent review pass. The new Bitrise build is running: https://app.bitrise.io/app/716921b669780314/build/43efd804-16eb-45b2-bd3d-473e774b2ccd

B1lli:
Remote verification complete for e3628ba: Bitrise is green. All 102 unit tests and 36 UI tests passed with zero failures; both previously failing application-filter tests passed on their first attempt (no retries). SwiftLint also passed.

Build: https://app.bitrise.io/app/716921b669780314/build/43efd804-16eb-45b2-bd3d-473e774b2ccd

PR #1541, opened 2026-09-15
Add Paste as Markdown action

frblondin:
## What

Adds a new "Paste as Markdown" action, bound to ⌘⌥ (Command+Option) while selecting a
history item — a previously unused modifier combination. Instead of pasting the item's
original rich formatting, it converts HTML/RTF clipboard content to Markdown before
pasting.

- Word/HTML documents: bold/italic/links/headings/lists → Markdown equivalents.
- Excel/Word tables: converted to GFM pipe tables, with the first row promoted to a
  header when the source has no `<th>` (which Excel/Word clipboard HTML never emits).
- Falls back to RTF (re-exported as HTML) or plain text when no HTML flavor is present.

image

## How

- Uses [html-to-markdown-swift](https://github.com/jaredhowland/html-to-markdown-swift)
  (SwiftSoup-based) as the HTML→Markdown engine, configured with `TablePlugin(options:)`
  registered ahead of `GFMPlugin()` for `.minimal` cell padding and `headerPromotion`.
  _Note: it could use a custom implementation if you don't want to add any dependency..._
- A small whitespace-normalization pre-pass collapses insignificant `\r\n`/`\t` runs
  outside `<pre>` blocks, since Word/Excel clipboard HTML is hard-wrapped and would
  otherwise leak stray line breaks into inline text.
- New `MarkdownConverter` (Maccy/MarkdownConverter.swift) owns pasteboard-flavor
  selection (HTML → RTF-as-HTML → plain string) and the conversion pipeline.
- `HistoryItemAction.pasteAsMarkdown` wires the new ⌘⌥ shortcut into the existing
  select/paste flow in `History.swift`.

## Testing

- Added `MaccyTests/MarkdownConverterTests.swift` covering bold/italic/links/lists,
  Excel-style tables (with and without a real header row), the Word CRLF stray-line-break
  regression, RTF fallback, plain-string fallback, and unsupported-content → `nil`.
- Manually verified against real clipboard captures from Word and Excel.
- `README.md` and the General Preferences pane updated to document the new shortcut.


frblondin:
Bitrise build is failing in tests that appear unrelated to the PR. The relevant results are:

- `MarkdownConverterTests`: **10 tests passed**
- Existing `MaccyTests`: **6 failures**, caused by two tests being retried:
  - `ClipboardTests.testIgnoreApplication`
  - `ClipboardTests.testIgnoreAllApplicationsExcept`
- `MaccyUITests`: **2 failures**, both in:
  - `MaccyUITests.testDeleteEntryDuringSearch`

The clipboard tests assume that Xcode or Finder is the frontmost application:

```swift
Defaults[.ignoredApps] = ["com.apple.dt.Xcode", "com.apple.finder"]
```

On Bitrise, the foreground application appears to be different, so the expectations fail. The UI test times out waiting for the history item to disappear, after which the app terminates while XCTest is collecting diagnostics.

The final result-bundle error appears to be secondary:

```text
Could not write the 3_Test action log because the log hasn't finished recording after waiting 30 seconds
```

Neither `ClipboardTests.swift` nor `MaccyUITests.swift` was modified by this PR. The Markdown-related tests pass successfully.

Could you advise whether these tests are known to be flaky or environment-dependent on Bitrise, and whether there is a preferred way to run or configure the CI test step for this repository? In particular, I’m unsure whether I should fix the existing test assumptions in this PR or leave them as unrelated CI issues.

PR #1545, opened 2026-09-17
Clean up orphaned contents in batches without blocking launch

### LINKED ISSUE #1535: Startup orphan cleanup from #1509 needs free disk equal to the store size and blocks launch for minutes on real-world stores

`master` (c376789) — the unreleased `cleanupOrphanedContents` migration from #1509. 2.7.1 (62) installed.

### macOS Version

26.6.2, Apple Silicon

### Maccy Settings

```
historySize = 200;
enabledPasteboardTypes = ("public.jpeg", "public.tiff", "public.file-url", "public.heic", "public.rtf", "public.png", "public.utf8-plain-text", "public.html");
```

### Description

#1509 stops the leak from #1496 going forward and adds a one-time migration that deletes the orphans already in the store through the context:

```swift
try context.delete(model: HistoryItemContent.self, where: #Predicate { $0.item == nil })
context.processPendingChanges()
try context.save()
```

The stores that need this are large — nobody notices the leak before it has eaten tens of GB. Mine was **81 GB with 398,886 orphaned rows against 511 live ones** when I rebuilt it this week (2.7.1, 49 days after the previous rebuild; the store before that had reached 211 GB / 1.14M orphans on 2.6.1, so the growth is ~1.7 GB/day with images enabled). I ran the migration's exact code against a synthetic store with the same two models — 200 items, 500 live contents, 100,000 orphaned `HistoryItemContent` rows of 200 KB each, 20.5 GB on disk:

| | |
|---|---:|
| wall time of the migration (main thread, before the panel exists) | **140 s** |
| peak `Storage.sqlite-wal` during the save | **20.5 GB — the size of the store** |
| `ACHANGE` rows added by the delete | 100,000 |
| peak RSS | 149 MB |
| file size afterwards | 40 MB (Core Data's post-save maintenance does vacuum it, so the space is reclaimed) |

It scales roughly with the store size: the 81 GB store above would block for ~10 minutes, the 211 GB one for ~25.

Three consequences:

1. **The migration needs free disk space equal to the store size.** The delete is a single transaction, and the WAL keeps every page it frees until the commit. The machines that have this problem are exactly the ones without another 81 GB to spare — the leak is what filled the disk. When the WAL cannot grow, the save fails with `SQLITE_FULL`.
2. **A failed cleanup is recorded as done.** `ensureMigration` sets the key unconditionally after `action()`, and the action is `_ = try? Storage.shared.cleanupOrphanedContents()`, so any error marks the migration complete and it is never retried. The users who needed the cleanup most keep their orphans forever.
3. **A force-quit repeats the whole thing.** The migration runs in `applicationDidFinishLaunching` before `FloatingPanel` is created, so for minutes the menu bar icon is there but nothing responds. A user who force-quits — as one does with an app that hangs on launch — never gets the key set, and every launch re-runs the same minutes-long transaction, rolled back each time. From the outside that is "Maccy is broken since the update".

(weisJ raised the blocking concern in the #1509 review; the numbers above are what it looks like on a real store.)

### Steps to reproduce

<details>
<summary>Seed + measure (Swift, `swiftc -O repro.swift`)</summary>

```swift
import Foundation
import SwiftData

@Model final class HistoryItem {
  var title: String = ""
  @Relationship(deleteRule: .cascade, inverse: \HistoryItemContent.item) var contents: [HistoryItemContent] = []
  init(title: String) { self.title = title }
}
@Model final class HistoryItemContent {
  var type: String = ""
  var value: Data?
  @Relationship var item: HistoryItem?
  init(type: String, value: Data? = nil) { self.type = type; self.value = value }
}

let url = URL(fileURLWithPath: CommandLine.arguments[1])
let container = try ModelContainer(for: HistoryItem.self, HistoryItemContent.self, configurations: ModelConfiguration(url: url))

if CommandLine.arguments[2] == "seed" {
  let context = ModelContext(container)
  context.autosaveEnabled = false
  for i in 0..<200 {
    let item = HistoryItem(title: "item \(i)")
    context.insert(item)
    let content = HistoryItemContent(type: "public.utf8-plain-text", value: Data(repeating: 1, count: 2000))
    context.insert(content)
    content.item = item
  }
  try context.save()
  for _ in 0..<200 {  // 100,000 orphans x 200 KB
    let batch = ModelContext(container)
    batch.autosaveEnabled = false
    for _ in 0..<500 { batch.insert(HistoryItemContent(type: "public.tiff", value: Data(repeating: 7, count: 204_800))) }
    try batch.save()
  }
} else {  // Storage.cleanupOrphanedContents() from master
  let started = Date()
  try MainActor.assumeIsolated {
    let context = container.mainContext
    try context.delete(model: HistoryItemContent.self, where: #Predicate { $0.item == nil })
    context.processPendingChanges()
    try context.save()
  }
  print("wall \(Date().timeIntervalSince(started))s")
}
```

Run `./repro store.sqlite seed`, then `./repro store.sqlite cleanup` while watching `ls -l store.sqlite-wal` from another shell.
</details>

### Suggested fix

The live data is tiny — `historySize` items with a few contents each, 17 MB inside my 81 GB store — so rebuilding the store is O(live) instead of O(garbage): copy `ZHISTORYITEM`, the `ZHISTORYITEMCONTENT` rows with `ZITEM IS NOT NULL`, and the `Z_PRIMARYKEY` / `Z_METADATA` / `Z_MODELCACHE` rows into a fresh file and swap it in. That took 2.4 s on the 20.5 GB synthetic store and 4.6 s on my real 81 GB one, with no WAL growth, no free-space requirement, and nothing to resume. It has to run before the `ModelContainer` opens, so it would sit at the top of `Storage.init` behind the same migration key.

If you would rather stay with the SwiftData delete, the minimum is: check `volumeAvailableCapacityForImportantUsage` against the store size before starting, only set the migration key when `cleanupOrphanedContents()` did not throw, and delete in batches with a save per batch so the WAL stays small (Core Data checkpoints after every save) and a force-quit keeps the progress made so far.

Happy to send a PR for either.

nanami-0713:
Fixes #1535.

## Problem

`cleanupOrphanedContents()` (introduced in #1509) deleted all orphans in a **single transaction**: one `context.delete(model:where:)` followed by one `save()`. On real-world stores this means:

- the SQLite WAL retains every freed page until commit, so peak disk need ≈ store size (`SQLITE_FULL` when space runs out), and
- the migration runs synchronously inside `applicationDidFinishLaunching` **before the panel is created**, so Maccy shows no window for the full duration — minutes on large stores (#1535 measured 140 s on a 20.5 GB store).

Two smaller defects in the same path:

- `ensureMigration` marks the key done **even when the action fails** (`_ = try?`), so a failed cleanup is permanently skipped; and
- the reverse (force-quit mid-cleanup) restarts the entire giant transaction on the next launch.

## Changes

- `Storage.cleanupOrphanedContents(batchSize:)` now deletes orphans in batches of 500 on a background `ModelContext` via `Task.detached`: each batch is its own transaction, so the WAL stays ~batch size, fetched content blobs stay lazy faults (memory stays flat), and committed batches survive a force-quit — an interrupted cleanup resumes where it stopped instead of starting over.
- The cleanup moved out of `migrateUserDefaults()` and now runs after the panel is shown (`Task { await cleanupOrphanedContentsIfNeeded() }`), so launch is never blocked. `sanitizeTitles` stays synchronous on purpose (it must complete before first render, per #1520).
- The migration key is only set **after** the cleanup succeeds. It is also a fresh key (`...-in-batches`): users whose old synchronous attempt failed with `SQLITE_FULL` already have the old key set to true and would otherwise never be healed; for everyone else the extra pass is one cheap `fetch` on an already-clean store.

## Testing

- `testCleaningUpOrphanedContents` updated to the async signature; added `testCleaningUpOrphanedContentsInBatches` (3 orphans, `batchSize: 2`) covering the multi-batch loop and cross-context visibility.
- Caveat: this machine has no full Xcode (CommandLineTools only), so I could not run `xcodebuild test` locally — the changes were verified by parsing/typechecking against the CLT SDK only. Please run the test plan before merging.

For context: this is complementary to #1534 (persistent-history purge); after both land, a follow-up could call `purgePersistentHistory()` once the batched cleanup finishes.


nanami-0713:
@p0deje Heads-up: the `Xcode Test for Mac` failure on this PR looks repo-wide rather than specific to these changes.

- `ci/bitrise` has been failing on `master` since 39e0ba5e (2026-08-31); the last green build is d994f91f (2026-08-13).
- The currently open PRs I checked (#1542–#1556) all show the same step failing with exit status 65.

Could you check the Bitrise log to see which test is failing? If it's one of the tests added in 39e0ba5e (e.g. `UnsafeForTitleLayoutTests`), it may be environment-sensitive — layout results can vary with system fonts/Xcode version.

